PR #9, opened 2019-04-19
Feature/higher half, closes #6

### LINKED ISSUE #6: Higher half
The kernel should be moved to the higher half to better facilitate user processes in the future. This can be done by putting the initial boot code at 1MiB, setting up paging and jumping to code addressed at 3 GB + 1 MiB with a rudimentary page directory.

SamTebbs33:
This patch puts the kernel in the higher half (3GiB) with a rudimentary page directory used to map in the kernel. Currently only 4MiB are mapped in, which will need to be increased as the binary size increases and we add a heap etc. (can be done by increasing KERNEL_NUM_PAGES as needed).

The control flow is as follows: _start (sets up paging) -> start_higher_half (sets up the stack) -> kmain

The usage of Zig's `newStackCall` has been replaced with explicit assembly since it was using ~10 instructions to do what should only take 2. I think this is because it has to work out the size of the stack, so I just put a label in the linker script that represents the end of the stack and reference that from the assembly. That's a cheap little optimisation.

Closes #6

### DIFF
--- a/src/kernel/kmain.zig
+++ b/src/kernel/kmain.zig
@@ -30,6 +30,8 @@ const KERNEL_NUM_PAGES = 1;
 
 // The initial page directory used for booting into the higher half. Should be overwritten later
 export var boot_page_directory: [1024]u32 align(4096) linksection(".rodata.boot") = init: {
+    // Increase max number of branches done by comptime evaluator
+    @setEvalBranchQuota(1024);
     // Temp value
     var dir: [1024]u32 = undefined;
 
--- a/src/kernel/kmain.zig
+++ b/src/kernel/kmain.zig
@@ -1,6 +1,6 @@
 //
 // kmain
-// Zig version: 
+// Zig version:
 // Author: DrDeano
 // Date: 2019-03-30
 //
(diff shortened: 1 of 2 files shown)


src/kernel/kmain.zig
```diff
@@ -17,17 +17,89 @@ const MEMINFO = 1 << 1;
...
+        idx += 1;
+    }
+    break :init dir;
+};
+
+export var kernel_stack: [16 * 1024]u8 align(16) linksection(".bss.stack") = undefined;
+
+export nakedcc fn _start() align(16) linksection(".text.boot") noreturn {
+    // Seth the page directory to the boot directory
+    asm volatile (
```
DrDeano: I don't know whether it is worth having these in there own function in a arch/x86.zig file as these are x86 specific

src/kernel/kmain.zig
```diff
@@ -17,17 +17,89 @@ const MEMINFO = 1 << 1;
...
+
+    // Map the kernel's higher half pages increasing by 4 MiB every time
+    i = 0;
+    while (i < KERNEL_NUM_PAGES) {
+        dir[idx] = 0x00000083 | (i << 22);
+        i += 1;
+        idx += 1;
+    }
+    // Increase max number of branches done by comptime evaluator
+    @setEvalBranchQuota(1024);
```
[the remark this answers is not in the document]
SamTebbs33: Good idea, that will catch cases where the other loops take more than 1000 iterations as well (if that ever happens)

src/kernel/kmain.zig
```diff
@@ -17,17 +17,89 @@ const MEMINFO = 1 << 1;
...
+    // Fill preceding pages with zeroes. May not be unecessary but incurs no runtime cost
+    while (i < KERNEL_PAGE_NUMBER - 1) {
+        dir[idx] = 0;
+        i += 1;
+        idx += 1;
+    }
+
+    // Map the kernel's higher half pages increasing by 4 MiB every time
+    i = 0;
+    while (i < KERNEL_NUM_PAGES) {
```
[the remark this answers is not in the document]
SamTebbs33: Nice, I didn't know about that

src/kernel/kmain.zig
SamTebbs33: That's the end goal, though I was planning on making another ticket where we go through the source and make a nice separation based on architecture.

DrDeano: Could do, but don't want to get out of hand with, "I'll do it later" ;)

PR #29, opened 2019-05-29
Added GDT, IDT, IRQ, updated build.zig

DrDeano:
Added new build.zig

Interrupts no work :(

Added isr

Interrupts work

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -4,11 +4,13 @@ const Builder = @import("std").build.Builder;
 const Step = @import("std").build.Step;
 const builtin = @import("builtin");
 const std = @import("std");
+const os = std.os;
 const ArrayList = std.ArrayList;
 const warn = std.debug.warn;
 const mem = std.mem;
 
 var src_files: ArrayList([]const u8) = undefined;
+var src_files_asm: ArrayList([]const u8) = undefined;
 
 fn concat(allocator: *std.mem.Allocator, str: []const u8, str2: []const u8) !std.Buffer {
     var b = try std.Buffer.init(allocator, str);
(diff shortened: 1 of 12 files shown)


build.zig
```diff
@@ -18,17 +20,48 @@ fn concat(allocator: *std.mem.Allocator, str: []const u8, str2: []const u8) !std
...
     src_files.append("kernel/kmain") catch unreachable;
 
     // Add the architecture init file to the source files
+    var arch_init = concat(b.allocator, "kernel/arch/", target) catch unreachable;
+    b.makePath("bin/kernel/arch") catch unreachable;
+
+    // Add the assemblies
+    const _arch_init = concat(b.allocator, arch_init.toSlice(), "/") catch unreachable;
+    const _arch_path = concat(b.allocator, src_path, "/") catch unreachable;
+    const arch_path = concat(b.allocator, _arch_path.toSlice(), arch_init.toSlice()) catch unreachable;
```
SamTebbs33: Instead of creating lots of const variables, you could just append to the existing ones

build.zig
```diff
@@ -18,17 +20,48 @@ fn concat(allocator: *std.mem.Allocator, str: []const u8, str2: []const u8) !std
...
+
+    var arch_dir = os.Dir.open(b.allocator, arch_path.toSlice()) catch unreachable;
```
SamTebbs33: It may potentially be better to add a check for the architecture and add files to `src_files_asm` based on that, something like:

```zig
switch (builtin_target) {
    builtin.Arch.i386 => {
        src_files_asm.append("src/kernel/arch/x86/irq_asm.s") catch unreachable;
        src_files_asm.append("src/kernel/arch/x86/isr_asm.s") catch unreachable;
    },
    else => {}
}
```
Since this is simpler and caters to the case when you may not want to assemble all .s files in the arch directory. What do you think?

src/kernel/arch/x86/arch.zig
```diff
@@ -45,3 +57,85 @@ pub fn inb(port: u16) u8 {
 pub fn ioWait() void {
     outb(0x80, 0);
 }
+
+// Can't use gdt.KERNEL_DATA_OFFSET in the assembly bellow, so have to store in a constant.
+const KERNEL_DATA_OFFSET: u16 = gdt.KERNEL_DATA_OFFSET;
```
SamTebbs33: You may be able to add `.extern KERNEL_DATA_OFFSET` to the inline assembly and use it in there, much like I did in the [boot code](https://github.com/SamTebbs33/pluto/blob/e90d5e4c87ecb14da608bed40c98ee7892f88be6/src/kernel/arch/x86/boot.zig#L70), not sure though

src/kernel/arch/x86/arch.zig
```diff
@@ -45,3 +57,85 @@ pub fn inb(port: u16) u8 {
...
+        sti();
+        hlt();
+        cli();
+    }
+}
+
+///
+/// Halt the kernel.
+///
+pub fn hltKernel() noreturn {
```
SamTebbs33: It may be wise to rename this to make the difference with `hlt()` more clear.

src/kernel/arch/x86/gdt.zig
```diff
@@ -0,0 +1,293 @@
...
+    .ss = 0,
+    .ds = 0,
+    .fs = 0,
+    .gs = 0,
+    .ldtr = 0,
+    .trap = 0,
+    .io_permissions_base_offset = @sizeOf(TtsEntry),
+};
+
+pub fn setKernelStack(esp0: u32) void {
```
SamTebbs33: Perhaps `setTssStack` would be a more descriptive name?

src/kernel/arch/x86/interrupt_context.zig
```diff
@@ -0,0 +1,32 @@
+// Zig version: 0.4.0
+
+pub const InterruptContext = struct {
```
SamTebbs33: This type could be moved into x86/arch.zig as I see it being used for multitasking as well.

src/kernel/arch/x86/irq.zig
```diff
@@ -0,0 +1,122 @@
...
+fn unhandled(context: *InterruptContext) void {
+    const interrupt_num = context.int_num - IRQ_OFFSET;
+    panic.panicFmt(null, "Unhandled IRQ number {}", interrupt_num);
+}
+
+// TODO Move to PIC
+fn sendEndOfInterrupt(irq_num: u8) {
+    if (irq >= 8) {
+        arch.out(0xA0, 0x20);
+	}
```
SamTebbs33: There's a weird indentation issue going on here :)

src/kernel/arch/x86/arch.zig
```diff
@@ -45,3 +57,85 @@ pub fn inb(port: u16) u8 {
...
+
+/// Load the IDT into the CPU.
+pub fn lidt(idt_ptr: *const idt.IdtPtr) void {
+    asm volatile ("lidt (%%eax)" : : [idt_ptr] "{eax}" (idt_ptr));
+}
+
+///
+/// Enable interrupts
+///
+pub fn sti() void {
```
SamTebbs33: `halt`, `sti` and `cli` are x86-specific names so more architecture-agnostic names would be `halt`, `enableInterrupts` and `disableInterrupts`.

build.zig
DrDeano: I didn't want to change the original constant as it is used further down, but have remove one of them

DrDeano: Found I can remove one more ;)

build.zig
DrDeano: Well, being a computer scientist, i had to over engineer the hell out of it. But ur idea is much simpler

src/kernel/arch/x86/arch.zig
DrDeano: This will mean i will have to export the KERNEL_DATA_OFFSET, which i don't want, but found a different way, u'll see in the next commit

src/kernel/arch/x86/gdt.zig
DrDeano: Sure

src/kernel/arch/x86/interrupt_context.zig
DrDeano: Moved

src/kernel/arch/x86/irq.zig
DrDeano: I see

src/kernel/arch/x86/arch.zig
DrDeano: Can do, just was making it clear that these are x86 instructions instead of functions

src/kernel/arch/x86/arch.zig
DrDeano: Have renamed

PR #41, opened 2019-06-17
Add serial output

SamTebbs33:
This patch adds serial output. Closes #37

PR #42, opened 2019-06-22
Add logging

### LINKED ISSUE #40: Logging
Logging to serial. Should support multiple levels (debug, warning, info and error). An example log message is `[DEBUG] This is the message`.

The logging interface should provide a function for each level which all call a generic function, and the generic function itself.

SamTebbs33:
This patch adds serial logging on top of the serial interface from #41. Closes #40. Depends on #41

### DIFF
--- a/src/kernel/arch/x86/arch.zig
+++ b/src/kernel/arch/x86/arch.zig
@@ -5,6 +5,7 @@ const gdt = @import("gdt.zig");
 const idt = @import("idt.zig");
 const irq = @import("irq.zig");
 const isr = @import("isr.zig");
+const log = @import("../../log.zig");
 
 pub const InterruptContext = struct {
     // Extra segments
--- a/src/kernel/log.zig
+++ b/src/kernel/log.zig
@@ -1,0 +1,31 @@
+const serial = @import("serial.zig");
+const fmt = @import("std").fmt;
+
+pub const Level = enum {
+    INFO,
+    DEBUG,
+    WARNING,
+    ERROR
+};
+
+fn logCallback(context: void, str: []const u8) anyerror!void {
+    serial.writeString(str, serial.Port.COM1);
+}
+
+pub fn log(comptime level: Level, comptime format: []const u8, args: ...) void {
+    fmt.format({}, anyerror, logCallback, "[" ++ @tagName(level) ++ "] " ++ format, args) catch unreachable;
+}
+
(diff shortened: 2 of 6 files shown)


src/kernel/log.zig
```diff
@@ -0,0 +1,37 @@
...
+    WARNING,
+    ERROR
+};
+
+pub fn init() void {
+    serial.init(serial.DEFAULT_BAUDRATE, serial.Port.COM1) catch unreachable;
+}
+
+fn logCallback(context: void, str: []const u8) anyerror!void {
+    for (str) |char| {
```
DrDeano: U have a writeStr in ur serial interface

src/kernel/log.zig
```diff
@@ -0,0 +1,37 @@
...
+
+pub const Level = enum {
+    INFO,
+    DEBUG,
+    WARNING,
+    ERROR
+};
+
+pub fn init() void {
+    serial.init(serial.DEFAULT_BAUDRATE, serial.Port.COM1) catch unreachable;
```
DrDeano: Its probs best to init the serial separately as other stuff may use the serial without the logging

src/kernel/log.zig
SamTebbs33: Indeed. This is from before that was added.

src/kernel/log.zig
SamTebbs33: Very true

src/kernel/arch/x86/arch.zig
```diff
@@ -43,7 +44,9 @@ pub const InterruptContext = struct {
 pub fn init() void {
     disableInterrupts();
 
+    log.log_info("Init gdt\n");
```
SamTebbs33: I'll move these log messages to the init function they apply to

src/kernel/log.zig
```diff
@@ -0,0 +1,37 @@
...
+    for (str) |char| {
+        serial.write(char, serial.Port.COM1);
+    }
+}
+
+pub fn log(comptime level: Level, comptime format: []const u8, args: ...) void {
+    fmt.format({}, anyerror, logCallback, "[" ++ @tagName(level) ++ "] " ++ format, args) catch unreachable;
+}
+
+pub fn log_info(comptime format: []const u8, args: ...) void {
```
SamTebbs33: I'll make these camel case to match the serial functions

PR #43, opened 2019-06-25
Added PIT interface

DrDeano:
Ready for PR

### DIFF
--- a/src/kernel/arch/x86/pic.zig
+++ b/src/kernel/arch/x86/pic.zig
@@ -167,7 +167,7 @@ inline fn readDataSlave() u8 {
     return arch.inb(SLAVE_DATA_REG);
 }
 
-fn readMasterIrr() u8 {
+inline fn readMasterIrr() u8 {
     sendCommandSlave(OCW3_DEFAULT | OCW3_ACT_ON_READ | OCW3_READ_IRR);
     return arch.inb(SLAVE_STATUS_REG);
 }
@@ -220,14 +220,16 @@ pub fn spuriousIrq(irq_num: u8) bool {
 }
 
 pub fn setMask(irq_num: u16) void {
-    const port: u16 = if (irq_num < 8) MASTER_COMMAND_REG else SLAVE_COMMAND_REG;
-    const value = arch.inb(port) | (1 << irq_num);
+    const port: u16 = if (irq_num < 8) MASTER_DATA_REG else SLAVE_DATA_REG;
+    const shift = @intCast(u3, irq_num % 8);
+    const value: u8 = arch.inb(port) | (u8(1) << shift);
     arch.outb(port, value);
 }
 
 pub fn clearMask(irq_num: u16) void {
-    const port: u16 = if (irq_num < 8) MASTER_COMMAND_REG else SLAVE_COMMAND_REG;
-    const value = arch.inb(port) & ~(1 << irq_num);
+    const port: u16 = if (irq_num < 8) MASTER_DATA_REG else SLAVE_DATA_REG;
+    const shift = @intCast(u3, irq_num % 8);
+    const value: u8 = arch.inb(port) & ~(u8(1) << shift);
     arch.outb(port, value);
 }
 
--- a/src/kernel/arch/x86/pit.zig
+++ b/src/kernel/arch/x86/pit.zig
@@ -1,0 +1,284 @@
+// Zig version: 0.4.0
+
+const arch = @import("arch.zig");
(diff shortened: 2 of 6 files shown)


src/kernel/arch/x86/pic.zig
```diff
@@ -220,14 +220,16 @@ pub fn spuriousIrq(irq_num: u8) bool {
 }
 
 pub fn setMask(irq_num: u16) void {
```
SamTebbs33: I probably should have mentioned this in the PIC PR, but could `clearMask` and `setMask` be merged since they are very similar?

src/kernel/arch/x86/pit.zig
```diff
@@ -0,0 +1,288 @@
...
+const OCW_SELECT_COUNTER_0: u8              = 0x00; // 00xxxxxx
+
+/// The OCW bits for selecting counter 1. Was for the memory refreshing.
+const OCW_SELECT_COUNTER_1: u8              = 0x40; // 01xxxxxx
+
+/// The OCW bits for selecting counter 2. Channel for the speaker.
+const OCW_SELECT_COUNTER_2: u8              = 0x80; // 10xxxxxx
+
+// The divisor constant
+// Reviewer: I have found 2 values for this, 1193182 and 1193180, and not sure which one is correct?
```
SamTebbs33: Hmm, since there's very little difference it's probably better to be conservative and choose the lower number.

src/kernel/arch/x86/pit.zig
```diff
@@ -0,0 +1,288 @@
...
+const MAX_FREQUENCY: u32 = 1193182;
+
+/// The number of tick that has passed when counter 0 was initially set up.
```
SamTebbs33: tick -> ticks

src/kernel/arch/x86/pit.zig
```diff
@@ -0,0 +1,288 @@
...
+    };
+
+    switch (counter) {
+        OCW_SELECT_COUNTER_0 => current_freq_0 = frequency,
+        OCW_SELECT_COUNTER_1 => current_freq_1 = frequency,
+        OCW_SELECT_COUNTER_2 => current_freq_2 = frequency,
+        else => unreachable,
+    }
+
+    var command: u8 = 0;
```
SamTebbs33: Not a big deal but you could assign it to `mode` instead of assigning it to 0 then or'ing with `mode`.

src/kernel/arch/x86/pit.zig
```diff
@@ -0,0 +1,288 @@
...
+///
+pub fn getFrequency() u32 {
+    return current_freq_0;
+}
+
+///
+/// Initialise the PIT with a handler to IRQ 0.
+///
+pub fn init() void {
+    // Set up counter 0 at 1000hz in a square wave mode counting in binary
```
SamTebbs33: If you rebase on top of the logging PR you could log something like "Init pit" (xD) and change the tty.print below to a log statement.

src/kernel/kmain.zig
```diff
@@ -14,18 +14,18 @@ const panic_root = @import("panic.zig");
 pub fn panic(msg: []const u8, error_return_trace: ?*builtin.StackTrace) noreturn {
     @setCold(true);
     arch.disableInterrupts();
+    asm volatile ("xchgw %%bx, %%bx");
```
SamTebbs33: What does this do?

DrDeano: This is a breakpoint for Bochs as i was getting panics in bochs and not in qemu, Will remove

src/kernel/arch/x86/pit.zig
DrDeano: Sure thing

src/kernel/arch/x86/pit.zig
DrDeano: Can do, less code

src/kernel/arch/x86/pit.zig
DrDeano: Ill go for the lower one :)

src/kernel/arch/x86/pic.zig
DrDeano: I have seen this as one function, but like to separate for a clear distinction instead of having another parameter for switching between the two. What do u think

SamTebbs33: Yeah on second thoughts having two is fine. An alternative would be to make them wrappers around a function that does take the extra parameter, but the compiler will probably extract the common code from them anyway.

PR #48, opened 2019-06-29
Add runtime testing harness

### LINKED ISSUE #12: Runtime testing framework
It would be useful to implement a system for testing the kernel at runtime. This would be used to test components of the kernel, such as paging, multitasking and the filesystem.

A possible approach to running a certain testcase would be to boot the kernel in testing mode (perhaps with a build specific for that testcase), have it initialise the component being tested and printing to serial with some predetermined format based on success or failure. A program could then be attached to the kernel's serial port ensuring that the correct output is received.

The results could then be compiled into a testing report.

Possible dependency on #37 if serial output is how we monitor test results.

SamTebbs33:
This patch adds a python script that acts as a runtime testing harness (test/rt-test.py), a new parameter to `zig build`, as well as some extra log messages for proper testing. The test harness script imports the arch-specific rt testing script from its arch directory.

It is recommended to run pkill on the qemu process after doing the testing as the process doesn't seem to be killed all the time: `zig build test -Drt-test=true; pkill qemu-system-i38`.

I would like to extend this to do more rigorous testing with more log statements once I find out how to tell the difference between an rt-test build and normal build from within the source, as adding more logging from a normal build would only pollute things.

Closes #12

### DIFF
--- a/test/rt-test.py
+++ b/test/rt-test.py
@@ -1,0 +1,64 @@
+import subprocess
+import signal
+import re
+import sys
+import datetime
+import os
+import importlib.util
+
+class TestCase:
+    def __init__(self, name, expected):
+        self.name = name
+        self.expected = expected
+
+def test_failure(case, exp, expected_idx, found):
+    print("FAILURE: %s #%d, expected '%s', found '%s'" %(case.name, expected_idx + 1, exp, found))
+    sys.exit(1)
+
+def test_pass(case, exp, expected_idx, found):
+    print("PASS: %s #%d, expected '%s', found '%s'" %(case.name, expected_idx + 1, exp, found))
+
+if __name__ == "__main__":
+    arch = sys.argv[1]
+    zig_path = sys.argv[2]
+    spec = importlib.util.spec_from_file_location("arch", "test/kernel/arch/" + arch + "/rt-test.py")
+    arch_module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(arch_module)
+
+    # The list of log statements to look for before arch init is called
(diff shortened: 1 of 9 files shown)


src/kernel/arch/x86/rt-test.py
```diff
@@ -0,0 +1,8 @@
+def getTestCases(TestCase):
+    return [
+            TestCase("GDT init", [r"Init gdt", r"Done"]),
+            TestCase("IDT init", [r"Init idt", r"Done"]),
+            TestCase("PIT init", [r"Init pit", r".+", "Done"]),
+            TestCase("Arch init finishes", [r"Arch init done"]),
+            TestCase("TTY init", [r"Init tty", r"Done"]),
```
SamTebbs33: This should be moved to the main rt-test.py file since it's arch-agnostic

test/rt-test.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # The list of log statements to look for before arch init is called
+    pre_archinit_cases = [
+            TestCase("Arch init starts", [r"Init arch \w+"])
+        ]
+    # The list of log statements to look for after arch init is called
+    post_archinit_cases = [
+            TestCase("Init finishes", [r"Init done"])
+        ]
+    # All log statements to look for include the arch-specific ones
```
SamTebbs33: s/include/including

SamTebbs33:
> lgtm

Thanks! I addressed my comments and moved `src/kernel/arch/x86/rt-test.py` to `test/kernel/arch/x86/rt-test.py` as the test code probably shouldn't be mixed with the Zig source.

PR #52, opened 2019-07-01
Add arch.registerInterruptHandler

### LINKED ISSUE #51: Add registerInterruptHandler function to arch interface
The arch interface should present a function that registers an interrupt handler so that arch-agnostic features (like system calls) can be implemented in an arch-agnostic way.

SamTebbs33:
This patch adds `registerIntHandler(...)` to arch.zig. Closes #51

### DIFF
--- a/src/kernel/arch/x86/arch.zig
+++ b/src/kernel/arch/x86/arch.zig
@@ -178,3 +178,14 @@ pub fn haltNoInterrupts() noreturn {
         halt();
     }
 }
+
+///
+/// Register an interrupt handler. The interrupt number should be the arch-specific number.
+///
+/// Arguments:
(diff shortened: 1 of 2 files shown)


src/kernel/arch/x86/arch.zig
```diff
@@ -178,3 +178,10 @@ pub fn haltNoInterrupts() noreturn {
         halt();
     }
 }
+
+///
+/// Register an interrupt handler. The interrupt number should be the arch-specific number.
+///
```
DrDeano: Could add the comments from arch_mock to here as this is the more important one

SamTebbs33: Good spot

SamTebbs33: Good spot

PR #54, opened 2019-07-19
Add continuous integration with Azure

### LINKED ISSUE #13: Continuous integration
It would be a good idea to set up continuous integration for the kernel, which would make sure that the codebase remains in a stable state. It could be used to ensure that each pull request can be built in a number of configurations (architecture etc.) and that all tests pass.

Criteria for a CI service for us to use (in order of priority) are:
1. Free: We're probably not in a situation to be able to pay for CI
2. Flexible configuration: Kernel development is a rather niche area and so not many CI services will support it out of the box

Relates to #12

SamTebbs33:
This patch adds continuous integration with Azure Pipelines at:
https://dev.azure.com/samueltebbs/pluto/_build?definitionId=1. This will
download zig v0.4.0 (will change to master once we upgrade our source),
build the kernel, install qemu and do runtime testing (mock testing will
also need to be done once that is up and running). Closes #13

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Pluto
 
+[![Build Status](https://dev.azure.com/samueltebbs/pluto/_apis/build/status/SamTebbs33.pluto?branchName=develop)](https://dev.azure.com/samueltebbs/samueltebbs/_build/latest?definitionId=1&branchName=develop)
+
 ## Build
 Requires *xorriso* and the grub tools (such as *grub-mkrescue*).
 ```
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,40 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
+# https://aka.ms/yaml
+
+trigger:
+- develop
+- feature/*
+
+pr:
+- develop
+- feature/*
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+variables:
+  zig_version: 0.4.0
+
+steps:
+- script: |
+    export PYTHONIOENCODING=utf8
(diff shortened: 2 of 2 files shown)


azure-pipelines.yml
```diff
@@ -0,0 +1,40 @@
...
+# Uncomment once mock testing is finished
+#- script: zig*/zig build test
+#  displayName: 'Mocked tests'
+
+- script: |
+    sudo apt-get update
+    sudo apt-get install qemu qemu-system --fix-missing
+  displayName: 'Download qemu'
+  
+- script: zig*/zig build test -Drt-test=true -Dzig-path=zig*/zig
```
DrDeano: This might be a big ask for this feature/issue/branch but could you add to your python script that will check for the zig tests being all 'OK' as well as the existing run time tests. So will run the zig tests and check each line is 'OK' and then run the run time tests

README.md
```diff
@@ -1,5 +1,7 @@
 # Pluto
 
+[![Build Status](https://dev.azure.com/samueltebbs/pluto/_apis/build/status/SamTebbs33.pluto?branchName=develop)](https://dev.azure.com/samueltebbs/samueltebbs/_build/latest?definitionId=1&branchName=develop)
+
```
DrDeano: Also I don't know if this is possible, but could this also provide a link to the built binaries so the person can just download and run without having to faf with building them self's if they don't want to

azure-pipelines.yml
SamTebbs33: I'm not sure if checking that the source file tests are all "OK" is necessary as Zig will exit with a non-zero code if any of them fail, causing the Azure step to fail. The same happens for the runtime tests.

README.md
SamTebbs33: Good idea, will investigate. It may be good to make this a separate issue though.

SamTebbs33:
Updated to fix the typo in the commit message :)

PR #56, opened 2019-08-03
Add x86 syscalls

### LINKED ISSUE #36: System calls
The kernel will need syscalls so that programs can interact with the kernel. We could choose the same interrupt number as [linux](https://blog.rchapman.org/posts/Linux_System_Call_Table_for_x86_64/), which would be 0x80. The number used to select which syscall is desired could be placed in `eax`, and the kernel will perform the appropriate action for that number. Parameters should be passed according to the architecture's ABI, so for x86 we'd follow the [cdecl](https://en.wikipedia.org/wiki/X86_calling_conventions#cdecl) ABI.

Depends on #51

SamTebbs33:
This patch adds syscalls for x86. They are triggered by putting the syscall number in `eax` then trriggering interrupt `128` (The same as linux). It adds support for calling syscalls with a maximum of 5 arguments. Closes #36

### DIFF
--- a/src/kernel/arch/x86/isr.zig
+++ b/src/kernel/arch/x86/isr.zig
@@ -3,6 +3,7 @@
 const panic = @import("../../panic.zig");
 const idt = @import("idt.zig");
 const arch = @import("arch.zig");
+const syscalls = @import("syscalls.zig");
 
 const NUMBER_OF_ENTRIES: u16 = 32;
 
@@ -39,6 +40,7 @@ extern fn isr28() void;
 extern fn isr29() void;
 extern fn isr30() void;
 extern fn isr31() void;
+extern fn isr128() void;
 
 /// The exception messaged that is printed when a exception happens
 const exception_msg: [NUMBER_OF_ENTRIES][]const u8 = [NUMBER_OF_ENTRIES][]const u8 {
@@ -76,8 +78,18 @@ const exception_msg: [NUMBER_OF_ENTRIES][]const u8 = [NUMBER_OF_ENTRIES][]const 
     "Reserved"
 };
 
+/// Errors that an isr function can return
+pub const IsrError = error {
+    UnrecognisedIsr
+};
+
+/// An isr handler. Takes an interrupt context and returns void.
+/// Should finish quickly to avoid delaying further interrupts and the previously running code
+pub const IsrHandler = fn(*arch.InterruptContext)void;
+
 // The of exception handlers initialised to unhandled.
-var isr_handlers: [NUMBER_OF_ENTRIES]fn(*arch.InterruptContext)void = []fn(*arch.InterruptContext)void{unhandled} ** NUMBER_OF_ENTRIES;
+var isr_handlers: [NUMBER_OF_ENTRIES]IsrHandler = []IsrHandler{unhandled} ** NUMBER_OF_ENTRIES;
+var syscall_handler: IsrHandler = unhandled;
 
 ///
 /// A dummy handler that will make a call to panic as it is a unhandled exception.
@@ -92,6 +104,17 @@ fn unhandled(context: *arch.InterruptContext) void {
 }
 
 ///
+/// Checks if the isr is valid and returns true if it is, else false.
+/// To be valid it must be greater than or equal to 0 and less than NUMBER_OF_ENTRIES.
+///
+/// Arguments:
+///     IN isr_num: u16 - The isr number to check
+///
+pub fn isValidIsr(isr_num: u32) bool {
+    return isr_num >= 0 and isr_num < NUMBER_OF_ENTRIES;
+}
+
+///
 /// The exception handler that each of the exceptions will call when a exception happens.
 ///
 /// Arguments:
@@ -100,7 +123,13 @@ fn unhandled(context: *arch.InterruptContext) void {
 ///
 export fn isrHandler(context: *arch.InterruptContext) void {
     const isr_num = context.int_num;
-    isr_handlers[isr_num](context);
+    if (isr_num == syscalls.INTERRUPT) {
+        syscall_handler(context);
+    } else if (isValidIsr(isr_num)) {
+        isr_handlers[isr_num](context);
+    } else {
+        panic.panicFmt(null, "Unrecognised isr: {}\n", isr_num);
+    }
 }
 
 ///
@@ -109,8 +138,17 @@ export fn isrHandler(context: *arch.InterruptContext) void {
 /// Arguments:
 ///     IN irq_num: u16 - The exception number to register.
 ///
-pub fn registerIsr(isr_num: u16, handler: fn(*arch.InterruptContext)void) void {
-    isr_handlers[isr_num] = handler;
+/// Errors:
+///     IsrError.UnrecognisedIsr - If `isr_num` is invalid (see isValidIsr)
+///
+pub fn registerIsr(isr_num: u16, handler: fn(*arch.InterruptContext)void) !void {
+    if (isr_num == syscalls.INTERRUPT) {
+        syscall_handler = handler;
+    } else if (isValidIsr(isr_num)) {
+        isr_handlers[isr_num] = handler;
+    } else {
+        return IsrError.UnrecognisedIsr;
+    }
 }
 
 ///
@@ -160,4 +198,5 @@ pub fn init() void {
     idt.openInterruptGate(29, isr29);
     idt.openInterruptGate(30, isr30);
     idt.openInterruptGate(31, isr31);
+    idt.openInterruptGate(syscalls.INTERRUPT, isr128);
 }
--- a/src/kernel/arch/x86/paging.zig
+++ b/src/kernel/arch/x86/paging.zig
@@ -142,7 +142,7 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void 
     mapDir(kernel_directory, p_start, p_end, v_start, v_end, allocator) catch unreachable;
     const dir_physaddr = @ptrToInt(kernel_directory) - constants.KERNEL_ADDR_OFFSET;
     asm volatile ("mov %[addr], %%cr3" :: [addr] "{eax}" (dir_physaddr));
-    isr.registerIsr(14, pageFault);
+    isr.registerIsr(14, pageFault) catch unreachable;
 }
 
 test "isAligned" {
--- a/src/kernel/arch/x86/syscalls.zig
+++ b/src/kernel/arch/x86/syscalls.zig
@@ -1,0 +1,287 @@
+const arch = @import("arch.zig");
+const testing = @import("std").testing;
+const assert = @import("std").debug.assert;
+const isr = @import("isr.zig");
+const log = @import("../../log.zig");
+
+/// The isr number associated with syscalls
+pub const INTERRUPT: u16 = 0x80;
+
+/// The maximum number of syscall handlers that can be registered
+pub const NUM_HANDLERS: u16 = 256;
+
+/// A syscall handler
+pub const SyscallHandler = fn (ctx: *arch.InterruptContext) u32;
+
+/// Errors that syscall utility functions can throw
+pub const SyscallError = error {
+    SyscallExists,
+    InvalidSyscall
+};
+
+/// The array of registered syscalls
+var handlers: [NUM_HANDLERS]?SyscallHandler = []?SyscallHandler{null} ** NUM_HANDLERS;
+
+///
+/// Returns true if the syscall is valid, else false.
+/// A syscall is valid if it's less than NUM_HANDLERS.
+///
+/// Arguments:
+///     IN syscall: u32 - The syscall to check
+///
+pub fn isValidSyscall(syscall: u32) bool {
+    return syscall < NUM_HANDLERS;
+}
+
+///
+/// Handle a syscall. Gets the syscall number from eax within the context and calls the registered handler.
+/// If there isn't a registered handler or the syscall is invalid (>= NUM_HANDLERS) then a warning is logged.
+///
+/// Arguments:
+///     IN ctx: *arch.InterruptContext - The cpu context when the syscall was triggered. The syscall number is
+///                                      stored in eax.
+///
+fn handle(ctx: *arch.InterruptContext) void {
+    // The syscall number is put in eax
+    const syscall = ctx.eax;
+    if (isValidSyscall(syscall)) {
+        if (handlers[syscall]) |handler| {
+            ctx.eax = handler(ctx);
+        } else {
+            log.logWarning("Syscall {} triggered but not registered\n", syscall);
+        }
+    } else {
+        log.logWarning("Syscall {} is invalid\n", syscall);
+    }
+}
+
+///
+/// Register a syscall so it can be called by triggering interrupt 128.
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall number to register the handler with.
+///     IN handler: SyscallHandler - The handler to register the syscall with.
+///
+/// Errors:
+///     SyscallError.InvalidSyscall - If the syscall is invalid (see isValidSyscall).
+///     SyscallError.SyscallExists - If the syscall has already been registered.
+///
+pub fn registerSyscall(syscall: u8, handler: SyscallHandler) SyscallError!void {
+    if (!isValidSyscall(syscall))
+        return SyscallError.InvalidSyscall;
+    if (handlers[syscall]) |_|
+        return SyscallError.SyscallExists;
+    handlers[syscall] = handler;
+}
+
+///
+/// Trigger a syscall with no arguments. Returns the value put in eax by the syscall.
+///
+/// Arguments:
+///     IN syscall: u32 - The syscall to trigger, put in eax.
+///
+inline fn syscall0(syscall: u32) u32 {
+    return asm volatile (
+        \\int $0x80
+        : [ret] "={eax}" (-> u32)
+        : [syscall] "{eax}" (syscall),
+    );
+}
+
+///
+/// Trigger a syscall with one argument. Returns the value put in eax by the syscall.
+///
+/// Arguments:
+///     IN syscall: u32 - The syscall to trigger, put in eax.
+///     IN arg: u32 - The argument to pass. Put in ebx.
+///
+inline fn syscall1(syscall: u32, arg: u32) u32 {
+    return asm volatile (
+        \\int $0x80
+        : [ret] "={eax}" (-> u32)
+        : [syscall] "{eax}" (syscall),
+          [arg1] "{ebx}" (arg),
+    );
+}
+
+///
+/// Trigger a syscall with two arguments. Returns the value put in eax by the syscall.
+///
+/// Arguments:
+///     IN syscall: u32 - The syscall to trigger, put in eax.
+///     IN arg1: u32 - The first argument to pass. Put in ebx.
+///     IN arg2: u32 - The second argument to pass. Put in ecx.
+///
+inline fn syscall2(syscall: u32, arg1: u32, arg2: u32) u32 {
+    return asm volatile (
+        \\int $0x80
+        : [ret] "={eax}" (-> u32)
+        : [syscall] "{eax}" (syscall),
+          [arg1] "{ebx}" (arg1),
+          [arg2] "{ecx}" (arg2),
+    );
+}
+
+///
+/// Trigger a syscall with three arguments. Returns the value put in eax by the syscall.
+///
+/// Arguments:
+///     IN syscall: u32 - The syscall to trigger, put in eax.
+///     IN arg1: u32 - The first argument to pass. Put in ebx.
+///     IN arg2: u32 - The second argument to pass. Put in ecx.
+///     IN arg3: u32 - The third argument to pass. Put in edx.
+///
+inline fn syscall3(syscall: u32, arg1: u32, arg2: u32, arg3: u32) u32 {
+    return asm volatile (
+        \\int $0x80
+        : [ret] "={eax}" (-> u32)
+        : [syscall] "{eax}" (syscall),
+          [arg1] "{ebx}" (arg1),
+          [arg2] "{ecx}" (arg2),
+          [arg3] "{edx}" (arg3),
+    );
+}
+
+///
+/// Trigger a syscall with four arguments. Returns the value put in eax by the syscall.
+///
+/// Arguments:
+///     IN syscall: u32 - The syscall to trigger, put in eax.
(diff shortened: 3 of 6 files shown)


src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
+const arch = @import("arch.zig");
+const testing = @import("std").testing;
+const assert = @import("std").debug.assert;
+const isr = @import("isr.zig");
+const log = @import("../../log.zig");
+
+pub const INTERRUPT: u16 = 0x80;
```
DrDeano: Could comment public constants

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+/// A syscall handler
+pub const SyscallHandler = fn (ctx: *arch.InterruptContext) void;
+/// Errors that syscall utility functions can throw
+pub const SyscallError = error {
+    SyscallExists
+};
+
+var handlers = []?SyscallHandler{null} ** NUM_HANDLERS;
+
+fn handle(ctx: *arch.InterruptContext) void {
```
DrDeano: Doc comment

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+
+///
+/// Register a syscall so it  can be called by triggering interrupt 128
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall number to register the handler with
+///     IN handler: SyscallHandler - The handler to register the syscall with
+///
+pub fn registerSyscall(syscall: u8, handler: SyscallHandler) !void {
+    if (handlers[syscall]) |_|
```
DrDeano: Not sure if this should be an error if wanting to re-register a syscall, but i guess u'll only be registering one syscall anyways, just a though if there is a need to re-register. Also could add a bounds check for the syscall number.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall to trigger. Is put in eax and must have been registered.
+///
+pub fn doSyscall(syscall: u8) void {
+    asm volatile ("mov %%ecx, %%eax" :: [syscall] "{ecx}" (syscall));
+    asm volatile ("int $0x80");
+}
+
+pub fn init() void {
```
DrDeano: Doc comment

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+
+///
+/// Trigger a syscall
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall to trigger. Is put in eax and must have been registered.
+///
+pub fn doSyscall(syscall: u8) void {
+    asm volatile ("mov %%ecx, %%eax" :: [syscall] "{ecx}" (syscall));
+    asm volatile ("int $0x80");
```
DrDeano: This can be done in one asm volatile statement. Also what if the syscall requires arguments. see: https://ziglang.org/documentation/0.4.0/#Assembly

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+fn testHandler(ctx: *arch.InterruptContext) void {
+}
+
+test "registerSyscall returns SyscallExists" {
+    registerSyscall(123, testHandler) catch unreachable;
+    registerSyscall(123, testHandler) catch |err| {
+        return;
+    };
+    assert(false);
+}
```
DrDeano: Could add tests for calling a syscall with ur testHandler

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+        if (handlers[syscall]) |handler| {
+            handler(ctx);
+        } else {
+            log.logWarning("Syscall {} triggered but not registered\n", syscall);
+        }
+    }
+}
+
+///
+/// Register a syscall so it  can be called by triggering interrupt 128
```
DrDeano: Have two spaces :P

src/kernel/arch/x86/isr.zig
```diff
@@ -78,6 +80,7 @@ const exception_msg: [NUMBER_OF_ENTRIES][]const u8 = [NUMBER_OF_ENTRIES][]const
 
 // The of exception handlers initialised to unhandled.
 var isr_handlers: [NUMBER_OF_ENTRIES]fn(*arch.InterruptContext)void = []fn(*arch.InterruptContext)void{unhandled} ** NUMBER_OF_ENTRIES;
+var syscall_handler = unhandled;
```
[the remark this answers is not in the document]
SamTebbs33: Will do. I'll add an `IsrHandler` type as well so the `isr_handlers` array and this share a type.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
+const arch = @import("arch.zig");
+const testing = @import("std").testing;
+const assert = @import("std").debug.assert;
+const isr = @import("isr.zig");
+const log = @import("../../log.zig");
+
+pub const INTERRUPT: u16 = 0x80;
+pub const NUM_HANDLERS = 256;
```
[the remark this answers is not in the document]
SamTebbs33: It being public means that code registering a syscall knows what the upper bound is.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+
+/// A syscall handler
+pub const SyscallHandler = fn (ctx: *arch.InterruptContext) void;
```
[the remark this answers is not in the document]
SamTebbs33: Exposing this type would be useful for code registering a syscall so they know what to provide.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,70 @@
...
+/// Errors that syscall utility functions can throw
+pub const SyscallError = error {
+    SyscallExists
+};
+
+var handlers = []?SyscallHandler{null} ** NUM_HANDLERS;
```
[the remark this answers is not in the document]
SamTebbs33: Is panicking appropriate? It's not like an exception where not handling an exception could be fatal. Triggering an unregistered syscall just means that some process won't have some functionality performed for it, which a warning is probably enough for. Panicking could introduce a problem where a malicious process would just trigger an unhandled syscall to bring the system to a halt.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Will add a bounds check. We probably don't want to allow re-registering syscalls as then a malicious process could override some basic syscall to run its own code instead.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Good idea. Do you know how to pass constants as arguments to the assembly? It would nice to not have to hardcode the `0x80`.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: I'll be able to do so once I figure out how to pass options from build.zig to the kernel to differentiate between a test build and a notmal build, as these tests would have to be done at runtime (since they depend on the CPU doing interrupts) and we don't want to be doing them for a normal build.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Oopsy

src/kernel/arch/x86/syscalls.zig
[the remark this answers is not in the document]
DrDeano: that fine, just bc nothing was using it out side this for the moment

src/kernel/arch/x86/syscalls.zig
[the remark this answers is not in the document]
DrDeano: yeh, makes sense, i was in a panicing mood lol

src/kernel/arch/x86/syscalls.zig
DrDeano: someone has got there security hat on :), i guess i can do the same with the irq handlers as well, as dont want malicious code to be re-registering irq's

src/kernel/arch/x86/syscalls.zig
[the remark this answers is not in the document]
SamTebbs33: True :) It does make the API nice-ish for the future

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Opening an issue for that would be good actually, so we don't forget

src/kernel/arch/x86/syscalls.zig
DrDeano: Haven't looked into this, but i do think u can.

SamTebbs33: Regarding the arguments, either the syscall handler should pop off the arguments itself or we can add some functionality here once the requirements are clear, or maybe you think that should be added in this PR.

SamTebbs33:
> I was wondering, is it possible to have a run time test for calling a fake syscall, could just have something hard coded for now, but then could extend this out to a proper runtime testing that can call test functions. So could pass a boolean into the build of the kernel which it can check

Yeah I could add a test that always runs when the kernel boots, and later move it to a proper runtime test when we figure out how to pass options to the build.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,105 @@
...
+    handlers[syscall] = handler;
+}
+
+///
+/// Trigger a syscall
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall to trigger. Is put in eax and must have been registered.
+///
+pub fn doSyscall(syscall: u8) void {
```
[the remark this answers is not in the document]
SamTebbs33: Yeah during the design process I realised that syscalls are too different on each architecture for it to be added to the standard arch.zig in a nice way.

SamTebbs33:
> > I was wondering, is it possible to have a run time test for calling a fake syscall, could just have something hard coded for now, but then could extend this out to a proper runtime testing that can call test functions. So could pass a boolean into the build of the kernel which it can check
> 
> Yeah I could add a test that always runs when the kernel boots, and later move it to a proper runtime test when we figure out how to pass options to the build.

I've actually found out how to pass options to the build so I will make a PR for that and add a syscall runtime test afterwards.

SamTebbs33:
I actually don't think this is quite ready to be merged yet since I need to do some more investigation on how syscall arguments should be handled.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: I decided to add them

SamTebbs33:
Ok I did some experimentation and made the syscall triggering much better and able to pass arguments. Ready for a review!

src/kernel/arch/x86/paging.zig
```diff
@@ -142,7 +142,7 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void
     mapDir(kernel_directory, p_start, p_end, v_start, v_end, allocator) catch unreachable;
     const dir_physaddr = @ptrToInt(kernel_directory) - constants.KERNEL_ADDR_OFFSET;
     asm volatile ("mov %[addr], %%cr3" :: [addr] "{eax}" (dir_physaddr));
-    isr.registerIsr(14, pageFault);
+    isr.registerIsr(14, pageFault) catch unreachable;
```
DrDeano: Magic number. I'll create a ticket for ISR numbers.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,173 @@
...
+}
+
+///
+/// Trigger a syscall. Returns the value put in eax by the syscall handler.
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall to trigger. Is put in eax and must have been registered.
+///     IN args: u32... - The args to send to the syscall. A maximum of 5 are put in ebx, ecx, edx, esi and edi
+///
+pub inline fn doSyscall(comptime syscall: u32, args: ...) u32 {
```
DrDeano: I guess this is the point the user land will make a syscall. If so, could this be moved to another file or clearly separate this out so it is clear that this is what the user will call and isn't part of the kernel code.

DrDeano: Also, is it worth adding a check for valid syscalls or is it fine to use the check after the syscall

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,173 @@
...
+    return syscall < NUM_HANDLERS;
+}
+
+///
+/// Handle a syscall. Gets the syscall number from eax within the context and calls the registered handler.
+/// If there isn't a registered handler or the syscall is invalid (>= NUM_HANDLERS) then a warning is logged
+///
+/// Arguments:
+///     IN ctx: *arch.InterruptContext - The cpu context when the syscall was triggered. The syscall number is
+///                                     stored in eax.
```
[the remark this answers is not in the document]
SamTebbs33: Oops

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Hmm good question, reckon we should return an error if it's an invalid syscall? How would you suggest we separate it?

SamTebbs33: Hmm good question, reckon we should return an error if it's an invalid syscall?

SamTebbs33:
> Also, wheres those runtime tests :P

Waiting for the build options to be merged :p I actually need to improve that patch a little. Will do so when I can.

DrDeano:
No problem, shall I'll wait until the build options PR is done

src/kernel/arch/x86/syscalls.zig
SamTebbs33: What would be nice later on is to add this kind of stuff to our own fork of Zig stdlib. That should probably be the final aim.

SamTebbs33:
> No problem, shall I'll wait until the build options PR is done

Alright that one should be ready now. I added some replies here too

src/kernel/arch/x86/syscalls.zig
DrDeano: Yeh, maybe the same error for the kernel one? Or a separate error

DrDeano: That would be cool. I guess this would be part of the OSs API so a separate file would be good, u can come up with the name ;)

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+          [arg5] "{edi}" (arg5),
+    );
+}
+
+///
+/// Gets the syscall argumentaccording to the given number. 0 => ebx, 1 => ecx, 2 => edx, 3 => esi and 4 => edi
+///
+/// Arguments:
+///     IN ctx: *arch.InterruptContext - The interrupt context from which to get the argument
+///     IN num: comptime u32 - The arg number to get. Between 0 and 4.
```
DrDeano: Sorry for being anal, but num could have a better name like arg_index

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+        0 => ctx.ebx,
+        1 => ctx.ecx,
+        2 => ctx.edx,
+        3 => ctx.esi,
+        4 => ctx.edi,
+        else => @compileError("Arg number must be between 0 and 4")
+    };
+}
+
+pub inline fn syscallArg2(ctx: *arch.InterruptContext) u32 {
```
DrDeano: Where syscallArg1? Also as you have syscallArg where you can index all the argument, do you need all these functions?

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+          [arg1] "{ebx}" (arg1),
+          [arg2] "{ecx}" (arg2),
+          [arg3] "{edx}" (arg3),
+          [arg4] "{esi}" (arg4),
+          [arg5] "{edi}" (arg5),
+    );
+}
+
+///
+/// Gets the syscall argumentaccording to the given number. 0 => ebx, 1 => ecx, 2 => edx, 3 => esi and 4 => edi
```
DrDeano: Missing space

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+///
+/// Arguments:
+///     IN syscall: u8 - The syscall number to register the handler with
+///     IN handler: SyscallHandler - The handler to register the syscall with
+///
+/// Errors:
+///     SyscallError.InvalidSyscall - If the syscall is invalid (see isValidSyscall)
+///     SyscallError.SyscallExists - If the syscall has already been registered
+///
+pub fn registerSyscall(syscall: u8, handler: SyscallHandler) !void {
```
DrDeano: As we talked about, the the return should be SyscallError!void

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+}
+
+///
+/// Gets the syscall argumentaccording to the given number. 0 => ebx, 1 => ecx, 2 => edx, 3 => esi and 4 => edi
+///
+/// Arguments:
+///     IN ctx: *arch.InterruptContext - The interrupt context from which to get the argument
+///     IN num: comptime u32 - The arg number to get. Between 0 and 4.
+///
+pub inline fn syscallArg(ctx: *arch.InterruptContext, comptime num: u32) u32 {
```
DrDeano: Not sure this need to be pub

src/kernel/arch/x86/isr.zig
```diff
@@ -109,8 +138,17 @@ export fn isrHandler(context: *arch.InterruptContext) void {
 /// Arguments:
 ///     IN irq_num: u16 - The exception number to register.
 ///
-pub fn registerIsr(isr_num: u16, handler: fn(*arch.InterruptContext)void) void {
-    isr_handlers[isr_num] = handler;
+/// Errors:
+///     IsrError.UnrecognisedIsr - If `isr_num` is invalid (see isValidIsr)
+///
+pub fn registerIsr(isr_num: u16, handler: fn(*arch.InterruptContext)void) !void {
```
DrDeano: IsrError!void

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
+const arch = @import("arch.zig");
+const testing = @import("std").testing;
+const assert = @import("std").debug.assert;
+const isr = @import("isr.zig");
+const log = @import("../../log.zig");
+
+pub const INTERRUPT: u16 = 0x80;
```
DrDeano: Doc comments

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+pub const NUM_HANDLERS: u16 = 256;
+
+/// A syscall handler
+pub const SyscallHandler = fn (ctx: *arch.InterruptContext) u32;
+/// Errors that syscall utility functions can throw
```
DrDeano: Missing new line. (I am really anal on formatting)

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+const testing = @import("std").testing;
+const assert = @import("std").debug.assert;
+const isr = @import("isr.zig");
+const log = @import("../../log.zig");
+
+pub const INTERRUPT: u16 = 0x80;
+pub const NUM_HANDLERS: u16 = 256;
+
+/// A syscall handler
+pub const SyscallHandler = fn (ctx: *arch.InterruptContext) u32;
```
DrDeano: I have seen you are a bit inconsistent in the formatting of types where you have spaces and sometimes not. We will need to add this to the formatting rules. Personally, I would go with having the spaces, easier to read.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+///
+pub fn registerSyscall(syscall: u8, handler: SyscallHandler) !void {
+    if (!isValidSyscall(syscall))
+        return SyscallError.InvalidSyscall;
+    if (handlers[syscall]) |_|
+        return SyscallError.SyscallExists;
+    handlers[syscall] = handler;
+}
+
+inline fn syscall0(syscall: u32) u32 {
```
DrDeano: Might seem a bit much, but doc comment? You can say no :P

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+/// Arguments:
+///     IN syscall: u32 - The syscall to check
+///
+pub fn isValidSyscall(syscall: u32) bool {
+    return syscall < NUM_HANDLERS;
+}
+
+///
+/// Handle a syscall. Gets the syscall number from eax within the context and calls the registered handler.
+/// If there isn't a registered handler or the syscall is invalid (>= NUM_HANDLERS) then a warning is logged
```
DrDeano: Full stop at the end

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Ah thanks for spotting this. I wrote `syscallArgX` before writing the general one and forgot to remove the old ones.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Hmm, yeah I agree on the spaces. Let me know where I've missed spaces before and I'll fix it. Zig fmt will probably notice it though :)

src/kernel/arch/x86/syscalls.zig
```diff
@@ -0,0 +1,248 @@
...
+    registerSyscall(125, testHandler2) catch unreachable;
+    registerSyscall(126, testHandler3) catch unreachable;
+    registerSyscall(127, testHandler4) catch unreachable;
+    registerSyscall(128, testHandler5) catch unreachable;
+    assert(testInt == 0);
+
+    if (syscall0(123) == 0 and testInt == 1)
+        log.logInfo("Syscalls: Tested no args\n");
+
+    if (syscall1(124, (2)) == 1 and testInt == 3)
```
SamTebbs33: I have lots of int literals in unnecessary parentheses here.

Issue #60, opened 2019-08-27
Add some extensive x86 runtime tests


SamTebbs33:
Once #59 is done, we can add some extensive runtime tests for x86 that we wouldn't want to run in a normal build.

- [x] [Syscalls](https://github.com/SamTebbs33/pluto/pull/56)
- [x] [Paging](https://github.com/SamTebbs33/pluto/pull/73)
- [x] [IRQs](https://github.com/SamTebbs33/pluto/pull/92) and [ISRs](https://github.com/SamTebbs33/pluto/pull/93)
- [x] [PIC](https://github.com/SamTebbs33/pluto/pull/94)
- [x] PIT

This work could be split as the person who implemented the features above is probably best suited to testing it.

Depends on #59

SamTebbs33:
It would also be good to add a TEST log level to differentiate them from normal log messages.

DrDeano:
I was thinking of this, but will need to hide this from the real build. Maybe add it to the log_mock.zig once #45 is done. But idk if added a TEST log level will be necessary as you can just use INFO or DEBUG.

SamTebbs33:
> I was thinking of this, but will need to hide this from the real build.

As in only expose the TEST level to code if we're doing an rt test?

> Maybe add it to the log_mock.zig once #45 is done. But idk if added a TEST log level will be necessary as you can just use INFO or DEBUG.

That's true, we could always add one later on if we feel the need to differentiate. My only thinking was that it would make scanning through logs easier.

DrDeano:
Yes, only expose TEST to rt_test, but will add a lot of testing logs to the code base.
It would be easier maybe but will have to change all logging to include testing logs where could just use DEBUG.

SamTebbs33:
I forgot about the PIT

PR #62, opened 2019-08-27
Add infrastructure for build options

### LINKED ISSUE #59: Pass options from build.zig to source
Options can be passed from build.zig to any zig source file. This can be used for configuration and creating builds for a specific purpose, such as running extensive runtime tests.

The option is added by calling `addBuildOption(...)` on a `LibObjExeStep` (e.g. the result from `addObject(...)`, then used by importing zig-cache/{object name}_build_options.zig.

SamTebbs33:
This patch adds the infrastructure for passing build options from build.zig to the source. At the moment the only option added is `rt_test` which will be used to run extensive runtime tests (#60). 

I added the config.zig file to hide the rather ugly build options file import from kmain.zig, much like we did with arch.zig.

Closes #59

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,5 +36,7 @@ steps:
     sudo apt-get install qemu qemu-system --fix-missing
   displayName: 'Download qemu'
   
-- script: zig*/zig build test -Drt-test=true -Dzig-path=zig*/zig
+- script: |
+    zig*/zig build -Drt-test=true
+    zig*/zig build test -Drt-test=true -Dzig-path=zig*/zig
   displayName: 'Runtime tests'
--- a/build.zig
+++ b/build.zig
@@ -37,6 +37,7 @@ pub fn build(b: *Builder) void {
     b.makePath(grub_path.toSlice()) catch unreachable;
     b.makePath(kern_path.toSlice()) catch unreachable;
     b.makePath(a_path.toSlice()) catch unreachable;
+    b.makePath("zig-cache/kernel") catch unreachable;
 
     src_files.append("kernel/kmain") catch unreachable;
 
@@ -54,11 +55,17 @@ pub fn build(b: *Builder) void {
     src_files.append(arch_boot.toSlice()) catch unreachable;
 
     const iso_path = concat(b.allocator, build_path, "/pluto.iso") catch unreachable;
-    var objects_steps = buildObjects(b, builtin_target, build_path, src_path);
+    var objects = buildObjects(b, builtin_target, build_path, src_path);
     var link_step = buildLink(b, builtin_target, build_path);
     const iso_step = buildISO(b, build_path, iso_path.toSlice());
 
-    for (objects_steps.toSlice()) |step| b.default_step.dependOn(step);
+    for (objects.toSlice()) |obj| {
+        if (std.mem.eql(u8, obj.name, "kernel/kmain")) {
+            // Add build options here
+            obj.addBuildOption(bool, "rt_test", rt_test);
+        }
+        b.default_step.dependOn(&obj.step);
+    }
     b.default_step.dependOn(link_step);
     for (iso_step.toSlice()) |step| b.default_step.dependOn(step);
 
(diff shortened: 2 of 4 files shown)


src/config.zig
```diff
@@ -0,0 +1 @@
+pub const options = @import("../zig-cache/config_build_options.zig");
```
DrDeano: Could this be made into a export const as so you wont have to pass an additional argument into every function that needs it, just extern it.

DrDeano: So expand all the build options and export them with doc comments explaining what they do

src/kernel/arch/x86/arch.zig
```diff
@@ -45,7 +45,7 @@ const MemProfile = @import("../../mem.zig").MemProfile;
 ///
 /// Initialise the architecture
 ///
-pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void {
+pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator, rt_test: bool) void {
```
DrDeano: So add a extern to the rt_test so don't need an additional parameter

build.zig
```diff
@@ -38,6 +38,7 @@ pub fn build(b: *Builder) void {
     b.makePath(kern_path.toSlice()) catch unreachable;
     b.makePath(a_path.toSlice()) catch unreachable;
 
+    src_files.append("config") catch unreachable;
```
DrDeano: I'm going to be wary as before we had the issue of double functions and variables, As config.zig doesn't use any other functions this shouldn't be a problem. But once someone does, this could cause the same issue. So may need to come up with a better solution for adding additional files.
You do import this file into kmain.zig, so does this need to be here?

azure-pipelines.yml
```diff
@@ -36,5 +36,7 @@ steps:
     sudo apt-get install qemu qemu-system --fix-missing
   displayName: 'Download qemu'
   
-- script: zig*/zig build test -Drt-test=true -Dzig-path=zig*/zig
+- script: |
```
DrDeano: Surely you would also have a normal build as well, as adding the testing and setting to true would have the OS be in testing mode when released. Then with the build OS with runtime testing enabled, add a command to start the kernel which would trigger the runtime tests

src/kernel/arch/x86/arch.zig
DrDeano: Also, if for ever reason need to add more build options, then these parameters could get very long

src/config.zig
SamTebbs33: I think expanding them is a good idea. What's the benefit of using `extern` and `export` rather than just doing `config.rt_test`?

src/kernel/arch/x86/arch.zig
SamTebbs33: How about passing the config struct instead of passing each individual option? I think that would be better than the current implementation at least.

build.zig
SamTebbs33: That's a good point. There were some reasons to do with build.zig why I added the extra file but with some extra experimentation I've realised how to overcome them. I'll remove config.zig and import the options directly into kmain.zig.

azure-pipelines.yml
SamTebbs33: There is actually a normal build without runtime testing enabled earlier on in the pipeline, check the "Build kernel" step.

src/kernel/arch/x86/arch.zig
DrDeano: yeh, that would be fine

src/config.zig
DrDeano: Yeh, pass the config struct would be good

src/kernel/kmain.zig
```diff
@@ -9,6 +9,7 @@ const vga = @import("vga.zig");
 const log = @import("log.zig");
 const serial = @import("serial.zig");
 const mem = @import("mem.zig");
+const options = @import("build_options");
```
DrDeano: This is cool, how can you just pass in this string like that?

SamTebbs33: I think that it's a hardcoded thing in the compiler and it just finds the right build options file :)

PR #73, opened 2019-09-09
Add x86 paging rt test

SamTebbs33:
Adds a runtime test to x86 paging which ensures that unmapped memory causes a page fault. If you can think of any other tests I could add please say so. Part of #60

### DIFF
--- a/src/kernel/arch/x86/paging.zig
+++ b/src/kernel/arch/x86/paging.zig
@@ -7,6 +7,9 @@ const arch = @import("arch.zig");
 const isr = @import("isr.zig");
 const MemProfile = @import("../../mem.zig").MemProfile;
 const tty = @import("../../tty.zig");
+const log = @import("../../log.zig");
+const options = @import("build_options");
+const testing = std.testing;
 
 /// An array of directory entries and page tables. Forms the first level of paging and covers the entire 4GB memory space.
 const Directory = packed struct {
@@ -271,16 +274,16 @@ fn mapTableEntry(entry: *align(1) TableEntry, phys_addr: usize) PagingError!void
 ///
 /// Arguments:
 ///     OUT entry: *Directory - The directory to map
-///     IN phys_start: usize - The physical address at which to start mapping
-///     IN phys_end: usize - The physical address at which to stop mapping
 ///     IN virt_start: usize - The virtual address at which to start mapping
 ///     IN virt_end: usize - The virtual address at which to stop mapping
+///     IN phys_start: usize - The physical address at which to start mapping
+///     IN phys_end: usize - The physical address at which to stop mapping
 ///     IN allocator: *Allocator - The allocator to use to map any tables needed
 ///
 /// Error: std.mem.allocator.Error || PagingError
 ///     * - See mapDirEntry.
 ///
-fn mapDir(dir: *Directory, phys_start: usize, phys_end: usize, virt_start: usize, virt_end: usize, allocator: *std.mem.Allocator) (std.mem.Allocator.Error || PagingError)!void {
+fn mapDir(dir: *Directory, virt_start: usize, virt_end: usize, phys_start: usize, phys_end: usize, allocator: *std.mem.Allocator) (std.mem.Allocator.Error || PagingError)!void {
     var virt_addr = virt_start;
     var phys_addr = phys_start;
     var page = virt_addr / PAGE_SIZE_4KB;
@@ -312,6 +315,7 @@ fn pageFault(state: *arch.InterruptContext) void {
 ///     IN allocator: *std.mem.Allocator - The allocator to use
 ///
 pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void {
+    log.logInfo("Init paging\n");
     ADDR_OFFSET = @ptrToInt(&KERNEL_ADDR_OFFSET);
     // Calculate start and end of mapping
     const v_start = std.mem.alignBackward(@ptrToInt(mem_profile.vaddr_start), PAGE_SIZE_4KB);
@@ -319,18 +323,24 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void 
     const p_start = std.mem.alignBackward(@ptrToInt(mem_profile.physaddr_start), PAGE_SIZE_4KB);
     const p_end = std.mem.alignForward(@ptrToInt(mem_profile.physaddr_end) + mem_profile.fixed_alloc_size, PAGE_SIZE_4KB);
 
-    var tmp = allocator.alignedAlloc(Directory, @truncate(u29, PAGE_SIZE_4KB), 1) catch panic(@errorReturnTrace(), "Failed to allocate page directory");
+    var tmp = allocator.alignedAlloc(Directory, @truncate(u29, PAGE_SIZE_4KB), 1) catch |e| {
+        panic(@errorReturnTrace(), "Failed to allocate page directory: {}\n", e);
+    };
     var kernel_directory = @ptrCast(*Directory, tmp.ptr);
     @memset(@ptrCast([*]u8, kernel_directory), 0, @sizeOf(Directory));
 
     // Map in kernel
-    mapDir(kernel_directory, p_start, p_end, v_start, v_end, allocator) catch panic(@errorReturnTrace(), "Failed to map kernel directory");
+    mapDir(kernel_directory, v_start, v_end, p_start, p_end, allocator) catch |e| {
+        panic(@errorReturnTrace(), "Failed to map kernel directory: {}\n", e);
+    };
     const tty_addr = tty.getVideoBufferAddress();
     // If the previous mappping space didn't cover the tty buffer, do so now
     if (v_start > tty_addr or v_end <= tty_addr) {
         const tty_phys = virtToPhys(tty_addr);
         const tty_buff_size = 32 * 1024;
-        mapDir(kernel_directory, tty_phys, tty_phys + tty_buff_size, tty_addr, tty_addr + tty_buff_size, allocator) catch panic(@errorReturnTrace(), "Failed to map vga buffer in kernel directory");
+        mapDir(kernel_directory, tty_addr, tty_addr + tty_buff_size, tty_phys, tty_phys + tty_buff_size, allocator) catch |e| {
+            panic(@errorReturnTrace(), "Failed to map vga buffer in kernel directory: {}\n", e);
+        };
     }
 
     const dir_physaddr = @ptrToInt(virtToPhys(kernel_directory));
@@ -338,7 +348,12 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void 
         :
         : [addr] "{eax}" (dir_physaddr)
     );
-    isr.registerIsr(14, pageFault) catch panic(@errorReturnTrace(), "Failed to register page fault ISR");
+    isr.registerIsr(14, if (options.rt_test) rt_pageFault else pageFault) catch |e| {
+        panic(@errorReturnTrace(), "Failed to register page fault ISR: {}\n", e);
+    };
+    log.logInfo("Done\n");
+
+    if (options.rt_test) runtimeTests(v_end);
 }
 
 fn checkDirEntry(entry: DirectoryEntry, virt_start: usize, virt_end: usize, phys_start: usize, table: *Table) void {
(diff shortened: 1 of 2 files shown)


src/kernel/arch/x86/paging.zig
```diff
@@ -196,3 +204,28 @@ test "mapDir" {
...
+fn runtimeTests(v_end: u32) void {
+    // Accessing unmapped mem causes a page fault
+    var ptr = @intToPtr(*u8, v_end + PAGE_SIZE);
+    ptr.* = 1;
+    // This is the label that we return to after processing the page fault
+    asm volatile (
+        \\.global rt_fault_callback
+        \\rt_fault_callback:
+    );
+    std.debug.assert(faulted);
```
DrDeano: Should use testing.expect() as assert will be optimized out in release-fast

src/kernel/arch/x86/paging.zig
```diff
@@ -142,8 +144,14 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void
...
+        :
+        : [addr] "{eax}" (dir_physaddr)
+    );
+    isr.registerIsr(14, if (options.rt_test) testPageFault else pageFault) catch unreachable; //else pageFault) catch unreachable;
+    log.logInfo("Done\n");
+
+    if (options.rt_test) runtimeTests(v_end);
 }
 
 test "isAligned" {
```
DrDeano: Could the tests be arrange in the same order as the functions.
Also, could you move these tests to a unit test file following my unit testing guide if possible. If not possible, nvm :)

src/kernel/arch/x86/paging.zig
```diff
@@ -184,7 +192,7 @@ test "mapDir" {
     var direct = std.heap.DirectAllocator.init();
```
DrDeano: Could you add more unit tests for mapDir to test all the return errors/code paths. You can use expectError.

src/kernel/arch/x86/paging.zig
```diff
@@ -196,3 +204,28 @@ test "mapDir" {
...
+
+var faulted = false;
+
+fn testPageFault(ctx: *arch.InterruptContext) void {
+    faulted = true;
+    // Return to the fault callback
+    ctx.eip = @ptrToInt(&rt_fault_callback);
+}
+
+fn runtimeTests(v_end: u32) void {
```
DrDeano: You could add a test that when accessing memory within a valid page doesn't cause a page fault. So just copy the runtimeTests function and change the pointer address and change the assert to `expect(!faulted)`

src/kernel/arch/x86/paging.zig
```diff
@@ -142,8 +144,14 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void
 
     mapDir(kernel_directory, p_start, p_end, v_start, v_end, allocator) catch unreachable;
     const dir_physaddr = @ptrToInt(kernel_directory) - @ptrToInt(&KERNEL_ADDR_OFFSET);
-    asm volatile ("mov %[addr], %%cr3" :: [addr] "{eax}" (dir_physaddr));
-    isr.registerIsr(14, pageFault) catch unreachable;
+    asm volatile ("mov %[addr], %%cr3"
+        :
+        : [addr] "{eax}" (dir_physaddr)
+    );
+    isr.registerIsr(14, if (options.rt_test) testPageFault else pageFault) catch unreachable; //else pageFault) catch unreachable;
```
DrDeano: Remove comment

src/kernel/arch/x86/paging.zig
```diff
@@ -142,8 +144,14 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void
 
     mapDir(kernel_directory, p_start, p_end, v_start, v_end, allocator) catch unreachable;
     const dir_physaddr = @ptrToInt(kernel_directory) - @ptrToInt(&KERNEL_ADDR_OFFSET);
-    asm volatile ("mov %[addr], %%cr3" :: [addr] "{eax}" (dir_physaddr));
-    isr.registerIsr(14, pageFault) catch unreachable;
+    asm volatile ("mov %[addr], %%cr3"
```
DrDeano: Idk if I mentioned this before, but could this be moved to the arch.zig for x86. This will then allow you to unit test the init function by mocking out all the arch specific calls.

src/kernel/arch/x86/paging.zig
DrDeano: You wont be able to move the unit tests to a separate file bc the functions are private. Ignore that comment.

src/kernel/arch/x86/paging.zig
SamTebbs33: I didn't add a test for accessing mapped memory since if it didn't work then we wouldn't even get past the point of setting the page directory without faulting xD

src/kernel/arch/x86/paging.zig
SamTebbs33: The only things in the arch.zig file should be things that all architectures can do and paging may not be one of them. Why do I need to do mocking for the init function? Runtime tests catch all the testable functionality anyway.

src/kernel/arch/x86/paging.zig
DrDeano: Good point lol. I was thinking maybe it is best not to write to a random location as if successes, then could cause problems if the memory was important. Maybe just read.

SamTebbs33: Agreed :). On second thoughts, maybe I should add an explicit test for reading mapped memory, since if we change it so that a page fault isn't always fatal then this test may be reached even if a fault occurred.

src/kernel/arch/x86/paging.zig
DrDeano: I mean the assembly, not the whole init function lol. As I have the assembly of loading the GDT and IDT there. Then can write a unit test that will mock out the assembly and check the addresse being passed to the cr3 register.

SamTebbs33: Yeah I mean  the assembly too :p I'm not even sure if lgdt and lidt should be in there either, since they're only used by one architecture and one file each.

DrDeano: Hmmm, but I think that all the x86 files will be cluttered with assembly and having it all in one place makes the code base neater. Maybe could have a assembly file instead of putting it in the arch.zig

SamTebbs33: Not sure I agree, I think adding more files makes it less neat. Really, inline assembly is no different to normal code so I don't see why it all has to be moved to one place.

DrDeano: ok, you win for now :P. You may keep the assembly here

test/unittests/test_all.zig
```diff
@@ -1,4 +1,5 @@
 test "all" {
     _ = @import("kernel/test_vga.zig");
     _ = @import("kernel/test_tty.zig");
+    _ = @import("../../src/kernel/arch.zig");
```
DrDeano: This file should only contain `test_*.zig` files, so you will need to create a test_paging.zig file which will import the paging file. Then import the test_paging.zig here.

SamTebbs33: Ah yes, I mistakenly included this from my testing. Good spot

test/unittests/test_all.zig
```diff
@@ -1,5 +1,4 @@
 test "all" {
     _ = @import("kernel/test_vga.zig");
     _ = @import("kernel/test_tty.zig");
-    _ = @import("../../src/kernel/arch.zig");
```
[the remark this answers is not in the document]
SamTebbs33: I believe we have decided that tests will be in the same source file as the tested code, so I will fix this.

SamTebbs33:
We should wait until my 4kb paging PR is merged so I can add some new paging tests to this PR.

SamTebbs33:
Ready for review again.

src/kernel/arch/x86/paging.zig
```diff
@@ -438,3 +455,51 @@ test "mapDir" {
...
+extern var rt_fault_callback: *u32;
+extern var rt_fault_callback2: *u32;
+
+var faulted = false;
+var use_callback2 = false;
+
+fn rt_pageFault(ctx: *arch.InterruptContext) void {
+    faulted = true;
+    // Return to the fault callback
+    ctx.eip = @ptrToInt(&if (use_callback2) rt_fault_callback2 else rt_fault_callback);
```
DrDeano: Thinking about this, I don't get how this works. Following the interrupt chain, there needs to be a `iret` to exit the interrupt. So the code following this, (the rest of the code) is in interrupt mode. And the stack will be messed up. Correct me if I am wrong. Do you need to modify the instruction pointer as shouldn't the code return to where the page fault happened. I haven't written a proper page fault handler, but in real code, if a page fault happened, the OS will load the page and then return the previous running code. But idk if there needs to be a `iret`

test/kernel/arch/x86/rt-test.py
```diff
@@ -7,6 +7,9 @@ def get_test_cases(TestCase):
             TestCase("PIC init", [r"Init pic", r"Done"]),
             TestCase("PIC tests", [r"PIC: Tested masking"]),
             TestCase("PIT init", [r"Init pit", r".+", r"Done"]),
+            TestCase("Paging init", [r"Init paging", r"Done"]),
+            TestCase("Paging tests", [r"Paging: Tested accessing unmapped memory"]),
```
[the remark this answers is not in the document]
SamTebbs33: Yeah it's much better to have them in one as that was the original intention :+1:

test/unittests/test_all.zig
```diff
@@ -0,0 +1,4 @@
+test "all" {
```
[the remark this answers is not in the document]
SamTebbs33: oopsie

src/kernel/arch/x86/paging.zig
SamTebbs33: This fault handler is called by one if the `isrX` functions, so it will return to it, which eventually does an `iret`, jumping to the instruction at `ctx.eip`. We don't want to return to where the page fault happened as that would cause an infinite faulting loop, so we have to jump to the next instruction, hence the callbacks.

DrDeano: I see, yes, bc ur not accurately modifying the EIP, but the context which will be popped in the iret.

src/kernel/arch/x86/paging.zig
```diff
@@ -323,26 +323,26 @@ pub fn init(mem_profile: *const MemProfile, allocator: *std.mem.Allocator) void
...
     // Map in kernel
-    mapDir(kernel_directory, v_start, v_end, p_start, p_end, allocator) catch panic(@errorReturnTrace(), "Failed to map kernel directory");
+    mapDir(kernel_directory, v_start, v_end, p_start, p_end, allocator) catch |e| panic(@errorReturnTrace(), "Failed to map kernel directory: {}", @errorName(e));
     const tty_addr = tty.getVideoBufferAddress();
     // If the previous mappping space didn't cover the tty buffer, do so now
     if (v_start > tty_addr or v_end <= tty_addr) {
         const tty_phys = virtToPhys(tty_addr);
         const tty_buff_size = 32 * 1024;
-        mapDir(kernel_directory, tty_addr, tty_addr + tty_buff_size, tty_phys, tty_phys + tty_buff_size, allocator) catch panic(@errorReturnTrace(), "Failed to map vga buffer in kernel directory");
+        mapDir(kernel_directory, tty_addr, tty_addr + tty_buff_size, tty_phys, tty_phys + tty_buff_size, allocator) catch |e| panic(@errorReturnTrace(), "Failed to map vga buffer in kernel directory: {}", @errorName(e));
```
DrDeano: You don't need the `@errorName` as fmt handles it for you. Also (u may reject), so the line isn't long, have this like: 
```Zig
mapDir(kernel_directory, tty_addr, tty_addr + tty_buff_size, tty_phys, tty_phys + tty_buff_size, allocator) catch |e| {
    panic(@errorReturnTrace(), "Failed to map vga buffer in kernel directory: {}\n", e);
};
```
Also for the other error printing.
Also add new lines (\n) to the end of the logging.

src/kernel/arch/x86/paging.zig
SamTebbs33: Exactly

src/kernel/arch/x86/paging.zig
SamTebbs33: Oh cool

PR #76, opened 2019-09-10
Fix gdb symbol-file argument

### LINKED ISSUE #75: Fix symbol file given to gdb
The symbol file path given to the gdb command is incorrect. We need to pass the path to the elf file in the same string as the "symbol-file" argument.

SamTebbs33:
Fixes the argument to symbol-file, which is passed to gdb. Closes #75

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -99,8 +99,8 @@ pub fn build(b: *Builder) !void {
     const debug_cmd = b.addSystemCommand([_][]const u8{
         "gdb",
         "-ex",
+        symbol_file_arg,
     });
-    debug_cmd.addArg(symbol_file_arg);
     debug_cmd.addArgs([_][]const u8{
         "-ex",
         "target remote localhost:1234",
--- a/build.zig
+++ b/build.zig
@@ -48,8 +48,9 @@ pub fn build(b: *Builder) !void {
     grub_cmd.step.dependOn(&mkdir_cmd.step);
 
     const cp_elf_cmd = b.addSystemCommand([_][]const u8{"cp"});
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -94,12 +95,12 @@ pub fn build(b: *Builder) !void {
     }
 
     const debug_step = b.step("debug", "Debug with gdb");
+    const symbol_file_arg = try std.mem.join(b.allocator, " ", [_][]const u8{ "symbol-file", elf_path });
     const debug_cmd = b.addSystemCommand([_][]const u8{
         "gdb",
         "-ex",
-        "symbol-file",
     });
```
DrDeano: why not add the arguments here. So have:
const debug_cmd = b.addSystemCommand([_][]const u8{
    "gdb",
    "-ex",
    "symbol-file",
    elf_path ,
});

build.zig
```diff
@@ -94,12 +95,12 @@ pub fn build(b: *Builder) !void {
...
-    debug_cmd.addArtifactArg(exec);
+    debug_cmd.addArg(symbol_file_arg);
     debug_cmd.addArgs([_][]const u8{
```
DrDeano: Same here, just add the args to the end of the addSystemCommand

build.zig
SamTebbs33: "symbol-file" and the path have to be in the same string, so I could do
```zig
const debug_cmd = b.addSystemCommand([_][]const u8{
    "gdb",
    "-ex",
    symbol_file_arg,
});
```
instead

DrDeano: which ever look neater/clearer

PR #85, opened 2019-09-17
Only test a single build mode

SamTebbs33:
This patch makes `zig build` take a build mode and only test that build mode. All modes will be tested using parallel jobs in CI to make sure that the kernel works with them fine.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -21,6 +21,7 @@ pub fn build(b: *Builder) !void {
         else => unreachable,
     };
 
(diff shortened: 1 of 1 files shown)


SamTebbs33:
> lgtm, but have one question, is it possible to include a test all build options with one line? If this is not easy to add, then nvm. I can cope with running 4 commands ;)

Perhaps I could add a `-Dall_build_modes` option that will build and test all of them

PR #87, opened 2019-09-18
Migrate to github actions for CI

SamTebbs33:
Migrate to github actions for CI. It offers easy parallelisation and is more integrated with github. We should explore what other workflows would be useful (such as the labeler).


README.md
```diff
@@ -1,6 +1,6 @@
 # Pluto
 
-[![Build Status](https://dev.azure.com/samueltebbs/pluto/_apis/build/status/SamTebbs33.pluto?branchName=develop)](https://dev.azure.com/samueltebbs/samueltebbs/_build/latest?definitionId=1&branchName=develop)
+[![Build Status](https://github.com/SamTebbs33/pluto/workflows/CI/badge.svg)](https://github.com/SamTebbs33/pluto/actions)
```
DrDeano: This file doesn't exist, as it doesn't show up in the read me.

SamTebbs33: The badge only exists for the default branch, which is develop, so it won't show up until we get this into develop :)

PR #93, opened 2019-10-01
Added unit and runtime tests

DrDeano:
Also fixed some typos
Also changed the panic a bit
Removed changed to unnecessary files

### DIFF
--- a/src/kernel/arch/x86/isr.zig
+++ b/src/kernel/arch/x86/isr.zig
@@ -1,46 +1,31 @@
-// Zig version: 0.4.0
-
-const panic = @import("../../panic.zig").panic;
-const idt = @import("idt.zig");
-const arch = @import("arch.zig");
+const std = @import("std");
+const builtin = @import("builtin");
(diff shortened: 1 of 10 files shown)


src/kernel/arch/x86/isr.zig
```diff
@@ -212,4 +302,182 @@ pub fn init() void {
...
+    // Call function
+    try registerIsr(syscalls.INTERRUPT, testFunction3);
+    expectError(IsrError.IsrExists, registerIsr(syscalls.INTERRUPT, testFunction4));
+
+    // Post testing
+    if (syscall_handler) |h| {
+        // Should still have the original handler
+        expectEqual(testFunction3, h);
+    } else {
+        std.debug.panic("Pre-testing failed");
```
SamTebbs33: Why `std.debug.panic` and not `@panic'?

DrDeano: have removed the panics and use `expect`'s

PR #94, opened 2019-10-01
Added unit tests for PIC

DrDeano:
Removed I/O wait as it isn't needed and uses the syscall interrupt.


DrDeano:
Given that the unit test covers everything and the PIT works. Along with that there isn't a init function for the PIC, doesn't seem necessary to add run time tests. But if you would like one, I can add one if you would like.

SamTebbs33:
> Given that the unit test covers everything and the PIT works. Along with that there isn't a init function for the PIC, doesn't seem necessary to add run time tests. But if you would like one, I can add one if you would like.

Agreed. Unit tests cover everything that we ourselves can implement and the rest is up to the hardware/emulator to get right.

SamTebbs33:
Part of #60

PR #95, opened 2019-10-01
Remove debug option

### LINKED ISSUE #70: Fix debug option
The help text for the `zig build` debug option is confusing, since it says that it causes the kernel to be built with debug info but this happens irrespective of the option's value. As it stands, our options are

* The option should be removed (meaning `zig build run` will have to be branched out to `zig build run-debug`).
* It should be made true by default.
* It should be made false by default.

For option 2 and 3, a value of  false will mean we have to strip the debug info from the build with `--strip` or the equivalent `build.zig` function.

SamTebbs33:
This patch removes the debug option in favour of a new `debug-run` step which replaces `run -Ddebug=true`. Closes #70

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -22,7 +22,6 @@ pub fn build(b: *Builder) !void {
     };
 
     const build_mode = b.standardReleaseOptions();
-    const debug = b.option(bool, "debug", "build with debug symbols / make qemu wait for a debug connection") orelse false;
     const rt_test = b.option(bool, "rt-test", "enable/disable runtime testing") orelse false;
 
     const main_src = "src/kernel/kmain.zig";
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -63,30 +62,36 @@ pub fn build(b: *Builder) !void {
...
-
-    if (debug) {
-        qemu_cmd.addArgs([_][]const u8{ "-s", "-S" });
-    }
+    };
+    const qemu_cmd = b.addSystemCommand(qemu_args);
+    const qemu_debug_cmd = b.addSystemCommand(qemu_args);
+    qemu_debug_cmd.addArgs([_][]const u8{ "-s", "-S" });
 
+    const qemu_rt_test_args = [_][]const u8{ "-display", "none" };
```
DrDeano: I guess this arg can be put in the if block as they aren't used else where.

PR #101, opened 2019-10-08
Add log tests

### LINKED ISSUE #99: Logging runtime tests and/or unit tests
Runtime tests and/or unit tests need to be added for logging.

To test:
* Formatting works (no need to test every possible data type as that is down to the std lib)
* All log levels work

SamTebbs33:
This patch adds log runtime tests. They will fail until #100 is merged as it needs the `prefix` argument added to `TestCase`. Closes #99

### DIFF
--- a/src/kernel/kmain.zig
+++ b/src/kernel/kmain.zig
@@ -36,7 +36,11 @@ export fn kmain(mb_info: *multiboot.multiboot_info_t, mb_magic: u32) void {
         var buffer = mem_profile.vaddr_end[0..mem_profile.fixed_alloc_size];
         var fixed_allocator = std.heap.FixedBufferAllocator.init(buffer);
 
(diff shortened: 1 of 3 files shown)


SamTebbs33:
Finally got the branch rebased and passing

src/kernel/kmain.zig
```diff
@@ -37,6 +37,8 @@ export fn kmain(mb_info: *multiboot.multiboot_info_t, mb_magic: u32) void {
         var fixed_allocator = std.heap.FixedBufferAllocator.init(buffer);
 
         serial.init(serial.DEFAULT_BAUDRATE, serial.Port.COM1) catch unreachable;
+        if (build_options.rt_test)
```
DrDeano: `catch unreachable`, Should handle error.

PR #104, opened 2019-10-10
Add physical memory manager

### LINKED ISSUE #67: Physical memory manager
Manage the physical memory.

- Some areas of physical memory are unusable and will need to be managed so not to use this memory.
- Manage the memory mapped I/O. May this should be a separate manager, but will have ties with the physical memory manager as shouldn't allocate memory here.
- Biggest of all, manage where everything is, as data and code have to be somewhere in memory.

Also, we can decide with the 32 bit OS, handle 4KB chunks where for 64 bit OS handle 4MB chunks, as who has less than 4GB of RAM these days, am I right ;).

Here is an example written in C:  
https://github.com/DrDeano/DeanOS/blob/master/src/kernel/kernel/pmm.c

A good tutorial:   
http://www.brokenthorn.com/Resources/OSDev17.html

SamTebbs33:
This patch adds a physical memory manager that keeps track of which physical blocks are used. It parses the multiboot memory map for reserved memory and sets the memory used by kernel code as occupied. Closes #67

### DIFF
--- a/src/kernel/arch/x86/arch.zig
+++ b/src/kernel/arch/x86/arch.zig
@@ -11,6 +11,7 @@ const paging = @import("paging.zig");
 const syscalls = @import("syscalls.zig");
 const mem = @import("../../mem.zig");
 const multiboot = @import("../../multiboot.zig");
+const pmm = @import("pmm.zig");
 const MemProfile = mem.MemProfile;
 
 /// The interrupt context that is given to a interrupt handler. It contains most of the registers
@@ -45,6 +46,9 @@ pub const InterruptContext = struct {
     user_esp: u32,
     ss: u32,
 };
+
+/// The size of each allocatable block of memory, normally set to the page size.
+pub const MEMORY_BLOCK_SIZE = paging.PAGE_SIZE_4KB;
 
 ///
 /// Assembly to write to a given port with a byte of data.
--- a/src/kernel/bitmap.zig
+++ b/src/kernel/bitmap.zig
@@ -1,0 +1,306 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const testing = std.testing;
+
+///
+/// A bitmap that uses a specific type to store the entries.
+///
+/// Arguments:
+///     IN BitmapType: type - The integer type to use to store entries.
+///
+/// Return: type.
+///     The bitmap type created.
+///
+pub fn Bitmap(comptime BitmapType: type) type {
+    return struct {
+        /// The possible errors thrown by bitmap functions
+        pub const BitmapError = error{
+            /// The address given was outside the region covered by a bitmap
+            OutOfBounds,
+        };
+
+        const Self = @This();
+
+        /// The number of entries that one bitmap type can hold. Evaluates to the number of bits the type has
+        pub const ENTRIES_PER_BITMAP: u32 = std.meta.bitCount(BitmapType);
+
+        /// The value that a full bitmap will have
+        pub const BITMAP_FULL = std.math.maxInt(BitmapType);
+
+        /// The type of an index into a bitmap entry. The smallest integer needed to represent all bit positions in the bitmap entry type
+        pub const IndexType = @Type(builtin.TypeInfo{ .Int = builtin.TypeInfo.Int{ .is_signed = false, .bits = std.math.log2(std.meta.bitCount(BitmapType)) } });
+
+        num_bitmaps: u32,
+        num_entries: u32,
+        bitmaps: []BitmapType,
+        num_free_entries: u32,
+
+        ///
+        /// Create an instance of this bitmap type.
+        ///
+        /// Arguments:
+        ///     IN num_entries: u32 - The number of entries that the bitmap created will have.
+        ///         The number of BitmapType required to store this many entries will be allocated and each will be zeroed.
+        ///     IN allocator: *std.mem.Allocator - The allocator to use when allocating the BitmapTypes required.
+        ///
+        /// Return: Self.
+        ///     The bitmap instance.
+        ///
+        /// Error: std.mem.Allocator.Error
+        ///     OutOfMemory: There isn't enough memory available to allocate the required number of BitmapType.
+        ///
+        pub fn init(num_entries: u32, allocator: *std.mem.Allocator) !Self {
+            const num = @floatToInt(u32, @ceil(@intToFloat(f32, num_entries) / @intToFloat(f32, ENTRIES_PER_BITMAP)));
+            const self = Self{
+                .num_bitmaps = num,
+                .num_entries = num_entries,
+                .bitmaps = try allocator.alloc(BitmapType, num),
+                .num_free_entries = num_entries,
+            };
+            for (self.bitmaps) |*bmp| {
+                bmp.* = 0;
+            }
+            return self;
+        }
+
+        ///
+        /// Set an entry within a bitmap as occupied.
+        ///
+        /// Arguments:
+        ///     INOUT self: *Self - The bitmap to modify.
+        ///     IN idx: BitmapIndex - The index within the bitmap to set.
+        ///
+        /// Error: BitmapError.
+        ///     OutOfBounds: The index given is out of bounds.
+        ///
+        pub fn setEntry(self: *Self, idx: u32) BitmapError!void {
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
+            if (!try self.isSet(idx)) {
+                const bit = self.indexToBit(idx);
+                self.bitmaps[idx / ENTRIES_PER_BITMAP] |= bit;
+                self.num_free_entries -= 1;
+            }
+        }
+
+        ///
+        /// Set an entry within a bitmap as unoccupied.
+        ///
+        /// Arguments:
+        ///     INOUT self: *Self - The bitmap to modify.
+        ///     IN idx: BitmapIndex - The index within the bitmap to clear.
+        ///
+        /// Error: BitmapError.
+        ///     OutOfBounds: The index given is out of bounds.
+        ///
+        pub fn clearEntry(self: *Self, idx: u32) BitmapError!void {
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
+            if (try self.isSet(idx)) {
+                const bit = self.indexToBit(idx);
+                self.bitmaps[idx / ENTRIES_PER_BITMAP] &= ~bit;
+                self.num_free_entries += 1;
+            }
+        }
+
+        ///
+        /// Convert a global bitmap index into the bit corresponding to an entry within a single BitmapType.
+        ///
+        /// Arguments:
+        ///     IN self: *Self - The bitmap to use.
+        ///     IN idx: u32 - The index into all of the bitmap's entries.
+        ///
+        /// Return: BitmapType.
+        ///     The bit corresponding to that index but within a single BitmapType.
+        ///
+        fn indexToBit(self: *Self, idx: u32) BitmapType {
+            return @as(BitmapType, 1) << @intCast(IndexType, idx % ENTRIES_PER_BITMAP);
+        }
+
+        ///
+        /// Set the first free entry within the bitmaps as occupied.
+        ///
+        /// Return: ?u32.
+        ///     The index within all bitmaps that was set or null if there wasn't one free.
+        ///     0 .. ENTRIES_PER_BITMAP - 1 if in the first bitmap, ENTRIES_PER_BITMAP .. ENTRIES_PER_BITMAP * 2 - 1 if in the second etc.
+        ///
+        pub fn setFirstFree(self: *Self) ?u32 {
+            if (self.num_free_entries == 0) return null;
+            for (self.bitmaps) |*bmp, i| {
+                if (bmp.* == BITMAP_FULL)
+                    continue;
+                const bit = @truncate(IndexType, @ctz(BitmapType, ~bmp.*));
+                const idx = bit + @intCast(u32, i) * ENTRIES_PER_BITMAP;
+                // Failing here means that the index is outside of the bitmap, so there are no free entries
+                self.setEntry(idx) catch return null;
+                return idx;
+            }
+            return null;
+        }
+
+        ///
+        /// Check if an entry is set.
+        ///
+        /// Arguments:
+        ///     IN self: *Bitmap - The bitmap to check.
+        ///     IN idx: u32 - The entry to check.
+        ///
+        /// Return: bool.
+        ///     True if the entry is set, else false.
+        ///
+        /// Error: BitmapError.
+        ///     OutOfBounds: The index given is out of bounds.
+        ///
+        pub fn isSet(self: *Self, idx: u32) BitmapError!bool {
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
+            return (self.bitmaps[idx / ENTRIES_PER_BITMAP] & self.indexToBit(idx)) != 0;
+        }
+    };
+}
+
+test "setEntry" {
+    var bmp = try Bitmap(u32).init(31, std.heap.direct_allocator);
+    testing.expectEqual(@as(u32, 31), bmp.num_free_entries);
+
+    try bmp.setEntry(0);
+    testing.expectEqual(@as(u32, 1), bmp.bitmaps[0]);
+    testing.expectEqual(@as(u32, 30), bmp.num_free_entries);
+
+    try bmp.setEntry(1);
+    testing.expectEqual(@as(u32, 3), bmp.bitmaps[0]);
+    testing.expectEqual(@as(u32, 29), bmp.num_free_entries);
+
+    // Repeat setting entry 1 to make sure state doesn't change
+    try bmp.setEntry(1);
+    testing.expectEqual(@as(u32, 3), bmp.bitmaps[0]);
+    testing.expectEqual(@as(u32, 29), bmp.num_free_entries);
+
+    testing.expectError(Bitmap(u32).BitmapError.OutOfBounds, bmp.setEntry(31));
+    testing.expectEqual(@as(u32, 29), bmp.num_free_entries);
+}
+
+test "clearEntry" {
+    var bmp = try Bitmap(u32).init(32, std.heap.direct_allocator);
+    testing.expectEqual(@as(u32, 32), bmp.num_free_entries);
+
+    try bmp.setEntry(0);
+    testing.expectEqual(@as(u32, 31), bmp.num_free_entries);
+    try bmp.setEntry(1);
+    testing.expectEqual(@as(u32, 30), bmp.num_free_entries);
+    testing.expectEqual(@as(u32, 3), bmp.bitmaps[0]);
+    try bmp.clearEntry(0);
+    testing.expectEqual(@as(u32, 31), bmp.num_free_entries);
+    testing.expectEqual(@as(u32, 2), bmp.bitmaps[0]);
+
+    // Repeat to make sure state doesn't change
+    try bmp.clearEntry(0);
+    testing.expectEqual(@as(u32, 31), bmp.num_free_entries);
+    testing.expectEqual(@as(u32, 2), bmp.bitmaps[0]);
+
+    // Try clearing an unset entry to make sure state doesn't change
+    try bmp.clearEntry(2);
+    testing.expectEqual(@as(u32, 31), bmp.num_free_entries);
+    testing.expectEqual(@as(u32, 2), bmp.bitmaps[0]);
+
+    testing.expectError(Bitmap(u32).BitmapError.OutOfBounds, bmp.clearEntry(32));
+}
+
+test "setFirstFree multiple bitmaps" {
+    var bmp = try Bitmap(u8).init(9, std.heap.direct_allocator);
+
+    // Allocate the first entry
+    testing.expectEqual(bmp.setFirstFree() orelse unreachable, 0);
+    testing.expectEqual(bmp.bitmaps[0], 1);
+
+    // Allocate the second entry
+    testing.expectEqual(bmp.setFirstFree() orelse unreachable, 1);
+    testing.expectEqual(bmp.bitmaps[0], 3);
+
+    // Allocate the entirety of the first bitmap
+    var entry: u32 = 2;
+    var expected: u8 = 7;
+    while (entry < Bitmap(u8).ENTRIES_PER_BITMAP) {
+        testing.expectEqual(bmp.setFirstFree() orelse unreachable, entry);
+        testing.expectEqual(bmp.bitmaps[0], expected);
+        if (entry + 1 < Bitmap(u8).ENTRIES_PER_BITMAP) {
+            entry += 1;
+            expected = expected * 2 + 1;
+        } else {
+            break;
+        }
+    }
+
+    // Try allocating an entry in the next bitmap
+    testing.expectEqual(bmp.setFirstFree() orelse unreachable, Bitmap(u8).ENTRIES_PER_BITMAP);
+    testing.expectEqual(bmp.bitmaps[0], Bitmap(u8).BITMAP_FULL);
+    testing.expectEqual(bmp.bitmaps[1], 1);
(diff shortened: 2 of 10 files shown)


SamTebbs33:
I reckon this is finally ready for review! @DrDeano

src/kernel/pmm.zig
```diff
@@ -0,0 +1,175 @@
...
+///
+/// Set the bitmap entry for an address as occupied
+///
+/// Arguments:
+///     IN addr: usize - The address.
+///
+/// Error: PmmBitmap.Error.
+///     *: See PmmBitmap.setEntry.
+///
+fn setAddr(addr: usize) PmmBitmap.Error!void {
```
DrDeano: Unit test

src/kernel/pmm.zig
```diff
@@ -0,0 +1,175 @@
...
+///
+/// Arguments:
+///     IN addr: usize - The address to check.
+///
+/// Return: True if occupied, else false.
+///
+/// Error: PmmBitmap.Error.
+///     *: See PmmBitmap.setEntry.
+///
+fn isSet(addr: usize) PmmBitmap.Error!bool {
```
DrDeano: Unit test

src/kernel/pmm.zig
```diff
@@ -0,0 +1,175 @@
...
+                    else => panic(@errorReturnTrace(), "Failed setting address 0x{x} from memory map as occupied: {}", addr, e),
+                };
+            }
+        }
+    }
+    // Occupy kernel memory
+    var addr = std.mem.alignBackward(@ptrToInt(mem.physaddr_start), BLOCK_SIZE);
+    while (addr < @ptrToInt(mem.physaddr_end)) : (addr += BLOCK_SIZE) {
+        setAddr(addr) catch |e| switch (e) {
+            PmmBitmap.Error.OutOfBounds => break,
```
DrDeano: Surely it will be  problem if we can't set the kernel address space, so should panic?

src/kernel/pmm.zig
```diff
@@ -0,0 +1,175 @@
...
+        runtimeTests(mem);
+    }
+}
+
+///
+/// Find the next free memory block, set it as occupied and return it. The region allocated will be of size BLOCK_SIZE.
+///
+/// Return: The address that was allocated.
+///
+pub fn alloc() ?usize {
```
DrDeano: File ordering, should be above the init function

src/kernel/pmm.zig
```diff
@@ -0,0 +1,175 @@
...
+pub fn free(addr: usize) (PmmBitmap.Error || Error)!void {
+    const idx = @intCast(u32, addr / BLOCK_SIZE);
+    if (try bitmap.isSet(idx)) {
+        try bitmap.clearEntry(idx);
+    } else {
+        return Error.NotAllocated;
+    }
+}
+
+pub fn runtimeTests(mem: *const MemProfile) void {
```
DrDeano: Not pub

src/kernel/pmm.zig
```diff
@@ -0,0 +1,175 @@
...
+    // Make sure that occupied memory can't be allocated
+    var prev_alloc: usize = std.math.maxInt(usize);
+    while (alloc()) |alloced| {
+        if (prev_alloc == alloced) {
+            panic(null, "PMM allocated the same address twice: 0x{x}", alloced);
+        }
+        prev_alloc = alloced;
+        for (mem.mem_map) |entry| {
+            if (entry.@"type" != MEMORY_AVAILABLE) {
```
DrDeano: If possible, add a case to ensure you can't allocate memory within the kernel space

DrDeano:
As a side, could the pmm implement the Allocater interface

src/kernel/arch/x86/arch.zig
```diff
@@ -46,6 +47,8 @@ pub const InterruptContext = struct {
     ss: u32,
 };
 
+pub const MEMORY_BLOCK_SIZE = paging.PAGE_SIZE_4KB;
```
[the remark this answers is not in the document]
SamTebbs33: No it's my fault. I keep forgetting :p

src/kernel/mem.zig
```diff
@@ -47,8 +48,11 @@ var ADDR_OFFSET: usize = undefined;
 ///
 pub fn init(mb_info: *multiboot.multiboot_info_t) MemProfile {
     log.logInfo("Init mem\n");
+    defer log.logInfo("Done\n");
```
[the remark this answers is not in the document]
SamTebbs33: A separate issue sounds good :)

src/kernel/pmm.zig
SamTebbs33: Done. I merged it with the alloc and free tests as testing isSet requires them to be called anyway.

src/kernel/pmm.zig
SamTebbs33: Yeah we should :+1:

src/kernel/pmm.zig
SamTebbs33: I'm assuming you're talking about the kernel code space, which should be allocated anyway, so is the check necessary?

SamTebbs33:
> As a side, could the pmm implement the Allocater interface

What is the advantage of it being an Allocator?

SamTebbs33:
> > As a side, could the pmm implement the Allocater interface
> 
> What is the advantage of it being an Allocator?

Discussed offline, will implement :+1

test/mock/kernel/arch_mock.zig
```diff
@@ -35,6 +35,8 @@ pub const InterruptContext = struct {
     ss: u32,
 };
 
+pub const MEMORY_BLOCK_SIZE = 4096;
```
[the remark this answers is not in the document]
SamTebbs33: Do you mean changing it to be `MEMORY_BLOCK_SIZE = paging_mock.PAGE_SIZE_4KB` and just creating a paging_mock file with that definition?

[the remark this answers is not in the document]
DrDeano: Yep

src/kernel/pmm.zig
DrDeano: The check is always necessary, it's a unit test, everything must be tested ;)

SamTebbs33: Indeed, from the snippet I thought that this was in the `init` function :) I'll add a check.

SamTebbs33: Isn't the kernel code space covered by the check underneath though? https://github.com/SamTebbs33/pluto/blob/b74addacf932b52199aae519dba99e5d0e600a29/src/kernel/pmm.zig#L143

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,132 @@
...
+        const Self = @This();
+        pub const ENTRIES_PER_BITMAP: u32 = std.meta.bitCount(BitmapType);
+        pub const BITMAP_FULL = std.math.maxInt(BitmapType);
+
+        num_bitmaps: u32,
+        num_entries: u32,
+        bitmaps: []BitmapType,
+
+        pub fn init(num_entries: u32, allocator: *std.mem.Allocator) !Self {
+            const num = num_entries / ENTRIES_PER_BITMAP + @boolToInt(num_entries % ENTRIES_PER_BITMAP != 0);
```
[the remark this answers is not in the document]
SamTebbs33: Perhaps it could be replaced with `@ceil(num_entries / @as(f32, ENTRIES_PER_BITMAP))`?

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,132 @@
...
+}
+
+test "isSet" {
+    var bmp = try Bitmap(u32).init(32, std.heap.direct_allocator);
+    bmp.bitmaps[0] = 1;
+    testing.expectEqual(true, try bmp.isSet(0));
+    bmp.bitmaps[0] = 3;
+    testing.expectEqual(true, try bmp.isSet(1));
+    bmp.bitmaps[0] = 11;
+    testing.expect(try bmp.isSet(3));
```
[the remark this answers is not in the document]
SamTebbs33: Good idea

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const testing = std.testing;
+
+pub fn Bitmap(comptime BitmapType: type) type {
```
DrDeano: Doc comment

src/kernel/bitmap.zig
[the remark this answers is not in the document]
DrDeano: I think `@round` is better. So can have `@round(num_entries / @as(f32, ENTRIES_PER_BITMAP))`

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+                bmp.* = 0;
+            }
+            return self;
+        }
+
+        ///
+        /// Set an entry within a bitmap as occupied.
+        ///
+        /// Arguments:
+        ///     OUT bmp: *Bitmap - The bitmap to modify.
```
DrDeano: INOUT self: *Self

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+        pub fn setEntry(self: *Self, idx: u32) BitmapError!void {
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
+            if (!try self.isSet(idx)) {
+                const bit = @as(BitmapType, 1) << @intCast(IndexType, idx % ENTRIES_PER_BITMAP);
+                self.bitmaps[idx / ENTRIES_PER_BITMAP] |= bit;
+                self.num_free_entries -= 1;
+            }
+        }
+
+        pub fn clearEntry(self: *Self, idx: u32) BitmapError!void {
```
DrDeano: Doc comment

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+        /// Set an entry within a bitmap as occupied.
+        ///
+        /// Arguments:
+        ///     OUT bmp: *Bitmap - The bitmap to modify.
+        ///     IN idx: BitmapIndex - The index within the bitmap to set.
+        ///
+        pub fn setEntry(self: *Self, idx: u32) BitmapError!void {
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
+            if (!try self.isSet(idx)) {
+                const bit = @as(BitmapType, 1) << @intCast(IndexType, idx % ENTRIES_PER_BITMAP);
```
DrDeano: This could be made into a function as it is the same as the `clearEntry`. Plus this is different to the `clearEntry`

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+                const bit = @as(u32, 1) << @intCast(u5, idx % ENTRIES_PER_BITMAP);
+                self.bitmaps[idx / ENTRIES_PER_BITMAP] &= ~bit;
+                self.num_free_entries += 1;
+            }
+        }
+
+        ///
+        /// Set the first free entry within the bitmaps as occupied.
+        ///
+        /// Return: The index within all bitmaps that was set.
```
DrDeano: `Return: ?u32`

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+        ///     0 .. ENTRIES_PER_BITMAP - 1 if in the first bitmap, ENTRIES_PER_BITMAP .. ENTRIES_PER_BITMAP * 2 - 1 if in the second etc.
+        ///
+        pub fn setFirstFree(self: *Self) ?u32 {
+            if (self.num_free_entries == 0) return null;
+            for (self.bitmaps) |*bmp, i| {
```
DrDeano: `bmp` doesn't need to be a reference because you don't updates its value. Unless I'm wrong

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+                const bit = @truncate(IndexType, @ctz(BitmapType, ~bmp.*));
+                const idx = bit + @intCast(u32, i) * ENTRIES_PER_BITMAP;
+                // Failing here means that the index is outside of the bitmap, so there are no free entries
+                self.setEntry(idx) catch return null;
+                return idx;
+            }
+            return null;
+        }
+
+        pub fn isSet(self: *Self, idx: u32) BitmapError!bool {
```
DrDeano: Doc comment

src/kernel/bitmap.zig
```diff
@@ -0,0 +1,228 @@
...
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
+            return (self.bitmaps[idx / ENTRIES_PER_BITMAP] & @as(u32, 1) << @intCast(u5, idx % ENTRIES_PER_BITMAP)) != 0;
```
DrDeano: `@as(BitmapType, 1) << @intCast(IndexType, idx % ENTRIES_PER_BITMAP)`. So maybe this could be a function as it is used in 3 places

src/kernel/mem.zig
```diff
@@ -4,13 +4,30 @@ const expectEqual = std.testing.expectEqual;
 const log = @import("log.zig");
 
 pub const MemProfile = struct {
```
DrDeano: Doc comment the struct

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+/// The possible errors thrown by bitmap functions
+const Error = error{
+    /// The address given hasn't been allocated
+    NotAllocated,
+};
+
+/// The size of memory associated with each bitmap entry
+const BLOCK_SIZE = arch.MEMORY_BLOCK_SIZE;
+
+var bitmap: PmmBitmap = undefined;
```
DrDeano: Doc comment

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+    try bitmap.setEntry(@intCast(u32, addr / BLOCK_SIZE));
+}
+
+///
+/// Check if an address is set as occupied.
+///
+/// Arguments:
+///     IN addr: usize - The address to check.
+///
+/// Return: True if occupied, else false.
```
DrDeano: `Return: bool`

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+///     *: See PmmBitmap.setEntry. Could occur if the address is out of bounds.
+///
+fn isSet(addr: usize) PmmBitmap.BitmapError!bool {
+    return bitmap.isSet(@intCast(u32, addr / BLOCK_SIZE));
+}
+
+///
+/// Find the next free memory block, set it as occupied and return it. The region allocated will be of size BLOCK_SIZE.
+///
+/// Return: The address that was allocated.
```
DrDeano: `Return: ?usize`

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+const testing = std.testing;
+const panic = @import("panic.zig").panic;
+const log = @import("log.zig");
+const MEMORY_AVAILABLE = @import("multiboot.zig").MULTIBOOT_MEMORY_AVAILABLE;
+const Bitmap = @import("bitmap.zig").Bitmap;
+
+const PmmBitmap = Bitmap(u32);
+
+/// The possible errors thrown by bitmap functions
+const Error = error{
```
DrDeano: Ambiguous. Maybe PmmError

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+    const idx = @intCast(u32, addr / BLOCK_SIZE);
+    if (try bitmap.isSet(idx)) {
+        try bitmap.clearEntry(idx);
+    } else {
+        return Error.NotAllocated;
+    }
+}
+
+///
+/// Intiialise the physical memory manager and set all unavailable regions as occupied (those from the memory map and those from the linker symbols).
```
DrDeano: Too many ii :P

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+///
+pub fn init(mem: *const MemProfile, allocator: *std.mem.Allocator) void {
+    log.logInfo("Init pmm\n", .{});
+    bitmap = PmmBitmap.init(mem.mem_kb * 1024 / BLOCK_SIZE, allocator) catch @panic("Bitmap allocation failed");
+
+    // Occupy the regions of memory that the memory map describes as reserved
+    for (mem.mem_map) |entry| {
+        if (entry.@"type" != MEMORY_AVAILABLE) {
+            var addr = std.mem.alignBackward(@intCast(usize, entry.addr), BLOCK_SIZE);
+            var end = @intCast(usize, entry.addr + (entry.len - 1));
```
DrDeano: Unnecessary bracets

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+            // If the end address can be aligned without overflowing then align it
+            if (end <= std.math.maxInt(usize) - BLOCK_SIZE)
```
DrDeano: Idk whether we should have `{}` after the if statements. Or at least have a space before the while loop so is clearer

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+                    else => panic(@errorReturnTrace(), "Failed setting address 0x{x} from memory map as occupied: {}", .{ addr, e }),
+                };
+            }
+        }
+    }
+    // Occupy kernel memory
+    var addr = std.mem.alignBackward(@ptrToInt(mem.physaddr_start), BLOCK_SIZE);
+    while (addr < @ptrToInt(mem.physaddr_end)) : (addr += BLOCK_SIZE) {
+        setAddr(addr) catch |e| switch (e) {
+            PmmBitmap.BitmapError.OutOfBounds => panic(@errorReturnTrace(), "Failed setting kernel code address 0x{x} as occupied. The amount of system memory seems to be too low for the kernel image: {}", .{ addr, e }),
```
DrDeano: `error.OutOfBounds`. If OutOfBounds is unique I think

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+            else => panic(@errorReturnTrace(), "Failed setting kernel code address 0x{x} as occupied: {}", .{ addr, e }),
+        };
+    }
+    log.logInfo("Done\n", .{});
+    if (build_options.rt_test) {
+        runtimeTests(mem);
+    }
+}
+
+fn runtimeTests(mem: *const MemProfile) void {
```
DrDeano: Doc comment

src/kernel/pmm.zig
DrDeano: It is

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+        }
+        if (alloced >= std.mem.alignBackward(@ptrToInt(mem.physaddr_start), BLOCK_SIZE) and alloced < std.mem.alignForward(@ptrToInt(mem.physaddr_end), BLOCK_SIZE)) {
+            panic(null, "PMM allocated an address that should be reserved by kernel code: 0x{x}", .{alloced});
+        }
+    }
+    log.logInfo("PMM: Tested allocation\n", .{});
+}
+
+test "alloc" {
+    bitmap = try Bitmap(u32).init(32, std.heap.direct_allocator);
```
DrDeano: Idk whether bitmap needs to be cleared after the test as it is a global. This could be tested at the beginning of each test to ensure the bitmap is clear. As subsequentially tests could fail or produce undefined results.

src/kernel/pmm.zig
```diff
@@ -0,0 +1,214 @@
...
+    bitmap = try Bitmap(u32).init(num_entries, std.heap.direct_allocator);
+    var addr: u32 = 0;
+    var i: u32 = 0;
+    while (i < num_entries) : ({
+        i += 1;
+        addr += BLOCK_SIZE;
+    }) {
+        // Ensure all previous blocks are still set
+        var h: u32 = 0;
+        var addr3: u32 = 0;
```
DrDeano: This naming is a bit confusing as u have 3 here and then 2 further down.

src/kernel/pmm.zig
SamTebbs33: It shouldn't need to be since each test creates its own bitmap so there won't be any re-usage.

DrDeano: But `bitmap` is global and if you forget to reinitialise it could be a problem. If you re-assign it for each test then it should be fine, but just to bare it in mind as I had this problem when testing the tty.

SamTebbs33: Indeed. If we look at the existing tests before creating a new one we should spot that the bitmap needs to be recreated.

src/kernel/bitmap.zig
```diff
@@ -36,39 +65,65 @@ pub fn Bitmap(comptime BitmapType: type) type {
         /// Set an entry within a bitmap as occupied.
         ///
         /// Arguments:
-        ///     OUT bmp: *Bitmap - The bitmap to modify.
+        ///     INOUT bmp: *Bitmap - The bitmap to modify.
```
DrDeano: The naming is different, should be self: *Self

src/kernel/bitmap.zig
```diff
@@ -36,39 +65,65 @@ pub fn Bitmap(comptime BitmapType: type) type {
...
             }
         }
 
-        pub fn clearEntry(self: *Self, idx: u32) Error!void {
-            if (idx >= self.num_entries) return Error.OutOfBounds;
+        ///
+        /// Set an entry within a bitmap as unoccupied.
+        ///
+        /// Arguments:
+        ///     INOUT bmp: *Bitmap - The bitmap to modify.
```
DrDeano: Same here self: *Self

src/kernel/bitmap.zig
```diff
@@ -77,32 +132,44 @@ pub fn Bitmap(comptime BitmapType: type) type {
...
-        pub fn isSet(self: *Self, idx: u32) Error!bool {
-            if (idx >= self.num_entries) return Error.OutOfBounds;
+        ///
+        /// Check if an entry is set.
+        ///
+        /// Arguments:
+        ///     IN self: *Bitmap - The bitmap to check.
+        ///     IN idx: u32 - The entry to check.
+        ///
+        /// Return: True if the entry is set, else false.
```
DrDeano: Return: bool

src/kernel/bitmap.zig
```diff
@@ -77,32 +132,44 @@ pub fn Bitmap(comptime BitmapType: type) type {
...
+        ///
+        /// Error: BitmapError.
+        ///     OutOfBounds: The index given is out of bounds.
+        ///
+        pub fn isSet(self: *Self, idx: u32) BitmapError!bool {
+            if (idx >= self.num_entries) return BitmapError.OutOfBounds;
             return (self.bitmaps[idx / ENTRIES_PER_BITMAP] & @as(u32, 1) << @intCast(u5, idx % ENTRIES_PER_BITMAP)) != 0;
```
DrDeano: Can use your new `indexToBit` function here

src/kernel/pmm.zig
```diff
@@ -69,11 +69,11 @@ pub fn alloc() ?usize {
...
 ///     IN addr: usize - The previously allocated address to free. Will be aligned down to the nearest multiple of BLOCK_SIZE.
 ///
-/// Error: Error || PmmBitmap.Error.
+/// Error: Error || PmmBitmap.BitmapError.
 ///     Error.NotAllocated: The address wasn't allocated.
-///     PmmBitmap.Error.OutOfBounds: The address given was out of bounds.
+///     PmmBitmap.BitmapError.OutOfBounds: The address given was out of bounds.
 ///
-pub fn free(addr: usize) (PmmBitmap.Error || Error)!void {
+pub fn free(addr: usize) (PmmBitmap.BitmapError || Error)!void {
```
DrDeano: Error is ambiguous, which error set is this coming from?

SamTebbs33: There is an error set in this file called Error, but I can rename it.

PR #105, opened 2019-11-02
Move virtToPhys to mem.zig and add physToVirt

SamTebbs33:
This patch moves `virtToPhys` from paging.zig to mem.zig as it will be used by the stacktrace code and isn't particular to any specific architecture, so should be in arch-independent code.

### DIFF
--- a/src/kernel/mem.zig
+++ b/src/kernel/mem.zig
@@ -1,4 +1,6 @@
 const multiboot = @import("multiboot.zig");
+const std = @import("std");
+const expectEqual = std.testing.expectEqual;
 
 pub const MemProfile = struct {
     vaddr_end: [*]u8,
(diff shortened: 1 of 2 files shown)


src/kernel/mem.zig
```diff
@@ -29,3 +39,55 @@ pub fn init(mb_info: *multiboot.multiboot_info_t) MemProfile {
...
+///
+/// Convert a virtual address to its physical counterpart by subtracting the kernel virtual offset from the virtual address.
+///
+/// Arguments:
+///     IN virt: var - The virtual address to covert. Either an integer or pointer.
+///
+/// Return: @typeOf(virt)
+///     The physical address.
+///
+pub inline fn virtToPhys(virt: var) @typeOf(virt) {
```
DrDeano: not sure about pub inline. Porbs don't need the inline

src/kernel/mem.zig
```diff
@@ -15,10 +16,19 @@ extern var KERNEL_VADDR_START: *u32;
 extern var KERNEL_PHYSADDR_END: *u32;
 extern var KERNEL_PHYSADDR_START: *u32;
 
+// The boot-time offset that the virtual addresses are from the physical addresses
```
DrDeano: ///

src/kernel/mem.zig
```diff
@@ -1,4 +1,5 @@
 const multiboot = @import("multiboot.zig");
+const expectEqual = @import("std").testing.expectEqual;
```
DrDeano: I do prefer:
```Zig
const std = @import("std");
const expectEqual = std.testing.expectEqual;
```
As it make it easier to use `std` in the future

src/kernel/mem.zig
```diff
@@ -15,10 +16,19 @@ extern var KERNEL_VADDR_START: *u32;
...
+
 // The size of the fixed allocator used before the heap is set up. Set to 1MiB.
 const FIXED_ALLOC_SIZE = 1024 * 1024;
 
+/// The kernel's virtual address offset. It's assigned in the init function and this file's tests.
+/// We can't just use KERNEL_ADDR_OFFSET since using externs in the virtToPhys test is broken in
+/// release-safe. This is a workaround until that is fixed.
+var ADDR_OFFSET: usize = undefined;
+
 pub fn init(mb_info: *multiboot.multiboot_info_t) MemProfile {
```
DrDeano: Doc comment :)

src/kernel/mem.zig
SamTebbs33: Good idea, this was actually why the virtToPhys tests used to fail on release-fast and why I ended up needing a separate `ADDR_OFFSET` variable, so I have now removed it.

SamTebbs33: Ok scratch that, it was actually failing on release-safe so ignore all of the above :p

src/kernel/mem.zig
```diff
@@ -10,23 +11,32 @@ pub const MemProfile = struct {
...
+/// The size of the fixed allocator used before the heap is set up. Set to 1MiB.
 const FIXED_ALLOC_SIZE = 1024 * 1024;
 
 /// The kernel's virtual address offset. It's assigned in the init function and this file's tests.
 /// We can't just use KERNEL_ADDR_OFFSET since using externs in the virtToPhys test is broken in
 /// release-safe. This is a workaround until that is fixed.
 var ADDR_OFFSET: usize = undefined;
 
+///
+/// Initialise the system's memory profile based on linker symbols and the multiboot info struct.
```
DrDeano: Args and return doc missing

PR #107, opened 2019-11-05
Added run time tests for PIT

DrDeano:
Also changed the TTY a bit
Merged new run time testing


src/kernel/arch/x86/pit.zig
```diff
@@ -140,6 +148,35 @@ inline fn sendCommand(cmd: u8) void {
...
+///
+/// Read the current reload value of the PIT counter.
+///
+/// Arguments:
+///     counter: u16 - The counter to read from.
+///
+/// Return: u16
+///     The reload value currently in the PIT counter.
+///
+inline fn readCounter(comptime counter: u16) u16 {
```
SamTebbs33: What is the reload value? If it's not something simple then it may be worth adding some info regarding that.

src/kernel/arch/x86/pit.zig
```diff
@@ -140,6 +148,35 @@ inline fn sendCommand(cmd: u8) void {
...
+    sendCommand(u8(0));
+
+    const res = u16(arch.inb(counter)) | (u16(arch.inb(counter)) << 8);
+
+    arch.enableInterrupts();
+
+    return res;
+}
+
+inline fn read_back_command(comptime counter: u16) u8 {
```
SamTebbs33: Camelcase

src/kernel/arch/x86/pit.zig
```diff
@@ -140,6 +148,35 @@ inline fn sendCommand(cmd: u8) void {
...
+    comptime expect(counter == COUNTER_0_REGISTER or counter == COUNTER_1_REGISTER or counter == COUNTER_2_REGISTER);
+
+    sendCommand(u8(0xC2));
```
SamTebbs33: Is the cast necessary? If the parameter type is u8 then it shouldn't be needed.

src/kernel/arch/x86/pit.zig
```diff
@@ -148,111 +185,135 @@ inline fn sendCommand(cmd: u8) void {
 ///     IN data: u8              - The data to send.
 ///
 inline fn sendDateToCounter(comptime counter: u16, data: u8) void {
```
SamTebbs33: s/data/data

src/kernel/arch/x86/pit.zig
```diff
@@ -148,111 +185,135 @@ inline fn sendCommand(cmd: u8) void {
...
 }
 
 ///
 /// Set up a counter with a tick rate and mode of operation.
 ///
 /// Arguments:
-///     IN freq: u16   - The frequency that the counter operates at.
 ///     IN counter: u8 - Which counter is to be set up, either OCW_SELECT_COUNTER_0,
 ///                      OCW_SELECT_COUNTER_1 or OCW_SELECT_COUNTER_2 only.
+///     IN freq: u32   - The frequency that the counter operates at. Any frequency that is between
```
SamTebbs33: I would suggest throwing an error or doing a comptime check on the frequency instead of setting to a min/max value if it's outside the range

src/kernel/arch/x86/pit.zig
```diff
@@ -148,111 +185,135 @@ inline fn sendCommand(cmd: u8) void {
...
+    time_under_1_ns = ((u32(1000000000) % frequency) * u32(1000)) / frequency;
 
-    comptime const port: u16 = switch (counter) {
+    // Select the port to set up
+    comptime const port = switch (counter) {
         OCW_SELECT_COUNTER_0 => COUNTER_0_REGISTER,
         OCW_SELECT_COUNTER_1 => COUNTER_1_REGISTER,
         OCW_SELECT_COUNTER_2 => COUNTER_2_REGISTER,
-        else => unreachable,
+        else => @compileError("Invalid counter"),
```
SamTebbs33: Using an enum would remove the need for a compile error and for `counter` to be comptime, but I know you don't like enums. The enum could even have a function that returns its register.

src/kernel/arch/x86/pit.zig
```diff
@@ -274,12 +335,11 @@ pub fn getFrequency() u32 {
 ///
 pub fn init() void {
     log.logInfo("Init pit\n");
-    // Set up counter 0 at 1000hz in a square wave mode counting in binary
-    // TODO: https://github.com/ziglang/zig/issues/557, Need type defined
-    const freq: u32 = 10000;
+    // Set up counter 0 at 10000hz in a square wave mode counting in binary
+    const freq = u32(10000);
```
SamTebbs33: `const freq: u32 = 10000`? There are other occurrences of this in the file.

src/kernel/arch/x86/pit.zig
DrDeano: Been writing some python and forgot :P

src/kernel/arch/x86/pit.zig
DrDeano: Ill have ago and see how it looks

src/kernel/arch/x86/pit.zig
DrDeano: Forgot to remove this function bc I couldn't get it working

src/kernel/arch/x86/pit.zig
```diff
@@ -12,6 +13,25 @@ const irq = @import("irq.zig");
 const pic = @import("pic.zig");
 const log = @import("../../log.zig");
 
+/// The enum for selecting the counter
+const CounterSelect = enum {
+    /// Counter 0.
+    Counter_0,
```
SamTebbs33: Maybe `Counter1` instead of `Counter_1`?

src/kernel/arch/x86/pit.zig
```diff
@@ -149,44 +169,62 @@ inline fn sendCommand(cmd: u8) void {
...
+///
+/// Get the register port for the selected counter.
+///
+/// Arguments:
+///     IN counter: CounterSelect - The counter to get the register port.
+///
+/// Return: u16
+///     The register port for the selected counter.
+///
+fn getRegister(counter: CounterSelect) u16 {
```
SamTebbs33: These functions could be moved into the enum definition, see an example from the zig docs:

```
const Suit = enum {
    Clubs,
    Spades,
    Diamonds,
    Hearts,

    pub fn isClubs(self: Suit) bool {
        return self == Suit.Clubs;
    }
};
```

PR #108, opened 2019-11-05
No more assembly

### LINKED ISSUE #55: Replace x86 interrupt assembly with zig
It would be nice if we could replace the x86 interrupt assembly (irq_asm.s and isr_asm.s) with zig code. This should be possible using naked functions and inline assembly, so we precisely control what happens. The interrupt handler functions could be generated with a pattern similar to [this](https://godbolt.org/z/VOTEbG), but something simpler would be nice (perhaps just write them out manually?).

DrDeano:
Pure zig
Added doc comments
Closes #55


src/kernel/arch/x86/interrupts.zig
```diff
@@ -0,0 +1,90 @@
...
+
+extern fn irqHandler(ctx: *arch.InterruptContext) void;
+extern fn isrHandler(ctx: *arch.InterruptContext) void;
+
+///
+/// The main handler for all exceptions and interrupts. This will then go and call the correct
+/// handler for an ISR or IRQ.
+///
+/// Arguments:
+///     IN ctx: *arch.InterruptContext - Pointer to the exception context containing the contents
```
SamTebbs33: s/exception/interrupt as that's more general. s/register/registers

src/kernel/arch/x86/interrupts.zig
```diff
@@ -0,0 +1,90 @@
...
+
+///
+/// Generate the function that is the entry point for each exception/interrupt. This will then be
+/// used as the handler for the corresponding IDT entry.
+///
+/// Arguments:
+///     INT interrupt_num: u32 - The interrupt number to generate the function for.
+///
+/// Return: idt.InterruptHandler
+///     The sub function that is called for each interrupt/exception.
```
SamTebbs33: sub function?

DrDeano: stub

DrDeano:
The asm files have been removed?

PR #109, opened 2019-11-09
Add simple stacktrace logging

### LINKED ISSUE #44: Print stack trace when kernel panic
Here is one example:
https://andrewkelley.me/post/zig-stack-traces-kernel-panic-bare-bones-os.html

SamTebbs33:
This adds basic stacktrace logging when we trigger a kernel panic using a symbol table file created by readelf, which is then loaded as a module by the bootloader. Closes #44

### DIFF
--- a/make_map.sh
+++ b/make_map.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# Read the symbols from the binary, remove all the unnecessary columns with awk and emit to a map file
+readelf -s $1 | grep -F "FUNC" | awk '{$1=$3=$4=$5=$6=$7=""; print $0}' | sort -k 1 > $2
+echo "" >> $2
--- a/src/kernel/arch/x86/boot.zig
+++ b/src/kernel/arch/x86/boot.zig
@@ -99,7 +99,7 @@ export nakedcc fn start_higher_half() noreturn {
     asm volatile (
         \\.extern KERNEL_STACK_END
         \\mov $KERNEL_STACK_END, %%esp
-        \\mov %%esp, %%ebp
+        \\xor %%ebp, %%ebp
     );
 
     // Push the bootloader magic number and multiboot header address with virtual offset
--- a/src/kernel/panic.zig
+++ b/src/kernel/panic.zig
@@ -1,13 +1,449 @@
+const std = @import("std");
 const builtin = @import("builtin");
 const tty = @import("tty.zig");
 const arch = @import("arch.zig").internals;
 const log = @import("log.zig");
+const multiboot = @import("multiboot.zig");
+const mem = @import("mem.zig");
+const ArrayList = std.ArrayList;
+const testing = std.testing;
+
+/// The possible errors from panic code
+const PanicError = error{
+    /// The symbol file is of an invalid format.
+    /// This could be because it lacks whitespace, a column or required newline characters.
+    InvalidSymbolFile,
+};
+
+/// An entry within a symbol map. Corresponds to one entry in a symbole file
+const MapEntry = struct {
+    /// The address that the entry corresponds to
+    addr: usize,
+
+    /// The name of the function that starts at the address
+    func_name: []const u8,
+};
+
+const SymbolMap = struct {
+    symbols: ArrayList(MapEntry),
+
+    ///
+    /// Initialise an empty symbol map.
+    ///
+    /// Arguments:
+    ///     IN allocator: *std.mem.Allocator - The allocator to use to initialise the array list.
+    ///
+    /// Return: SymbolMap
+    ///     The symbol map.
+    ///
+    pub fn init(allocator: *std.mem.Allocator) SymbolMap {
+        return SymbolMap{
+            .symbols = ArrayList(MapEntry).init(allocator),
+        };
+    }
+
+    ///
+    /// Deinitialise the symbol map, freeing all memory used.
+    ///
+    pub fn deinit(self: *SymbolMap) void {
+        self.symbols.deinit();
+    }
+
+    ///
+    /// Add a symbol map entry with a name and address.
+    ///
+    /// Arguments:
+    ///     IN name: []const u8 - The name of the entry.
+    ///     IN addr: usize - The address for the entry.
+    ///
+    /// Error: std.mem.Allocator.Error
+    ///      * - See ArrayList.append
+    ///
+    pub fn add(self: *SymbolMap, name: []const u8, addr: u32) !void {
+        try self.addEntry(MapEntry{ .addr = addr, .func_name = name });
+    }
+
+    pub fn addEntry(self: *SymbolMap, entry: MapEntry) !void {
+        try self.symbols.append(entry);
+    }
+
+    ///
+    /// Search for the function name associated with the address.
+    ///
+    /// Arguments:
+    ///     IN addr: usize - The address to search for.
+    ///
+    /// Return: ?[]const u8
+    ///     The function name associated with that program address, or null if one wasn't found.
+    ///
+    pub fn search(self: *const SymbolMap, addr: usize) ?[]const u8 {
+        if (self.symbols.count() == 0)
+            return null;
+        // Find the first element whose address is greater than addr
+        var previous_name: ?[]const u8 = null;
+        var it = self.symbols.iterator();
+        while (it.next()) |entry| {
+            if (entry.addr > addr)
+                return previous_name;
+            previous_name = entry.func_name;
+        }
+        return previous_name;
+    }
+};
+
+var symbol_map: ?SymbolMap = null;
+
+///
+/// Log a stacktrace address. Logs "(no symbols are available)" if no symbols are available,
+/// "?????" if the address wasn't found in the symbol map, else logs the function name.
+///
+/// Arguments:
+///     IN addr: usize - The address to log.
+///
+fn logTraceAddress(addr: usize) void {
+    const str = if (symbol_map) |syms| syms.search(addr) orelse "?????" else "(no symbols available)";
+    log.logError("{x}: {}\n", addr, str);
+}
 
 pub fn panic(trace: ?*builtin.StackTrace, comptime format: []const u8, args: ...) noreturn {
     @setCold(true);
-    arch.disableInterrupts();
-    log.logInfo("KERNEL PANIC\n");
-    log.logInfo(format, args);
-    log.logInfo("HALTING\n");
+    log.logError("Kernel panic: " ++ format ++ "\n", args);
+    if (trace) |trc| {
+        var last_addr: u64 = 0;
+        for (trc.instruction_addresses) |ret_addr| {
+            if (ret_addr != last_addr) logTraceAddress(ret_addr);
+            last_addr = ret_addr;
+        }
+    } else {
+        const first_ret_addr = @returnAddress();
+        var last_addr: u64 = 0;
+        var it = std.debug.StackIterator.init(first_ret_addr);
+        while (it.next()) |ret_addr| {
+            if (ret_addr != last_addr) logTraceAddress(ret_addr);
+            last_addr = ret_addr;
+        }
+    }
     arch.haltNoInterrupts();
+}
+
+///
+/// Parse a hexadecimal address from the pointer up until the end pointer. Must be terminated by a
+/// whitespace character.
+///
+/// Arguments:
+///     INOUT ptr: *[*]const u8 - The address at which to start looking, updated after all
+///         characters have been consumed.
+///     IN end: *const u8 - The end address at which to start looking. A whitespace character must
+///         be found before this.
+///
+/// Return: usize
+///     The address parsed.
+///
+/// Error: PanicError || std.fmt.ParseUnsignedError
+///     PanicError.InvalidSymbolFile: A terminating whitespace wasn't found before the end address.
+///     std.fmt.ParseUnsignedError: See std.fmt.parseInt
+///
+fn parseAddr(ptr: *[*]const u8, end: *const u8) !usize {
+    const addr_start = ptr.*;
+    ptr.* = try parseNonWhitespace(ptr.*, end);
+    const len = @ptrToInt(ptr.*) - @ptrToInt(addr_start);
+    const addr_str = addr_start[0..len];
+    return try std.fmt.parseInt(usize, addr_str, 16);
+}
+
+///
+/// Parse a single character. The address given cannot be greater than or equal to the end address
+/// given.
+///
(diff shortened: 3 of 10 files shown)


src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        const mod_str = mod_str_ptr[0..strlen(mod_str_ptr)];
+        if (std.mem.eql(u8, mod_str, "kernel.map")) {
+            kmap_start = mod_start;
+            kmap_end = mod_end;
+            break;
+        }
+    }
+    std.debug.assert(kmap_start != 0 and kmap_end != 0);
+
+    var syms = SymbolMap.init(allocator);
```
DrDeano: Could have an `errdefer` to deinit the table if there was an error creating an entry. idk if you need all entries to perform a back trace. Or, lets say an entry halfway through is corrupt, could the loop below continue with the other entries and just insert a null for missing ones?

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        for (self.symbols.toSlice()[0..]) |entry| {
+            if (entry.addr > addr)
+                return previous_name;
+            previous_name = entry.func_name;
+        }
+        return previous_name;
+    }
+};
+
+pub fn init(mem_profile: *const mem.MemProfile, allocator: *std.mem.Allocator) !void {
```
DrDeano: Doc comment

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+}
+
+fn strlen(chars: [*]u8) usize {
+    var i: usize = 0;
+    while (chars[i] != 0) : (i += 1) {}
+    return i;
+}
+
+fn logTraceAddress(addr: usize) void {
+    var str: []const u8 = "(no symbols available)"[0..];
```
DrDeano: Cheeky one liner ;)
```Zig
const str = if (symbol_map) |syms| syms.search(addr) orelse "?????" else "(no symbols available)";
log.logError("{x}: {}\n", addr, str);
```
This removed the type and `[0..]`

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        if (std.mem.eql(u8, mod_str, "kernel.map")) {
+            kmap_start = mod_start;
+            kmap_end = mod_end;
+            break;
+        }
+    }
+    std.debug.assert(kmap_start != 0 and kmap_end != 0);
+
+    var syms = SymbolMap.init(allocator);
+    var i: u32 = kmap_start;
```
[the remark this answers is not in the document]
DrDeano: Plus you don't need the type

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        const addr_str = @intToPtr([*]u8, addr_start)[0..(i - addr_start)];
+        const addr = try std.fmt.parseInt(u32, addr_str, 16);
+        // Skip to the function name in the second column
+        while (std.fmt.isWhiteSpace(@intToPtr(*u8, i).*)) : (i += 1) {}
+        const name_start = i;
+        while (!std.fmt.isWhiteSpace(@intToPtr(*u8, i).*)) : (i += 1) {}
+        const name = @intToPtr([*]u8, name_start)[0..(i - name_start)];
+        try syms.add(name, addr);
+        // Skip to the next line
+        while (std.fmt.isWhiteSpace(@intToPtr(*u8, i).*)) : (i += 1) {}
```
DrDeano: You don't need this while isWhiteSpace as the one at the start does this

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        for (trc.instruction_addresses) |ret_addr| {
+            if (ret_addr != last_addr) logTraceAddress(ret_addr);
+            last_addr = ret_addr;
+        }
+    } else {
+        const first_ret_addr = @returnAddress();
+        var last_addr: u64 = 0;
+        var it = std.debug.StackIterator.init(first_ret_addr);
+        while (it.next()) |ret_addr| {
+            if (ret_addr != last_addr) logTraceAddress(ret_addr);
```
DrDeano: Surly you don't need this comparison as the iterator will return null for the last case?

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+    if (mem_profile.boot_modules.len < 1)
+        return;
+    var kmap_start: u32 = 0;
+    var kmap_end: u32 = 0;
+    for (mem_profile.boot_modules) |module| {
+        const mod_start = mem.physToVirt(module.mod_start);
+        const mod_end = mem.physToVirt(module.mod_end);
+        const mod_str_ptr = mem.physToVirt(@intToPtr([*]u8, module.cmdline));
+        const mod_str = mod_str_ptr[0..strlen(mod_str_ptr)];
+        if (std.mem.eql(u8, mod_str, "kernel.map")) {
```
DrDeano:
```Zig
std.mem.eql(u8, std.mem.toSlice(u8, mod_str_ptr), "kernel.map")
```
Then can remove the `strlen` function

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        }
+        return previous_name;
+    }
+};
+
+pub fn init(mem_profile: *const mem.MemProfile, allocator: *std.mem.Allocator) !void {
+    log.logInfo("Init panic\n");
+    // Exit if we haven't loaded all debug modules
+    if (mem_profile.boot_modules.len < 1)
+        return;
```
DrDeano: If this is correct and no error should be returned, then should this also log `Done`

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+            kmap_end = mod_end;
+            break;
+        }
+    }
+    std.debug.assert(kmap_start != 0 and kmap_end != 0);
+
+    var syms = SymbolMap.init(allocator);
+    var i: u32 = kmap_start;
+    while (i < kmap_end) {
+        while (std.fmt.isWhiteSpace(@intToPtr(*u8, i).*)) : (i += 1) {}
```
DrDeano: Maybe include a bounds check in the whiles. So maybe abstract this out to a function that returns the index

DrDeano: Maybe could have ago at using the `std.fmt.trim` function. Idk if it will work in this case.

DrDeano: Also, is it possible to convert `kmap_start` to a pointer instead of a integer so you don't have to use `@intToPtr` everywhere. So can treat the `kmap_*` as a string buffer

src/kernel/arch/x86/boot.zig
```diff
@@ -99,7 +99,7 @@ export nakedcc fn start_higher_half() noreturn {
     asm volatile (
         \\.extern KERNEL_STACK_END
         \\mov $KERNEL_STACK_END, %%esp
-        \\mov %%esp, %%ebp
+        \\xor %%ebp, %%ebp
```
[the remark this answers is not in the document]
SamTebbs33: It actually sets it to 0, which is required for the stacktracing code as that's where it knows to stop.

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+    pub fn add(self: *SymbolMap, name: []const u8, addr: u32) !void {
+        try self.symbols.append(MapEntry{ .addr = addr, .func_name = name });
+    }
+
+    pub fn search(self: *const SymbolMap, addr: usize) ?[]const u8 {
+        if (self.symbols.count() == 0)
+            return null;
+        // Find the first element whose address is greater than addr
+        var previous_name: ?[]const u8 = null;
+        for (self.symbols.toSlice()[0..]) |entry| {
```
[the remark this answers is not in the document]
SamTebbs33: Nice idea, I'll look into it

src/kernel/panic.zig
```diff
@@ -1,13 +1,148 @@
...
+        const mod_end = mem.physToVirt(module.mod_end);
+        const mod_str_ptr = mem.physToVirt(@intToPtr([*]u8, module.cmdline));
+        const mod_str = mod_str_ptr[0..strlen(mod_str_ptr)];
+        if (std.mem.eql(u8, mod_str, "kernel.map")) {
+            kmap_start = mod_start;
+            kmap_end = mod_end;
+            break;
+        }
+    }
+    std.debug.assert(kmap_start != 0 and kmap_end != 0);
```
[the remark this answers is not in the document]
SamTebbs33: I think I will change this assertion to an early exit as that mimicks the module count check above.

src/kernel/panic.zig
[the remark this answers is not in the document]
SamTebbs33: True, I'll rename it to `file_index` as that's more descriptive of what it is.

src/kernel/panic.zig
SamTebbs33: It certainly should, well spotted

src/kernel/panic.zig
SamTebbs33: That would be much better

src/kernel/panic.zig
SamTebbs33: We do need all entries for the map to be valid so I'll make an `errdefer` and make the loop return a `InvalidSymbolMap` error if it is corrupt.

src/kernel/panic.zig
SamTebbs33: I'll abstract it into smaller functions. That would help with the invalid format detection too. Nice idea with string buffer, will look into it.

src/kernel/panic.zig
SamTebbs33: Oh nice, does that work? I didn't try that as I thought the compiler would complain that the string literals were of different sizes.

src/kernel/panic.zig
SamTebbs33: I added this check so that it wouldn't print loads of duplicate function names in the case of a recursive function, I wouldn't be against removing it though.

src/kernel/panic.zig
```diff
@@ -1,13 +1,382 @@
...
+}
+
+///
+/// Parse a name from the pointer up until the end pointer. Must be terminated by a whitespace character.
+///
+/// Arguments:
+///     INOUT ptr: *[*]const u8 - The address at which to start looking, updated after all characters have been consumed.
+///     IN: end: *const u8 - The end address at which to start looking. A whitespace character must be found before this.
+///
+/// Return: The name parsed.
```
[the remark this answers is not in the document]
DrDeano: There are a couple of places

src/kernel/panic.zig
```diff
@@ -1,13 +1,382 @@
...
+    while (std.fmt.isWhiteSpace(try parseChar(ptr + i, end))) : (i += 1) {}
+    return ptr + i;
+}
+
+///
+/// Parse until a whitespace character. Must be terminated by a whitespace character before the end address.
+///
+/// Arguments:
+///     IN ptr: [*]const u8 - The address at which to start looking.
+///     IN: end: *const u8 - The end address at which to start looking. A whitespace character must be found before this.
```
DrDeano: You have colon that you don't need (after the IN). Also found in a couple of places

src/kernel/panic.zig
```diff
@@ -1,13 +1,382 @@
+const std = @import("std");
```
DrDeano: There are some doc comments that are longer than 100 characters.

make_map.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+readelf -s $1 | grep -F "FUNC" | awk '{$1=$3=$4=$5=$6=$7=""; print $0}' | sort -k 1 > $2
```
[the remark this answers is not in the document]
SamTebbs33: Sure :)

src/kernel/panic.zig
```diff
@@ -1,13 +1,382 @@
...
+    /// Arguments:
+    ///     IN name: []const u8 - The name of the entry.
+    ///     IN addr: usize - The address for the entry.
+    ///
+    /// Error: std.mem.Allocator.Error
+    //      * - See ArrayList.append
+    ///
+    pub fn add(self: *SymbolMap, name: []const u8, addr: u32) !void {
+        try self.symbols.append(MapEntry{ .addr = addr, .func_name = name });
+    }
```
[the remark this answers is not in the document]
SamTebbs33: Good idea

src/kernel/panic.zig
```diff
@@ -1,13 +1,382 @@
...
+    var ptr = str.ptr;
+    testing.expectEqualSlices(u8, try parseName(&ptr, end), "func_name");
+}
+
+test "parseName fails" {
+    // Test without terminating whitespace, should fail
+    const str: []const u8 = "func_name";
+    const end = @ptrCast(*const u8, str.ptr + 9);
+    var ptr = str.ptr;
+    testing.expectError(PanicError.InvalidSymbolFile, parseName(&ptr, end));
```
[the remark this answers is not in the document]
SamTebbs33: What do you mean exactly?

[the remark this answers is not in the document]
DrDeano: I actually don't remember why lol. Ignore this

src/kernel/panic.zig
```diff
@@ -309,10 +360,16 @@ test "parseAddr" {
...
 test "parseAddr fails" {
     // Test without terminating whitespace, should fail
-    const str: []const u8 = "1a2b3c4d";
-    const end = @ptrCast(*const u8, str.ptr + 9);
+    var str: []const u8 = "1a2b3c4d";
+    var end = @ptrCast(*const u8, str.ptr + 9);
     var ptr = str.ptr;
     testing.expectError(PanicError.InvalidSymbolFile, parseAddr(&ptr, end));
+
+    // Test with invalid hexdecimal integer
```
DrDeano: Might be better to have this as a separate test and the name will convey what this is testing. Then can keep the `const` variables.

src/kernel/panic.zig
```diff
@@ -350,18 +407,33 @@ test "parseMapEntry" {
 
 test "parseMapEntry fails" {
     // Test without terminating whitespace, should fail
-    const str: []const u8 = "1a2b3c4d func_name";
+    var str: []const u8 = "1a2b3c4d func_name";
     var ptr = str.ptr;
     testing.expectError(PanicError.InvalidSymbolFile, parseMapEntry(&ptr, @ptrCast(*const u8, str.ptr + 18)));
+
+    // Test with invalid address
```
DrDeano: Same here, have then as seperete tests

PR #111, opened 2019-11-10
Fix T(x) style casts

### LINKED ISSUE #110: Replace old style casts with `@as`
Zig has recently removed the `T(x)` style casts and replaced them with an `@as` builtin: https://ziglang.org/documentation/master/#as

Unnecessary casts will be removed entirely, casts in the form of `const/var x = T(y)` will be replaced with `const/var x: T = y` and anything else will be replaced with `@as`.

SamTebbs33:
This patch fixes the `T(x)` style casts in the codebase. Unnecessary casts were removed, those of the form `const/var x = T(y)` were replaced with `const/var x: T = y` and other were replaced with the new `@as` builtin. Closes #110

### DIFF
--- a/src/kernel/arch/x86/gdt.zig
+++ b/src/kernel/arch/x86/gdt.zig
@@ -346,7 +346,7 @@ var gdt_ptr: GdtPtr = GdtPtr{
 var tss: TtsEntry = TtsEntry{
     .prev_tss = 0,
     .esp0 = 0,
-    .ss0 = u32(KERNEL_DATA_OFFSET),
+    .ss0 = KERNEL_DATA_OFFSET,
     .esp1 = 0,
     .ss1 = 0,
     .esp2 = 0,
@@ -370,7 +370,7 @@ var tss: TtsEntry = TtsEntry{
     .gs = 0,
     .ldtr = 0,
     .trap = 0,
-    .io_permissions_base_offset = u16(@sizeOf(TtsEntry)),
+    .io_permissions_base_offset = @sizeOf(TtsEntry),
 };
 
 ///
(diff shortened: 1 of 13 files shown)


src/kernel/arch/x86/gdt.zig
```diff
@@ -532,106 +532,106 @@ test "makeEntry alternating bit pattern" {
...
-    expectEqual(u32(0), tss.es);
-    expectEqual(u32(0), tss.cs);
-    expectEqual(u32(0), tss.ss);
-    expectEqual(u32(0), tss.ds);
-    expectEqual(u32(0), tss.fs);
-    expectEqual(u32(0), tss.gs);
-    expectEqual(u32(0), tss.ldtr);
-    expectEqual(u16(0), tss.trap);
-    expectEqual(u16(@sizeOf(TtsEntry)), tss.io_permissions_base_offset);
+    expectEqual(@as(u32, 0), tss.prev_tss);
```
DrDeano: I think if you swap the arguments around you may not need to cast

src/kernel/arch/x86/paging.zig
```diff
@@ -190,10 +190,10 @@ fn mapDirEntry(dir: *Directory, virt_start: usize, virt_end: usize, phys_start:
     var dir_entry = &dir.entries[entry];
     dir_entry.* |= DENTRY_PRESENT;
     dir_entry.* |= DENTRY_WRITABLE;
-    dir_entry.* &= ~u32(DENTRY_USER);
+    dir_entry.* &= ~@as(u32, DENTRY_USER);
     dir_entry.* |= DENTRY_WRITE_THROUGH;
-    dir_entry.* &= ~u32(DENTRY_CACHE_DISABLED);
-    dir_entry.* &= ~u32(DENTRY_4MB_PAGES);
+    dir_entry.* &= ~@as(u32, DENTRY_CACHE_DISABLED);
+    dir_entry.* &= ~@as(u32, DENTRY_4MB_PAGES);
```
SamTebbs33: I don't think these casts are necessary

src/kernel/arch/x86/gdt.zig
SamTebbs33: True, but then if I e.g. change 0 to 1 then it fails with "expected 0, found 1", but in reality 1 is the expected value.

DrDeano: True

SamTebbs33: It would be nice to not have to cast, but I reckon the expectEqual function needs to be changed for that.

PR #122, opened 2020-01-08
Unittest in build arch

### LINKED ISSUE #102: Unit testing on architecture
Run the architecture specific unit tests in the architecture they will be built for.

See: https://www.reddit.com/r/Zig/comments/d7fl0x/cross_compiled_unit_tests

DrDeano:
Also change the build process to abstract out the Linux part of the building. This will later help with the native Windows build.

Closes #102

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,14 +19,14 @@ jobs:
         wget $(curl -s 'https://ziglang.org/download/index.json' | python3 -c "import sys, json; print(json.load(sys.stdin)['master']['x86_64-linux']['tarball'])")
         sudo apt-get install mtools
         tar -xvf zig*
-    - name: Build kernel
-      run: zig*/zig build ${{ matrix.build_mode }}
-    - name: Run unit tests
-      run: zig*/zig build test ${{ matrix.build_mode }}
     - name: Install qemu
       run: |
         sudo apt-get update
         sudo apt-get install qemu qemu-system --fix-missing
+    - name: Build kernel
+      run: zig*/zig build ${{ matrix.build_mode }}
+    - name: Run unit tests
+      run: zig*/zig build test ${{ matrix.build_mode }}
     - name: Run runtime tests
       run: zig*/zig build test -Drt-test=true ${{ matrix.build_mode }}
     - name: Check formatting
--- a/build.zig
+++ b/build.zig
@@ -9,9 +9,17 @@ const Mode = builtin.Mode;
 
 pub fn build(b: *Builder) !void {
     const target = Target{
-        .Cross = std.build.CrossTarget{
+        .Cross = Target.Cross{
             .arch = .i386,
             .os = .freestanding,
+            .abi = .gnu,
+        },
+    };
+
+    const test_target = Target{
+        .Cross = Target.Cross{
+            .arch = .i386,
+            .os = .linux,
             .abi = .gnu,
         },
     };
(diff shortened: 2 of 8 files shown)


.github/workflows/main.yml
```diff
@@ -19,14 +19,16 @@ jobs:
...
-      run: zig*/zig build ${{ matrix.build_mode }}
-    - name: Run unit tests
-      run: zig*/zig build test ${{ matrix.build_mode }}
     - name: Install qemu
       run: |
         sudo apt-get update
         sudo apt-get install qemu qemu-system --fix-missing
+    - name: chmod
+      run: chmod +x ./makeiso.sh
+    - name: Build kernel
```
SamTebbs33: I had the build and test steps before downloading qemu as they don't depend on it and it means that if installing qemu fails for whatever reason, we can still test and build. I can see an advantage of installing qemu first though as it means installation failures will appear sooner and mean you can re-run sooner. What do you think?

build.zig
```diff
@@ -88,31 +106,14 @@ pub fn build(b: *Builder) !void {
...
-        unit_tests.addPackagePath("constants", constants_path);
-        unit_tests.addBuildOption(bool, "rt_test", rt_test);
-        unit_tests.addBuildOption([]const u8, "mock_path", mock_path);
-        unit_tests.addBuildOption([]const u8, "arch_mock_path", arch_mock_path);
-        test_step.dependOn(&unit_tests.step);
-    }
-
     const debug_step = b.step("debug", "Debug with gdb and connect to a running qemu instance");
-    const symbol_file_arg = try std.mem.join(b.allocator, " ", &[_][]const u8{ "symbol-file", elf_path });
+    const symbol_file_arg = try std.mem.join(b.allocator, " ", &[_][]const u8{ "symbol-file", exec.getOutputPath() });
```
SamTebbs33: Much nicer :+1:

makeiso.sh
```diff
@@ -0,0 +1,31 @@
...
+OUTPUT_FILE=$5
+
+MAP_FILE=$MODULES_DIR/'kernel.map'
+
+exit_missing() {
+    printf "$_ must be installed\n";
+    exit 1;
+}
+
+# Check dependancies
```
SamTebbs33: dependencies*

build.zig
```diff
@@ -9,58 +9,76 @@ const Mode = builtin.Mode;
...
     const build_mode = b.standardReleaseOptions();
     const rt_test = b.option(bool, "rt-test", "enable/disable runtime testing") orelse false;
 
-    const main_src = "src/kernel/kmain.zig";
-
     const exec = b.addExecutable("pluto", main_src);
-    const constants_path = try fs.path.join(b.allocator, &[_][]const u8{ "src/kernel/arch", target_str, "constants.zig" });
     exec.addPackagePath("constants", constants_path);
-    exec.setBuildMode(build_mode);
+    exec.setOutputDir("zig-cache");
```
SamTebbs33: Do we need to set this explicitly? zig-cache is the default anyway and I'm not sure how this will interfere with the use specifying a custom output dir when running `zig build`.

DrDeano: I need to set is explicitly bc I use `getOutputPath()` and this is only set when `make` is called at the end.

.github/workflows/main.yml
DrDeano: The testing is now done in qemu, so need to download before testing. Also thought that the download steps could be grouped together

build.zig
SamTebbs33: I see, that's a weird restriction. The only problem is that it outputs to zig-cache even if the user specifies --cache-dir when building. It's fine if there isn't a workaround but it would be nice if we didn't have to set things to something hardcoded.

.github/workflows/main.yml
SamTebbs33: Yep that makes sense.

.github/workflows/main.yml
```diff
@@ -19,14 +19,16 @@ jobs:
...
-    - name: Build kernel
-      run: zig*/zig build ${{ matrix.build_mode }}
-    - name: Run unit tests
-      run: zig*/zig build test ${{ matrix.build_mode }}
     - name: Install qemu
       run: |
         sudo apt-get update
         sudo apt-get install qemu qemu-system --fix-missing
+    - name: chmod
+      run: chmod +x ./makeiso.sh
```
SamTebbs33: Giving the file execute permissions and committing that change would make the chmod step unnecessary.

build.zig
DrDeano: If you specify `--cache-dir` it overrides the cache directory even if one is hardcoded.

SamTebbs33: It does but it still puts the pluto binary in zig-cache. Could we set the output dir to whatever cache dir has been set?

DrDeano: Ah yes, it does.

PR #131, opened 2020-04-13
use /usr/bin/env bash instead of /bin/bash

marler8997:
`!/usr/bin/env bash` works on more systems than `!/bin/bash`.  I use NixOS which doesn't have `/bin/bash` and I know some BSDs also don't have `/bin/bash`.

### DIFF
--- a/makeiso.sh
+++ b/makeiso.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 BOOT_DIR=$1
 MODULES_DIR=$2


SamTebbs33:
Thanks for spotting this. Would you mind force-pushing to the branch to re-launch the checks? They seem to be broken at the moment.

SamTebbs33:
I think you will need to enable GitHub actions in your fork for the checks to happen

marler8997:
It looks like actions are already enabled:
image

SamTebbs33:
I believe that the problem lies in our workflow file and should be fixed by #132 . Once that is merged, the checks should be run once you reopen the PR.

marler8997:
Looks like after I rebased to get #132 that fixed the issue.

SamTebbs33:
Brilliant, thanks for this :)

PR #132, opened 2020-04-14
Run CI checks on pull requests and pushes

SamTebbs33:
This patch enables GitHub Actions status checks on pull requests from forks.﻿


SamTebbs33:
I realised that this would stop checks from happening on branches over than master and develop, fixed by using a wildcard.

Issue #35, opened 2019-06-03
Multitasking


SamTebbs33:
We'll need a rudimentary cooperative multitasking system that presents an interface to a program, allowing it to have its state stored, relinquish control to another task. When control is returned to the task in the future, its state should be restored.

Control could be relinquished by sending an interrupt since the interrupt handling system already does a lot of the work of saving the CPU state, and this would lay the foundation for a more advanced multitasking and scheduling system, where all task transitions occur via syscalls and timer interrupts.

Preemptive scheduling with a timer interrupt should probably be done in a separate issue.

Depends on #36 if interrupts are the only way for a program to relinquish control.

SamTebbs33:
Closed by #138

PR #138, opened 2020-04-27
Initial scheduler

DrDeano:
Fix TSS

Also change to .{} syntax where appropriate.
Added the SS segment
Fixed spelling

Refactoring GDT

Handlers return the stack pointer that will be used to restore the tasks stack, normal handlers will return the same stack pointer it was called with where task switching will return the stack pointer of the next task and restore its state using the interrupt stub.

Initial scheduler done


src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    ///
+    /// Save a new stack pointer for this task.
+    ///
+    /// Arguments:
+    ///     IN/OUT self: *Self - The pointer to self.
+    ///     IN esp: u32        - The new stack pointer to be saved.
+    ///
+    pub fn saveEsp(self: *Self, esp: u32) void {
```
SamTebbs33, line 132: I think this should also be moved to the arch file.

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+fn pickNextTask() void {
+    // Get the next task
+    if (tasks.pop()) |new_task_node| {
+        var next_task = new_task_node.data;
+        tasks.destroyNode(new_task_node, alloc);
+
+        // Now that we have the new task and deleted its node, there "will" be space for another
+        // so the allocation shouldn't fail as interrupts are disabled.
+        // So if there is a failure, then wow :P This will be the case for multiprocessor, but
+        // for a later time
```
SamTebbs33, line 172: I think `catch unreachable` is justified in this case, but that can come down to taste :)

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+/// Create a new task and add it to the scheduling queue. No locking.
+///
+/// Arguments:
+///     IN comptime entry_point: var - The entry point into the task. This must be a function.
+///
+/// Error: std.mem.Allocator.Error
+///     OutOfMemory - If there isn't enough memory for the a task/stack. Any memory allocated will
+///                   be freed on return.
+///
+pub fn createNewTask(comptime entry_point: var) Allocator.Error!void {
```
SamTebbs33, line 205: Perhaps this function should be made to just add to the task queue since task initialisation is done with `Task.init()` anyway. If so then a suitable new name for it could be `schedule`.

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+    errdefer task.destroy();
+
+    var elem = try tasks.createNode(task, alloc);
+    tasks.prepend(elem);
+}
+
+///
+/// Schedule a task switch. This will first save the the current tasks stack pointer, then will
+/// pick the next task to be run and return the stack pointer for the next task to be loaded into
+/// ESP to load the next task stack to pop off its state. Interrupts are assumed disabled.
```
SamTebbs33, line 219: ESP is arch-specific :)

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+        // for a later time
+
+        // Put the current task back in the queue, at the back
+        var cur_elem = tasks.createNode(current_task, alloc) catch |e| {
+            panic(@errorReturnTrace(), "Error in pickNextTask, can't save previous task: {}\n", .{e});
+        };
+        tasks.prepend(cur_elem);
+
+        // Set the next task
+        current_task = next_task;
```
SamTebbs33, line 181: Making this return the next task rather than set the global variable would make it more generic and usable in contexts when you don't want to set it

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+/// ESP to load the next task stack to pop off its state. Interrupts are assumed disabled.
+///
+/// Argument:
+///     IN ctx: *arch.CpuState - Pointer to the exception context containing the contents
+///                              of the registers at the time of a exception.
+///
+/// Return: u32
+///     The new stack pointer to the next stack of the next task.
+///
+pub fn schedule(ctx: *arch.CpuState) u32 {
```
SamTebbs33, line 228: In my mind "schedule" means put a task on the queue. Would you consider changing it to be something like "switchTask" with it taking the task to switch to, then anyone wanting to switch to the next would call "pickNextTask" then call this to perform the actual switch?

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+    // Pick the next task and update the current task
+    pickNextTask();
+
+    // Context switch in the interrupt stub handler which will pop the next task state off the
+    // stack
+    return current_task.esp;
+}
+
+///
+/// Initialise the scheduler.
```
SamTebbs33, line 241: Perhaps add some details on how the current task and idle task are set up.

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+///
+/// Arguments:
+///     IN/OUT allocator: *Allocator - The allocator to use when needing to allocate memory.
+///
+/// Error: std.mem.Allocator.Error
+///     OutOfMemory - There is no more memory. Any memory allocated will be freed on return.
+///
+pub fn init(allocator: *Allocator) Allocator.Error!void {
+    log.logInfo("Init schedule\n", .{});
```
SamTebbs33, line 250: s/schedule/scheduler ?

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+
+    setup_test();
+
+    var task = try Task.init(test_fn1);
+    defer task.destroy();
+
+    // Will allocate the first PID 0
+    expectEqual(task.pid, 0);
+
+    // This looks pointless tbh as it is just copy paste from init function
```
SamTebbs33, line 338: Just like Task.init I think this should be moved to the arch file

src/kernel/arch/x86/rtc.zig
```diff
@@ -206,14 +207,20 @@ fn readRtc() DateTime {
 /// The interrupt handler for the RTC.
 ///
 /// Arguments:
-///     IN ctx: *arch.InterruptContext - Pointer to the interrupt context containing the contents
+///     IN ctx: *arch.CpuState - Pointer to the interrupt context containing the contents
 ///                                      of the register at the time of the interrupt.
 ///
-fn rtcHandler(ctx: *arch.InterruptContext) void {
+fn rtcHandler(ctx: *arch.CpuState) u32 {
```
SamTebbs33, line 213: Just for curiosity and learning's sake, why did you choose the RTC instead of the PIT for firing off task switches?

src/kernel/arch/x86/interrupts.zig
```diff
@@ -40,7 +41,9 @@ export fn commonStub() callconv(.Naked) void {
         \\mov   %%esp, %%eax
         \\push  %%eax
         \\call  handler
-        \\pop   %%eax
+        \\add   $0x4, %%esp
```
SamTebbs33, line 44: The add becomes redundant when you set esp on the next line

src/kernel/arch/x86/interrupts.zig
```diff
@@ -40,7 +41,9 @@ export fn commonStub() callconv(.Naked) void {
...
+        \\mov   %%eax, %%esp
```
SamTebbs33, line 45: Clever workaround for popa not restoring esp :+1:

src/kernel/arch/x86/arch.zig
```diff
@@ -258,18 +258,9 @@ pub fn init(mb_info: *multiboot.multiboot_info_t, mem_profile: *const MemProfile
...
-    _ = @import("idt.zig");
-    _ = @import("pic.zig");
-    _ = @import("isr.zig");
-    _ = @import("irq.zig");
-    _ = @import("pit.zig");
-    _ = @import("cmos.zig");
-    _ = @import("rtc.zig");
-    _ = @import("syscalls.zig");
-    _ = @import("paging.zig");
+    std.meta.refAllDecls(@This());
```
SamTebbs33, line 265: Oooh nice :)

SamTebbs33:
Part of #35

src/kernel/arch/x86/interrupts.zig
DrDeano: Good point +1

src/kernel/arch/x86/rtc.zig
DrDeano: The PIT is more accurate and so can be used as a time since boot and stuff as adding a scheduler to the handler will skew the ticks. But mostly just a design chose really, it doesn't really matter. We will probs move to the APIC soon so this may change

src/kernel/scheduler.zig
DrDeano: This will be moved when the task is a separate file in the arch folder, leaving the scheduler file in the kernel folder

src/kernel/arch/x86/arch.zig
```diff
@@ -258,18 +258,9 @@ pub fn init(mb_info: *multiboot.multiboot_info_t, mem_profile: *const MemProfile
 
     syscalls.init();
 
-    enableInterrupts();
+    //enableInterrupts();
```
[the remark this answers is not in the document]
DrDeano: Yep

src/kernel/arch/x86/interrupts.zig
```diff
@@ -3,22 +3,22 @@ const syscalls = @import("syscalls.zig");
 const irq = @import("irq.zig");
 const idt = @import("idt.zig");
 
-extern fn irqHandler(ctx: *arch.InterruptContext) void;
-extern fn isrHandler(ctx: *arch.InterruptContext) void;
+extern fn irqHandler(ctx: *arch.CpuState) u32;
```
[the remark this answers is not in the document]
DrDeano: sure

src/kernel/arch/x86/paging.zig
```diff
@@ -361,9 +363,27 @@ pub fn unmap(virtual_start: usize, virtual_end: usize, dir: *Directory) (std.mem
 /// Called when a page fault occurs.
 ///
 /// Arguments:
-///     IN state: *arch.InterruptContext - The CPU's state when the fault occurred.
-///
-fn pageFault(state: *arch.InterruptContext) void {
+///     IN state: *arch.CpuState - The CPU's state when the fault occurred.
+///
+fn pageFault(state: *arch.CpuState) u32 {
```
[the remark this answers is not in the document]
DrDeano: Effort lel, So move the change to returning a stack pointer to a diff PR

src/kernel/vmm.zig
```diff
@@ -333,7 +333,7 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
 ///     INOUT allocator: *std.mem.Allocator - The allocator to use when needing to allocate memory
 ///     IN comptime Payload: type - The type of the data to pass as a payload to the virtual memory manager
 ///     IN mapper: Mapper - The memory mapper to call when allocating and free virtual memory
-///     IN payload: Paylaod - The payload data to pass to the virtual memory manager
+///     IN payload: Payload - The payload data to pass to the virtual memory manager
```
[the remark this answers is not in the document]
DrDeano: ;)

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+const Allocator = std.mem.Allocator;
+const TailQueue = std.TailQueue;
+
+/// Pointer to the start of the main kernel stack
+extern var KERNEL_STACK_START: []u32;
+
+/// The default stack size of a task. Currently this is 4KB.
+const STACK_SIZE: u32 = arch.MEMORY_BLOCK_SIZE / @sizeOf(u32);
+
+/// The stack size in bytes as the stack is allocated as u32, which is 4 bytes.
```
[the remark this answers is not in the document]
DrDeano: This wan't used anyway, so have removed

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+    /// The current stack pointer into the stack.
+    esp: u32,
+
+    ///
+    /// Initialise a task. This will allocated a PID and the stack. The stack will be set up as a
+    /// kernel task. As this is a new task, the stack will need to be initialised with the CPU
+    /// state as described in arch.CpuState struct.
+    ///
+    /// Arguments:
+    ///     IN comptime entry_point: var - The entry point into the task. This must be a function.
```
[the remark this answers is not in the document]
DrDeano: But this can also be `fn () noreturn`. I was looking into this, but this is how the std.thread does it. But for now can be `fn () void`

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,583 @@
...
+        var task = try alloc.create(Task);
+        errdefer alloc.destroy(task);
+
+        task.pid = allocatePid();
+        errdefer freePid(task.pid);
+
+        // Will need to copy the cpu state onto the stack from eip onwards
+        task.stack = try alloc.alloc(u32, STACK_SIZE);
+
+        // TODO Will need to add the exit point
```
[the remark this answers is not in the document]
DrDeano: The task struct is being moved to the arch folder, so this can stay :P

src/kernel/scheduler.zig
DrDeano: Task struct is being moved to arch file

src/kernel/scheduler.zig
DrDeano: I guess so yep, I am just reluctant in using unreachable as this is removed in releasefast builds and would lead to undefined behaviour

src/kernel/scheduler.zig
DrDeano: Not really sure what you mean

src/kernel/scheduler.zig
DrDeano: So will `pickNextTask` then return a Task and then in the schedule function: `current_task = pickNextTask();`

src/kernel/scheduler.zig
DrDeano: Sure, so the RTC will call pickNextTask instead

src/kernel/scheduler.zig
DrDeano: I might move some stuff around

src/kernel/arch/x86/paging.zig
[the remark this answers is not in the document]
SamTebbs33: Oh I meant the changes to the page fault function rather than the return type change! The return type of course needs to change in this PR. I just think changing what the pageFault function actually does would be nice to have in a different PR.

src/kernel/scheduler.zig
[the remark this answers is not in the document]
SamTebbs33: Ah that is a good point. I'm OK with what you'd prefer.

src/kernel/scheduler.zig
SamTebbs33: That's my thought

src/kernel/scheduler.zig
SamTebbs33: Yeah it would call pickNextTask to get the next task to run and then switchTask to do the actual switch.

src/kernel/arch/x86/rtc.zig
SamTebbs33: Oh I see. So delaying returning from the PIT handler could cause time inaccuracies?

src/kernel/scheduler.zig
SamTebbs33: True! I'm OK with whatever you choose.

src/kernel/arch/x86/paging.zig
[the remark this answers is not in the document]
DrDeano: Ah, ok, can do :)

src/kernel/scheduler.zig
DrDeano: Ok, it make sense now :) I was thinking just thinking that if pickNext would return the task to switch to, then at some point it will set the current_task, so why not just set it in the pickNext to reduce that step. All the switchTask function will do will return the stack pointer of the task as the interrupt handler will swap the task

src/kernel/scheduler.zig
DrDeano: I have changed this so no allocations are needed

src/kernel/arch/x86/arch.zig
```diff
@@ -319,6 +329,14 @@ pub fn initMem(mb_info: BootPayload) std.mem.Allocator.Error!MemProfile {
...
     defer log.logInfo("Done mem\n", .{});
 
+    log.logDebug("KERNEL_ADDR_OFFSET:    0x{X}\n", .{@ptrToInt(&KERNEL_ADDR_OFFSET)});
+    log.logDebug("KERNEL_STACK_START:    0x{X}\n", .{@ptrToInt(&KERNEL_STACK_START)});
+    log.logDebug("KERNEL_STACK_END:      0x{X}\n", .{@ptrToInt(&KERNEL_STACK_END)});
+    log.logDebug("KERNEL_VADDR_START:    0x{X}\n", .{@ptrToInt(&KERNEL_VADDR_START)});
+    log.logDebug("KERNEL_VADDR_END:      0x{X}\n", .{@ptrToInt(&KERNEL_VADDR_END)});
+    log.logDebug("KERNEL_PHYSADDR_START: 0x{X}\n", .{@ptrToInt(&KERNEL_PHYSADDR_START)});
+    log.logDebug("KERNEL_PHYSADDR_END:   0x{X}\n", .{@ptrToInt(&KERNEL_PHYSADDR_END)});
+
```
SamTebbs33, line 339: Hopefully debugging this is over so let's purge these from existence

src/kernel/arch/x86/arch.zig
```diff
@@ -388,49 +407,60 @@ pub fn initMem(mb_info: BootPayload) std.mem.Allocator.Error!MemProfile {
...
+    // _ = @import("isr.zig");
+    // _ = @import("paging.zig");
+    // _ = @import("pic.zig");
+    // _ = @import("pit.zig");
+    // _ = @import("rtc.zig");
+    // _ = @import("syscalls.zig");
+    // _ = @import("task.zig");
+    // _ = @import("tty.zig");
+    // _ = @import("vga.zig");
+    std.meta.refAllDecls(@This());
```
SamTebbs33, line 465: Keep or delete these?

src/kernel/arch/x86/interrupts.zig
```diff
@@ -40,7 +41,8 @@ export fn commonStub() callconv(.Naked) void {
         \\mov   %%esp, %%eax
         \\push  %%eax
         \\call  handler
-        \\pop   %%eax
+        \\mov   %%eax, %%esp
```
SamTebbs33, line 44: After my experiences with how the release modes can change the ABI, I was wondering if it would be better to delegate the `push eax; call handler; mov eax, esp` to Zig, with something like.
```
const esp = handler(asm ("mov %[stack], %%esp" : [stack] "=r" (-> usize));
asm volatile ("mov %%esp, %[stack]" :: [stack] "r" (esp));
```
That way, this code will work even when a different ABI is being used, due to optimisations.

src/kernel/arch/x86/paging.zig
```diff
@@ -316,6 +318,7 @@ pub fn map(virt_start: usize, virt_end: usize, phys_start: usize, phys_end: usiz
         virt_addr += PAGE_SIZE_4MB;
         entry_idx += 1;
     }) {
+        //log.logInfo("Mapping: virt_start={}, virt_end={}, phys_start={}, phys_end={}\n", .{virt_start, virt_end, phys_start, phys_end});
```
SamTebbs33, line 321: PURGE

src/kernel/arch/x86/task.zig
```diff
@@ -0,0 +1,265 @@
...
+
+/// The kernels main stack start as this is used to check for if the task being destroyed is this stack
+/// as we cannot deallocate this.
+extern var KERNEL_STACK_START: *u32;
+
+/// The function type for the entry point.
+const EntryPointFn = fn () void;
+
+/// The bitmap type for the PIDs
+const PidBitmap = if (is_test) ComptimeBitmap(u128) else ComptimeBitmap(u1024);
```
SamTebbs33, line 24: It would be good to move the PID stuff to scheduler.zig since all architectures will use them.

src/kernel/kmain.zig
```diff
@@ -28,7 +29,14 @@ var kernel_vmm: vmm.VirtualMemoryManager(arch.VmmPayload) = undefined;
...
 // This is for unit testing as we need to export KERNEL_ADDR_OFFSET as it is no longer available
 // from the linker script
+// These will need to be kept up to date with the debug logs in the mem init.
 export var KERNEL_ADDR_OFFSET: u32 = if (builtin.is_test) 0xC0000000 else undefined;
+export var KERNEL_STACK_START: u32 = if (builtin.is_test) 0xc013c000 else undefined;
+export var KERNEL_STACK_END: u32 = if (builtin.is_test) 0xc0140000 else undefined;
+export var KERNEL_VADDR_START: u32 = if (builtin.is_test) 0xC0100000 else undefined;
+export var KERNEL_VADDR_END: u32 = if (builtin.is_test) 0xc0140000 else undefined;
+export var KERNEL_PHYSADDR_START: u32 = if (builtin.is_test) 0x100000 else undefined;
+export var KERNEL_PHYSADDR_END: u32 = if (builtin.is_test) 0x140000 else undefined;
```
SamTebbs33, line 39: How come we need these here now?

src/kernel/kmain.zig
```diff
@@ -64,10 +72,34 @@ export fn kmain(boot_payload: arch.BootPayload) void {
...
     tty.init(&kernel_heap.allocator, boot_payload);
 
+    scheduler.init(&kernel_heap.allocator) catch |e| {
+        panic_root.panic(@errorReturnTrace(), "Failed to initialise scheduler: {}", .{e});
+    };
+
+    // Initialisation is finished, now does other stuff
     log.logInfo("Init done\n", .{});
 
+    // Main initialisation finished that relay on no interrupts so can enable interrupts
```
SamTebbs33, line 85: You might want to reword this sentence

src/kernel/scheduler.zig
```diff
@@ -0,0 +1,310 @@
...
+/// Create a new task and add it to the scheduling queue. No locking.
+///
+/// Arguments:
+///     IN entry_point: EntryPointFn - The entry point into the task. This must be a function.
+///
+/// Error: Allocator.Error
+///     OutOfMemory - If there isn't enough memory for the a task/stack. Any memory allocated will
+///                   be freed on return.
+///
+pub fn createNewTask(entry_point: EntryPointFn) Allocator.Error!void {
```
SamTebbs33, line 93: Could this take a `Task` instead and add it to the queue? I think that would make for a cleaner and more flexible API. A name change could follow too (`queueTask`? `scheduleTask`?).

src/kernel/arch/x86/task.zig
```diff
@@ -0,0 +1,265 @@
...
+    ///     IN allocator: *Allocator     - The allocator for allocating memory for a task.
+    ///
+    /// Return: *Task
+    ///     Pointer to an allocated task. This will then need to be added to the task queue.
+    ///
+    /// Error: Allocator.Error
+    ///     OutOfMemory - If there is no more memory to allocate. Any memory or PID allocated will
+    ///                   be freed on return.
+    ///
+    pub fn init(entry_point: EntryPointFn, allocator: *Allocator) Allocator.Error!*Task {
```
SamTebbs33, line 66: Could this return a `Task` rather than a `*Task`? Then it's up to the caller to decide if it should go on the stack or heap. Makes it a bit more flexible.

SamTebbs33:
I meant to request changes instead of approve!

src/kernel/arch/x86/arch.zig
DrDeano: I was thinking that this will be used to keep the exports in kmain up to date

src/kernel/arch/x86/arch.zig
DrDeano: ah yes, removed

src/kernel/arch/x86/paging.zig
DrDeano: Tried to find them all, but I failed

src/kernel/kmain.zig
DrDeano: If they are needed for unit testing, then don't need to add them. Plus some of them are needed

src/kernel/arch/x86/task.zig
DrDeano: I might have an idea for moving the task struct out of arch (along with the PID stuff)

src/kernel/scheduler.zig
DrDeano: Ah yes, I think there was a comment before, I will rework :)

src/kernel/arch/x86/task.zig
DrDeano: But when would you want to put a task on the stack as most likely the function called that creates a task will go out of scope so will be safer if it is in allocated memory

src/kernel/kmain.zig
SamTebbs33: Ok, we could keep them here for now, but I think we should clean this sort of thing up in another PR since they're used all over the place. Perhaps we could move them to mem.zig and abstract them away with functions.

src/kernel/arch/x86/task.zig
SamTebbs33: True. I guess we could change it if the necessity arises.

src/kernel/arch/x86/arch.zig
SamTebbs33: Keep them up to date?

src/kernel/arch/x86/interrupts.zig
DrDeano: I have had ago at this, but always get a page fault. When I go back to the old (original) version, it works. Could you see if you are able to get this working.

SamTebbs33: Sure, I'll take a look at the assembly too

SamTebbs33: Yeah doing this adds a load of gunk in debug mode that screws with the stack. Let's keep it like it is at the moment but be aware that we may need to pay attention to the ABI (passing the argument and receiving the return value differently) if we run into problems later.

SamTebbs33: It adds
```
mov    %esp,%esi
c011bd86:       8b 7e 34                mov    0x34(%esi),%edi
c011bd89:       83 ff 20                cmp    $0x20,%edi
c011bd8c:       72 20                   jb     c011bdae <commonStub+0x3e>
c011bd8e:       81 ff 80 00 00 00       cmp    $0x80,%edi
c011bd94:       74 18                   je     c011bdae <commonStub+0x3e>
c011bd96:       56                      push   %esi
c011bd97:       e8 54 f2 ff ff          call   c011aff0 <irqHandler>
c011bd9c:       83 c4 04                add    $0x4,%esp
c011bd9f:       89 c4                   mov    %eax,%esp
```
in release-fast and release-small, and

```
c013af74:       89 e0                   mov    %esp,%eax
c013af76:       89 45 fc                mov    %eax,-0x4(%ebp)
c013af79:       8b 45 fc                mov    -0x4(%ebp),%eax
c013af7c:       83 f8 00                cmp    $0x0,%eax
c013af7f:       89 45 f4                mov    %eax,-0xc(%ebp)
c013af82:       75 0d                   jne    c013af91 <commonStub+0x31>
c013af84:       31 d2                   xor    %edx,%edx
c013af86:       8d 0d f4 e5 13 c0       lea    0xc013e5f4,%ecx
c013af8c:       e8 6f 80 fc ff          call   c0103000 <panic>
c013af91:       8b 45 f4                mov    -0xc(%ebp),%eax
c013af94:       83 e0 00                and    $0x0,%eax
c013af97:       83 f8 00                cmp    $0x0,%eax
c013af9a:       74 0d                   je     c013afa9 <commonStub+0x49>
c013af9c:       31 d2                   xor    %edx,%edx
c013af9e:       8d 0d fc e5 13 c0       lea    0xc013e5fc,%ecx
c013afa4:       e8 57 80 fc ff          call   c0103000 <panic>
c013afa9:       8b 45 f4                mov    -0xc(%ebp),%eax
c013afac:       89 04 24                mov    %eax,(%esp)
c013afaf:       e8 4c ff ff ff          call   c013af00 <handler>
c013afb4:       89 45 f8                mov    %eax,-0x8(%ebp)
c013afb7:       8b 45 f8                mov    -0x8(%ebp),%eax
```

in debug mode thanks to the safety checks and lack of optimisations

DrDeano: yeh, I was expecting this to happen. Will just have to keep an eye on it

src/kernel/arch/x86/arch.zig
DrDeano: As we add code to the kernel, these values will change so can update the values in kmain that export these values for the unit testing so can keep things correct

src/kernel/scheduler.zig
SamTebbs33: Nice, then it could be used to enqueue the old task when switching.

DrDeano: I manipulate the queue pointers instead so to reduce the allocations and make the switching faster

SamTebbs33: That's a nice optimisation. Forget the second bit then.

src/kernel/arch/x86/arch.zig
```diff
@@ -403,6 +402,38 @@ pub fn initMem(mb_info: BootPayload) std.mem.Allocator.Error!MemProfile {
     };
 }
 
+pub fn initTaskStack(entry_point: usize, allocator: *Allocator) Allocator.Error!struct { stack: []u32, pointer: usize } {
```
SamTebbs33, line 405: Doc comment ;)

src/kernel/task.zig
```diff
@@ -178,43 +145,43 @@ test "init out of memory for stack" {
     expectEqual(all_pids.bitmap, 1);
 }
 
-test "init expected setup" {
-    var task = try Task.init(test_fn1, std.testing.allocator);
+test "create expected setup" {
```
SamTebbs33, line 148: Hmm I think this should be moved back to arch.zig since it depends on the implementation there.

src/kernel/arch/x86/arch.zig
DrDeano: lel yep

src/kernel/task.zig
```diff
@@ -152,28 +152,33 @@ test "create expected setup" {
...
-    expectEqual(task.stack[arch.STACK_SIZE - 7], 0);
-    expectEqual(task.stack[arch.STACK_SIZE - 6], 0);
-    expectEqual(task.stack[arch.STACK_SIZE - 5], 0);
-    expectEqual(task.stack[arch.STACK_SIZE - 4], 0);
-    expectEqual(task.stack[arch.STACK_SIZE - 3], @ptrToInt(test_fn1));
-    expectEqual(task.stack[arch.STACK_SIZE - 2], 0x08);
-    expectEqual(task.stack[arch.STACK_SIZE - 1], 0x202);
-
-    expectEqual(task.stack_pointer, @ptrToInt(&task.stack[arch.STACK_SIZE - 18]));
+    switch (builtin.arch) {
```
SamTebbs33, line 155: Oh I meant put this in arch.zig, where `initTaskStack` is defined.

DrDeano: In that case, it isn't really testing anything as it is all hard coded values and no complicated set up, so could just remove this test

SamTebbs33: That sounds good to me

PR #140, opened 2020-05-06
Log CPU state on page fault

DrDeano:
When a page fault occurs, will log the CPU state and the control registers.

### DIFF
--- a/src/kernel/arch/x86/paging.zig
+++ b/src/kernel/arch/x86/paging.zig
@@ -358,12 +358,30 @@ pub fn unmap(virtual_start: usize, virtual_end: usize, dir: *Directory) (std.mem
 }
 
 ///
-/// Called when a page fault occurs.
+/// Called when a page fault occurs. This will log the CPU state and control registers.
 ///
 /// Arguments:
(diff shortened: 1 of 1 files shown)


src/kernel/arch/x86/arch.zig
```diff
@@ -210,6 +210,54 @@ pub fn halt() void {
...
+/// Return the value of control register 4.
+///
+/// Return: u32
+///     The value in CR4.
+///
+pub fn CR4() u32 {
+    return asm volatile ("mov %%cr4, %[cr4]"
+        : [cr4] "=r" (-> u32)
+    );
+}
```
SamTebbs33, line 259: Would you agree on moving the inline assembly from these functions into the handler itself? I think having them in here bloats the file a bit.

DrDeano: But what if other files need to use the control registers

SamTebbs33: If that happens we could abstract them out. These are so simple I think it's unnecessary to have them as functions, at the moment at least :)

PR #141, opened 2020-05-06
Changed runtime tests to just look for FAILURE in the logging

DrDeano:
This allows for easier runtime testing in the face of the scheduler as the logging may not be in the order expected


src/kernel/arch/x86/syscalls.zig
```diff
@@ -286,29 +287,38 @@ test "registerSyscall returns SyscallExists" {
...
 
-    if (syscall4(127, 2, 3, 4, 5) == 4 and testInt == 31)
-        log.logInfo("Syscalls: Tested 4 args\n", .{});
+    if (syscall4(127, 2, 3, 4, 5) != 4 and testInt != 31) {
+        panic(@errorReturnTrace(), "FAILURE\n", .{});
+    }
 
-    if (syscall5(128, 2, 3, 4, 5, 6) == 5 and testInt == 51)
-        log.logInfo("Syscalls: Tested 5 args\n", .{});
+    if (syscall5(128, 2, 3, 4, 5, 6) != 5 and testInt != 51) {
```
SamTebbs33, line 321: The `and` in the if statement conditions should be `or` since the test has failed if either value is incorrect.

src/kernel/vmm.zig
```diff
@@ -309,12 +309,12 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
                 for (physical.items) |block, i| {
                     // Clear the address space entry, unmap the virtual memory and free the physical memory
                     try self.bmp.clearEntry(entry + i);
-                    pmm.free(block) catch |e| panic(@errorReturnTrace(), "Failed to free PMM reserved memory at {x}: {}\n", .{ block * BLOCK_SIZE, e });
+                    pmm.free(block) catch |e| panic(@errorReturnTrace(), "Failed to free PMM reserved memory at 0x{X}: {}\n", .{ block * BLOCK_SIZE, e });
```
SamTebbs33, line 312: Thanks for noticing the missing 0x, we should be consistent.

DrDeano:
Kinda feels bad that I basically deleted everything lel.

test/rt-test.py
```diff
@@ -79,42 +26,33 @@ def cleanup():
...
+            if (finished_test):
+                print("\nTesting Successful")
+                break
+            failure("Timed out")
+
+        if "FAILURE" in line:
+            failure(line)
+        elif "[ERROR] Kernel panic:" in line:
+            # As the panic test will be last, then the timeout can elapse
+            finished_test = True
```
SamTebbs33, line 53: What if we get a panic that isn't the one we expected? Is that possible? I also noticed that you have to wait 5 seconds (the timeout length) after the tests have been run until the program exits. Can you think of a way to make the program exit as soon as all tests have been run and were successful? If not that's fine, we can get back to it if it gets annoying.

DrDeano: I will have another think :)

SamTebbs33:
> Kinda feels bad that I basically deleted everything lel.

That's what iterative improvement does :D

build.zig
```diff
@@ -118,3 +122,44 @@ pub fn build(b: *Builder) !void {
...
+                    break;
+                }
+            }
+
+            b.allocator.free(buffer);
+            file.close();
+        }
+    }
+    return count;
+}
```
SamTebbs33, line 165: This is an interesting idea but I think it's a bit unnecessary. This will fall down if we have two logTest calls but have them conditionally executed so only one will run. Why not have a list in the python file of expected test log regexes and remove them as you find TESTING log messages that match? Then once the list is empty you can exit the python file knowing everything has passed.

DrDeano: But would the list of of regex also have this same problem of if there is a log in a condition

DrDeano: Have change to remove cases from a list

test/rt-test.py
```diff
@@ -90,33 +76,38 @@ def cleanup():
...
+        line = line.strip()
+
+        # Print the line so can see what is going on
+        print(line)
+
+        # If there is a FAILURE message in the log, then fail the testing
+        if "FAILURE" in line:
+            failure("Test failed")
+        
+        if not cases:
```
SamTebbs33, line 104: Why is this check necessary? Doesn't the loop run until the list is empty? I may be misinterpreting the python here :)

test/rt-test.py
```diff
@@ -90,33 +76,38 @@ def cleanup():
...
+            # Empty cases, so expecting something, fail
+            failure("Empty cases")
+
+        # Remove the line from the cases, this is slow for now
+        cases = [item for item in cases if item not in line]
```
SamTebbs33, line 109: Say the list was ["abcde", "abc"] and `line` was "abc", would this remove "abcde"?

src/kernel/panic.zig
```diff
@@ -90,6 +90,57 @@ const SymbolMap = struct {
...
+            while (address != first_address) {
+                address = self.next_internal() orelse return null;
+            }
+            self.first_address = null;
+        }
+
+        return address;
+    }
+
+    fn next_internal(self: *StackIterator) ?usize {
```
SamTebbs33, line 123: The fact that there's a `next_internal` functions must mean that `next` is supposed to be public, but we can use this until the std lib is fixed (if it is going to be) 👍

DrDeano: Sure, also still working on the python stuff :)

test/rt-test.py
DrDeano: Yeh, I can remove this

test/rt-test.py
DrDeano: It would remove both, either we can make sure that the cases are unique, or add additional information in the logs e.g. the process it is coming from. I could write a checker in the python that makes sure that none of the cases are substrings of other cases

SamTebbs33: Wouldn't checking for equality between `item` and `line` work though? That would avoid removing the wrong one.

build.zig
```diff
@@ -27,16 +27,28 @@ pub fn build(b: *Builder) !void {
     });
     b.default_step.dependOn(&fmt_step.step);
 
+    const test_enum = enum {
+        UNIT,
+        ALL_RUNTIME,
+        NORMAL,
+        PANIC,
+    };
```
SamTebbs33, line 35: I think these names could be more clear and perhaps add a comment above each describing what they mean.

build.zig
```diff
@@ -27,16 +27,28 @@ pub fn build(b: *Builder) !void {
...
+
+    comptime var available_tests: []const u8 = "";
+    inline for (@typeInfo(test_enum).Enum.fields) |field| {
+        available_tests = available_tests ++ field.name ++ ", ";
```
SamTebbs33, line 39: You could add a description of what each test mode does so the user knows what to choose. You could do so by making a function in `test_enum` that returns the description and adding it to the string (perhaps within parentheses) here.

build.zig
```diff
@@ -27,16 +27,28 @@ pub fn build(b: *Builder) !void {
...
+    }
+
     const main_src = "src/kernel/kmain.zig";
     const constants_path = try fs.path.join(b.allocator, &[_][]const u8{ "src/kernel/arch", target_str, "constants.zig" });
 
     const build_mode = b.standardReleaseOptions();
-    const rt_test = b.option(bool, "rt-test", "enable/disable runtime testing") orelse false;
+    const test_type = b.option(test_enum, "test-type", "Run a specific runtime test. Available options are: " ++ available_tests) orelse test_enum.UNIT;
```
SamTebbs33, line 46: Doesn't the test mode also determine whether or not to do unit testing not just a specific type of runtime test?

build.zig
```diff
@@ -27,16 +27,28 @@ pub fn build(b: *Builder) !void {
...
 
     const exec = b.addExecutable("pluto", main_src);
     exec.addPackagePath("constants", constants_path);
     exec.setOutputDir(b.cache_root);
-    exec.addBuildOption(bool, "rt_test", rt_test);
+    exec.addBuildOption(test_enum, "test_type", test_type); // Or test_mode?
```
SamTebbs33, line 51: I think `test_mode` is better :+1:

build.zig
```diff
@@ -27,16 +27,28 @@ pub fn build(b: *Builder) !void {
     });
     b.default_step.dependOn(&fmt_step.step);
 
+    const test_enum = enum {
```
SamTebbs33, line 30: Maybe `TestMode`?

src/kernel/arch/x86/gdt.zig
```diff
@@ -438,7 +438,10 @@ pub fn init() void {
     // Load the TSS
     arch.ltr(TSS_OFFSET);
 
-    if (build_options.rt_test) runtimeTests();
+    switch (build_options.test_type) {
+        .NORMAL => runtimeTests(),
```
SamTebbs33, line 442: What if it's `ALL_RUNTIME`?

src/kernel/arch/x86/syscalls.zig
```diff
@@ -286,29 +291,40 @@ test "registerSyscall returns SyscallExists" {
...
+    }
 
-    if (syscall4(127, 2, 3, 4, 5) == 4 and testInt == 31)
-        log.logInfo("Syscalls: Tested 4 args\n", .{});
+    if (syscall4(127, 2, 3, 4, 5) != 4 or testInt != 31) {
+        panic(@errorReturnTrace(), "FAILURE\n", .{});
+    }
+
+    if (syscall5(128, 2, 3, 4, 5, 6) != 5 or testInt != 51) {
+        panic(@errorReturnTrace(), "FAILURE\n", .{});
```
SamTebbs33, line 326: These failures could do with a message so we know what has failed.

src/kernel/panic.zig
```diff
@@ -281,10 +284,11 @@ pub fn init(mem_profile: *const mem.MemProfile, allocator: *std.mem.Allocator) !
...
 
     // Exit if we haven't loaded all debug modules
-    if (mem_profile.boot_modules.len < 1)
+    if (mem_profile.boot_modules.len < 1) {
         return;
-    var kmap_start: usize = 0;
-    var kmap_end: usize = 0;
+    }
+    var kmap_start: u32 = 0;
+    var kmap_end: u32 = 0;
```
SamTebbs33, line 291: Why change it back to `u32`? Rebase error maybe?

src/kernel/vmm.zig
```diff
@@ -581,41 +584,41 @@ fn runtimeTests(comptime Payload: type, vmm: VirtualMemoryManager(Payload), mem_
...
-        if (!set) panic(null, "Address for kernel code was not set: {x}", .{vaddr});
+        const set = vmm.isSet(vaddr) catch |e| panic(@errorReturnTrace(), "FAILURE: Failed to check if kernel code address 0x{X} is set: 0x{X}\n", .{ vaddr, e });
+        if (!set) panic(null, "FAILURE: Address for kernel code was not set: 0x{X}\n", .{vaddr});
     }
 
     // Make sure all blocks after the kernel code are not set
     while (vaddr < vmm.end - BLOCK_SIZE) : (vaddr += BLOCK_SIZE) {
-        const set = vmm.isSet(vaddr) catch |e| panic(@errorReturnTrace(), "Failed to check if address after {x} is set: {x}", .{ vaddr, e });
-        if (set) panic(null, "Address after kernel code was set: {x}", .{vaddr});
+        const set = vmm.isSet(vaddr) catch |e| panic(@errorReturnTrace(), "Failed to check if address after 0x{X} is set: 0x{X}", .{ vaddr, e });
```
SamTebbs33, line 620: The second set of braces in each message should actually be `{}` since they correspond to the error not a number.

build.zig
DrDeano: Yep :)

build.zig
DrDeano: A function sounds good

build.zig
DrDeano: yes, the `test_type` soon to be `test_mode` will also determin if it is to do the unit test, as the default test is `.UNIT`

build.zig
DrDeano: Ah yes it is a type lel

src/kernel/arch/x86/gdt.zig
DrDeano: The python script will run each test individually, `build_options.test_type` should never be `ALL_RUNTIME` when it gets to the OS code. Maybe could add a check for this. `ALL_RUNTIME` is really only to tell the python runtime to run all the runtime tests.

src/kernel/panic.zig
DrDeano: ah, maybe, it didn't come up for me when rebasing

build.zig
SamTebbs33: Cool, I suggest you change the help message then :)

src/kernel/arch/x86/gdt.zig
SamTebbs33: Hmm ok, if it's never supposed to be used by the Zig code I'd prefer if it wasn't part of the enum at all.

DrDeano: hmm, idk how to remove it without adding another build option for just running all the runtime tests

SamTebbs33: How about using something similar the previous system where `zig build test` runs the unit tests but adding `-Drt-test=x` will build with runtime testing enabled and can be `normal`, `panic` or `multitasking`? That has the advantage that you can run the kernel with rt-testing enabled and see all the log messages without actually testing. If you want to test all runtime tests you could `&&` each command together in the shell.

build.zig
```diff
@@ -8,6 +8,40 @@ const CrossTarget = std.zig.CrossTarget;
...
+    ///
+    /// Return: []const u8
+    ///     The string description for the test mode.
+    ///
+    pub fn getDescription(mode: TestMode) []const u8 {
+        return switch (mode) {
+            .UNIT => "Run the unit tests",
+            .ALL_RUNTIME => "Run all the runtime tests",
+            .INITIALISATION => "Ensure the OS is initialised correctly: Runtime test",
+            .PANIC => "Ensure panic behaviour correctly: Runtime test",
```
SamTebbs33, line 40: I think having "Run" here again is unnecessary since it's already in the option's help text and the colon part could be removed if you reword them to "Initialisation runtime tests" and "Panic runtime tests" or something like that

build.zig
```diff
@@ -27,28 +61,27 @@ pub fn build(b: *Builder) !void {
...
         available_tests = available_tests ++ field.name ++ ", ";
     }
 
     const main_src = "src/kernel/kmain.zig";
     const constants_path = try fs.path.join(b.allocator, &[_][]const u8{ "src/kernel/arch", target_str, "constants.zig" });
 
     const build_mode = b.standardReleaseOptions();
-    const test_type = b.option(test_enum, "test-type", "Run a specific runtime test. Available options are: " ++ available_tests) orelse test_enum.UNIT;
+    comptime var test_mode_desc: []const u8 = "\n                         ";
+    inline for (@typeInfo(TestMode).Enum.fields) |field| {
```
SamTebbs33, line 74: Very nice :)

src/kernel/arch/x86/gdt.zig
SamTebbs33: How about we remove the ALL option and if you want to run all runtime tests you just `&&` the shell commands together?

DrDeano: Could do, but the command could get very long if we have lots of tests, also will allow us to easily paralise the the tests if we let the build script do it for us. I don't mind having it in the enum, but would have to add an additional option for running all the runtimes.

build.zig
```diff
@@ -47,57 +63,82 @@ pub fn build(b: *Builder) !void {
...
+        .Debug => "",
+        .ReleaseSafe => "-Drelease-safe",
+        .ReleaseFast => "-Drelease-fast",
+        .ReleaseSmall => "-Drelease-small",
+    };
+    //const script = b.addSystemCommand(&[_][]const u8{ "python3", "test/rt-test.py", b.zig_exe, target_str, build_mode_str,  @tagName(test_mode) });
+    //rt_test_step.dependOn(&script.step);
+    //_ = try rt.run(b.allocator, b.zig_exe, target_str, build_mode_str, @tagName(test_mode));
+
+    var qemu_args = ArrayList([]const u8).init(b.allocator);
```
SamTebbs33, line 99: Why an array list and not an array literal like before?

DrDeano:
I am closing this pull request in favor for another with replaced the python runtime testing with zig.

PR #149, opened 2020-05-15
Abstract away boot modules and memory maps

### LINKED ISSUE #147: Remove dependence on multiboot for all architectures
Currently the kernel depends on multiboot information being present at boot time. This works well for generic x86 where grub is packaged with the boot image but won't work for boards like the Raspberry Pi that have bootloaders in firmware that don't comply with the multiboot standard.

To solve this we should delegate `MemProfile` initialisation to the architecture and allow it to specify the type of the data passed to kmain.

SamTebbs33:
Closes #147 
This patch removes the dependence on multiboot since not all architectures will use a multiboot-compliant bootloader (the Raspberry Pi for example). This meant allowing the architecture to decide what regions of memory are reserved and which are available for use by the VMM, which had the consequence of making the VMM's allocation of reserved memory cleaner.


src/kernel/arch/x86/arch.zig
```diff
@@ -231,6 +250,80 @@ pub fn haltNoInterrupts() noreturn {
...
+
+    var allocator = std.heap.FixedBufferAllocator.init(vaddr_end[0..mem.FIXED_ALLOC_SIZE]);
+    var reserved_physical_mem = std.ArrayList(mem.Range).init(&allocator.allocator);
+    var reserved_virtual_mem = std.ArrayList(mem.Map).init(&allocator.allocator);
+    const mem_map = @intToPtr([*]multiboot.multiboot_memory_map_t, mmap_addr)[0..num_mmap_entries];
+
+    // Reserve the unavailable sections from the multiboot memory map
+    for (mem_map) |entry| {
+        if (entry.@"type" != multiboot.MULTIBOOT_MEMORY_AVAILABLE) {
+            const end: usize = if (entry.addr + entry.len > std.math.maxInt(usize)) std.math.maxInt(usize) else @intCast(usize, entry.addr + entry.len);
```
DrDeano, line 283: This if can cause a integer overflow, would be better: `entry.addr > std.math.maxInt(usize) - entry.len`. Also, what happens if the end is too long as this would lead to the OS using unavailable memory as it couldn't fit it in a uzise. Maybe best to error here.

src/kernel/arch/x86/arch.zig
```diff
@@ -231,6 +250,80 @@ pub fn haltNoInterrupts() noreturn {
...
+    const mb_physical = mem.Range{ .start = mem.virtToPhys(mb_region.start), .end = mem.virtToPhys(mb_region.end) };
+    try reserved_virtual_mem.append(.{ .virtual = mb_region, .physical = mb_physical });
+
+    // Map the tty buffer
+    const tty_addr = mem.virtToPhys(tty.getVideoBufferAddress());
+    const tty_region = mem.Range{
+        .start = tty_addr,
+        .end = tty_addr + 32 * 1024,
+    };
+    try reserved_virtual_mem.append(.{ .physical = tty_region, .virtual = .{ .start = mem.physToVirt(tty_region.start), .end = mem.physToVirt(tty_region.end) } });
```
DrDeano, line 301: Maybe do the same as for `mb_physical` for the tty memory using the `mem.Rage` to make it more clear. As anonymous structs are good, but can be confusing to read.

src/kernel/pmm.zig
```diff
@@ -105,20 +104,18 @@ pub fn init(mem: *const MemProfile, allocator: *std.mem.Allocator) void {
...
-                    PmmBitmap.BitmapError.OutOfBounds => break,
-                    else => panic(@errorReturnTrace(), "Failed setting address 0x{x} from memory map as occupied: {}", .{ addr, e }),
-                };
-            }
+    for (mem.physical_reserved) |entry| {
+        var addr = std.mem.alignBackward(entry.start, BLOCK_SIZE);
+        var end = entry.end - 1;
+        // If the end address can be aligned without overflowing then align it
+        if (end <= std.math.maxInt(usize) - BLOCK_SIZE)
+            end = std.mem.alignForward(end, BLOCK_SIZE);
```
DrDeano, line 112: `if () {
    blah;
}`
;)

src/kernel/vmm.zig
```diff
@@ -211,36 +208,42 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
+            if (virtual.start > virtual.end)
                 return VmmError.InvalidVirtAddresses;
 
-            virt = virtual_start;
-            while (virt < virtual_end) : (virt += BLOCK_SIZE) {
+            if (physical) |p| {
+                if (virtual.end - virtual.start != p.end - p.start)
+                    return VmmError.PhysicalVirtualMismatch;
+                if (p.start > p.end)
+                    return VmmError.InvalidPhysAddresses;
```
DrDeano, line 224: I like the `{}`

src/kernel/vmm.zig
```diff
@@ -211,36 +208,42 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
+                var phys = p.start;
+                while (phys < p.end) : (phys += BLOCK_SIZE) {
+                    if (try pmm.isSet(phys))
+                        return VmmError.PhysicalAlreadyAllocated;
```
DrDeano, line 228: `{}`

src/kernel/vmm.zig
```diff
@@ -517,7 +500,7 @@ fn testInit(num_entries: u32) std.mem.Allocator.Error!VirtualMemoryManager(u8) {
         }
     }
     var allocations = test_allocations orelse unreachable;
-    const mem_profile = mem.MemProfile{ .vaddr_end = undefined, .vaddr_start = undefined, .physaddr_start = undefined, .physaddr_end = undefined, .mem_kb = num_entries * BLOCK_SIZE / 1024, .fixed_alloc_size = undefined, .mem_map = &[_]multiboot.multiboot_memory_map_t{}, .boot_modules = &[_]multiboot.multiboot_module_t{} };
+    const mem_profile = mem.MemProfile{ .vaddr_end = undefined, .vaddr_start = undefined, .physaddr_start = undefined, .physaddr_end = undefined, .mem_kb = num_entries * BLOCK_SIZE / 1024, .fixed_allocator = undefined, .virtual_reserved = &[_]mem.Map{}, .physical_reserved = &[_]mem.Range{}, .modules = &[_]mem.Module{} };
```
DrDeano, line 503: Could newline for each comma

src/kernel/arch/x86/arch.zig
SamTebbs33: Nice idea on the if statement. Regarding the other point, what do you mean by the end being too long?

src/kernel/arch/x86/arch.zig
SamTebbs33: Sure thing. I'm only using an explicit `Range` above because you have to when assigning to a `?Range`.

src/kernel/arch/x86/arch.zig
DrDeano: if `entry.addr + entry.len > std.math.maxInt(usize)` this is true, then you just use `std.math.maxInt(usize)` so there there will be memory that won't be mapped out.

src/kernel/arch/x86/arch.zig
DrDeano: Or instead, could add new lines to make it clear where each parameter is as there are a lot of brackets and dots

src/kernel/arch/x86/arch.zig
SamTebbs33: I don't quite see how using `maxInt(usize)` is a problem. Since x86 is 32-bit it can't address the memory beyond `maxInt(usize)` anyway so ignoring whatever comes after that in memory map entry should be fine.

src/kernel/arch/x86/arch.zig
SamTebbs33: Sure!

src/kernel/vmm.zig
SamTebbs33: That would be much better

src/kernel/arch/x86/arch.zig
DrDeano: Ah yes, good point, maybe add a comment :)

SamTebbs33: Sure thing :)

SamTebbs33:
> Bam, don't forget to squash :P

Noice, now I can do some more work on the aarch64 port :D

PR #151, opened 2020-06-01
Remove start.asm and start.s

### LINKED ISSUE #150: start.asm and start.s are unused
We have files in src/kernel (start.s and start.asm) that we don't use so we should remove them.

SamTebbs33:
Closes #150﻿


SamTebbs33:
> Lel, Can't believe we didn't remove them earlier

I know! I only noticed them because I wanted to look at the codebase in a nice file tree.

PR #153, opened 2020-06-02
Move linker script

### LINKED ISSUE #152: Move linker script to arch folder
Architectures aren't always going to share linker scripts since the sections names, addresses or symbols could differ so the linker script should be moved to the arch folder.

SamTebbs33:
Closes #152﻿

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -31,7 +31,8 @@ pub fn build(b: *Builder) !void {
     b.default_step.dependOn(&fmt_step.step);
 
     const main_src = "src/kernel/kmain.zig";
-    const constants_path = try fs.path.join(b.allocator, &[_][]const u8{ "src/kernel/arch", arch, "constants.zig" });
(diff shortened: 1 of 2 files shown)


build.zig
```diff
@@ -38,7 +38,8 @@ pub fn build(b: *Builder) !void {
     exec.setOutputDir(b.cache_root);
     exec.addBuildOption(bool, "rt_test", rt_test);
     exec.setBuildMode(build_mode);
-    exec.setLinkerScriptPath("link.ld");
+    const linker_script_path = try fs.path.join(b.allocator, &[_][]const u8{ "src/kernel/arch", target_str, "link.ld" });
```
DrDeano, line 41: The string `"src/kernel/arch"` is used for the constants, so maybe can refactor this out

SamTebbs33: Nice idea

PR #155, opened 2020-06-02
Add architecture option to build.zig

### LINKED ISSUE #154: Add architecture option
We need to add an option to build.zig that will allow the user to select which architecture to build for. Currently only x86 is supported but more will be added as time goes on. This option should be used to add arch-specific options to qemu, gdb, the compiler etc.

SamTebbs33:
Adds a "-Darch" option to specify the architecture to build for, which has consequences on the qemu binary used and the image creation process. Closes #154﻿

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ zig build debug-run
 
 ## Debug
 
-Launch a gdb instance and connect to qemu.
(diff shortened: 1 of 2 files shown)


SamTebbs33:
> Bam

Noice. I forgot to add info to the readme so added a commit for that.

PR #157, opened 2020-06-04
Make serial dependent on architecture

### LINKED ISSUE #156: Make serial implementation dependent on the architecture
Serial communication is done differently on different architectures and machines so the way that serial communication is done should be delegated to the architecture.

A design I thought of is similar to the `Allocator` interface in that we have a `Serial` struct that contains write and read functions and is populated and initialised by the architecture.

SamTebbs33:
Closes #156

### DIFF
--- a/src/kernel/arch/x86/arch.zig
+++ b/src/kernel/arch/x86/arch.zig
@@ -8,6 +8,7 @@ const irq = @import("irq.zig");
 const isr = @import("isr.zig");
 const pit = @import("pit.zig");
 const rtc = @import("rtc.zig");
+const serial = @import("serial.zig");
 const paging = @import("paging.zig");
 const syscalls = @import("syscalls.zig");
 const mem = @import("../../mem.zig");
@@ -16,6 +17,8 @@ const pmm = @import("pmm.zig");
 const vmm = @import("../../vmm.zig");
 const log = @import("../../log.zig");
 const tty = @import("../../tty.zig");
+const Serial = @import("../../serial.zig").Serial;
+const panic = @import("../../panic.zig").panic;
 const MemProfile = mem.MemProfile;
 
 /// The virtual end of the kernel code
@@ -249,6 +252,31 @@ pub fn haltNoInterrupts() noreturn {
         disableInterrupts();
         halt();
     }
+}
+
+///
+/// Write a byte to serial port com1. Used by the serial initialiser
+///
+/// Arguments:
+///     IN byte: u8 - The byte to write
+///
+fn writeSerialCom1(byte: u8) void {
+    serial.write(byte, serial.Port.COM1);
+}
+
+///
+/// Initialise serial communication using port COM1 and construct a Serial instance
+///
+/// Return: serial.Serial
+///     The Serial instance constructed with the function used to write bytes
+///
+pub fn initSerial() Serial {
+    serial.init(serial.DEFAULT_BAUDRATE, serial.Port.COM1) catch |e| {
+        panic(@errorReturnTrace(), "Failed to initialise serial: {}", .{e});
+    };
+    return Serial{
+        .write = writeSerialCom1,
+    };
 }
 
 ///
--- a/src/kernel/kmain.zig
+++ b/src/kernel/kmain.zig
@@ -7,12 +7,11 @@ const arch = @import("arch.zig").internals;
 const tty = @import("tty.zig");
 const vga = @import("vga.zig");
 const log = @import("log.zig");
-const serial = @import("serial.zig");
 const pmm = @import("pmm.zig");
+const serial = @import("serial.zig");
 const vmm = if (is_test) @import(mock_path ++ "vmm_mock.zig") else @import("vmm.zig");
 const mem = if (is_test) @import(mock_path ++ "mem_mock.zig") else @import("mem.zig");
 const panic_root = if (is_test) @import(mock_path ++ "panic_mock.zig") else @import("panic.zig");
-const options = @import("build_options");
 const heap = @import("heap.zig");
 
 comptime {
@@ -38,11 +37,9 @@ pub fn panic(msg: []const u8, error_return_trace: ?*builtin.StackTrace) noreturn
 }
 
 export fn kmain(boot_payload: arch.BootPayload) void {
-    serial.init(serial.DEFAULT_BAUDRATE, serial.Port.COM1) catch |e| {
-        panic_root.panic(@errorReturnTrace(), "Failed to initialise serial: {}", .{e});
-    };
+    const serial_stream = serial.init();
 
-    if (build_options.rt_test) log.runtimeTests();
+    log.init(serial_stream);
 
     const mem_profile = arch.initMem(boot_payload) catch |e| panic_root.panic(@errorReturnTrace(), "Failed to initialise memory profile: {}", .{e});
     var fixed_allocator = mem_profile.fixed_allocator;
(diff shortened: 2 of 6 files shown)


test/mock/kernel/arch_mock.zig
```diff
@@ -6,6 +6,7 @@ const gdt = @import("gdt_mock.zig");
 const idt = @import("idt_mock.zig");
 const vmm = @import("vmm_mock.zig");
 const paging = @import("paging_mock.zig");
+const Serial = @import("../../../src/kernel/serial.zig").Serial;
```
DrDeano, line 9: Why not create a serial_mock.zig file?

src/kernel/arch/x86/arch.zig
```diff
@@ -251,6 +254,31 @@ pub fn haltNoInterrupts() noreturn {
...
+///
+pub fn initSerial() Serial {
+    x86_serial.init(x86_serial.DEFAULT_BAUDRATE, x86_serial.Port.COM1) catch |e| {
+        panic(@errorReturnTrace(), "Failed to initialise serial: {}", .{e});
+    };
+    return Serial{
+        .write = writeSerialCom1,
+    };
+}
+
```
DrDeano, line 281: Could move this to the Serial struct and have it as an `init()` function.

src/kernel/arch/x86/arch.zig
```diff
@@ -251,6 +254,31 @@ pub fn haltNoInterrupts() noreturn {
...
 
+///
+/// Write a byte to serial port com1. Used by the serial initialiser
+///
+/// Arguments:
+///     IN byte: u8 - The byte to write
+///
+fn writeSerialCom1(byte: u8) void {
+    x86_serial.write(byte, x86_serial.Port.COM1);
+}
```
DrDeano, line 265: Then this can be used for the Serial.init(), so be passed the function: `Serial.init(writeSerialCom1);` So this will be exposed as `pub`.

src/kernel/kmain.zig
```diff
@@ -38,11 +37,10 @@ pub fn panic(msg: []const u8, error_return_trace: ?*builtin.StackTrace) noreturn
 }
 
 export fn kmain(boot_payload: arch.BootPayload) void {
-    serial.init(serial.DEFAULT_BAUDRATE, serial.Port.COM1) catch |e| {
-        panic_root.panic(@errorReturnTrace(), "Failed to initialise serial: {}", .{e});
-    };
+    const serial_stream = arch.initSerial();
```
DrDeano, line 40: So here, would be `Serial.init(arch.writeSerialCom1)`.

src/kernel/serial.zig
```diff
@@ -1,217 +1,41 @@
...
 
 ///
 /// Run all the runtime tests
 ///
-fn runtimeTests() void {
-    rt_writeByte();
-    rt_writeBytes();
+pub fn runtimeTests(serial: Serial) void {
+    rt_writeByte(serial);
+    rt_writeBytes(serial);
```
DrDeano, line 25: If there is a `init()` in the Serial struct, then this can become private.

src/kernel/serial.zig
```diff
@@ -1,217 +1,41 @@
...
-                inline for ([_]u1{ 0, 1 }) |msb| {
-                    const val = try lcrValue(char_len, stop_bit, parity_bit, msb);
-                    const expected = char_len & 0x3 |
-                        @boolToInt(stop_bit) << 2 |
-                        @boolToInt(parity_bit) << 3 |
-                        @intCast(u8, msb) << 7;
-                    testing.expectEqual(val, expected);
-                }
-            }
+pub const Serial = struct {
```
DrDeano, line 1: Would need to import the x86 version of the serial via the arch. So would have to expose the serial in the arch file. Or could wait until the arch init and init the serial at the beginning of the arch.init() so wouldn't need to `pub` the serial in the arch file. But that would break logging for the memory managers. :( What do you think about this refactoring. Or could just keep it as is :P

test/mock/kernel/arch_mock.zig
SamTebbs33: The serial file is literally just the struct though so I didn't think that would be necessary xD

src/kernel/arch/x86/arch.zig
SamTebbs33: But the architecture is what does the initialisation

src/kernel/arch/x86/arch.zig
SamTebbs33: The thing is that the architecture is what decides how serial is done. Not all architectures will have the equivalent of a writeSerialCom1 function so it's best left to the arch to organise.

src/kernel/serial.zig
SamTebbs33: I think it's best as it is :) The way it is, the common code doesn't need to have any knowledge of how serial works under the hood and just lets the architecture do its thing, much like how allocators work.

SamTebbs33: Unless you can think of a better way!

src/kernel/arch/x86/arch.zig
```diff
@@ -8,6 +8,7 @@ const irq = @import("irq.zig");
 const isr = @import("isr.zig");
 const pit = @import("pit.zig");
 const rtc = @import("rtc.zig");
+const x86_serial = @import("serial.zig");
```
DrDeano, line 11: Does it need to be `x86_serial` as we are already in x86, so `serial` would be fine?

src/kernel/arch/x86/arch.zig
```diff
@@ -251,6 +254,31 @@ pub fn haltNoInterrupts() noreturn {
...
+/// Return: serial.Serial
+///     The Serial instance constructed with the function used to write bytes
+///
+pub fn initSerial() Serial {
+    x86_serial.init(x86_serial.DEFAULT_BAUDRATE, x86_serial.Port.COM1) catch |e| {
+        panic(@errorReturnTrace(), "Failed to initialise serial: {}", .{e});
+    };
+    return Serial{
+        .write = writeSerialCom1,
+    };
```
DrDeano, line 279: Maybe could move the runtime test from kmain to here to follow the convention in the init function.

src/kernel/kmain.zig
```diff
@@ -38,11 +37,10 @@ pub fn panic(msg: []const u8, error_return_trace: ?*builtin.StackTrace) noreturn
...
+    if (build_options.rt_test) serial.runtimeTests(serial_stream);
```
DrDeano, line 41: Move this to the `initSerial`

src/kernel/arch/x86/arch.zig
SamTebbs33: No it certainly doesn't!

src/kernel/arch/x86/arch.zig
SamTebbs33: Doing the serial tests isn't arch-specific so I thought it would be best to keep it in common code. How about I write an init function in the main serial.zig and have that initialise and run the tests?

DrDeano: Sure, just not a fan of calling the runtime tests from kmain, as it is part of the module not kmain. You get me :P

SamTebbs33: Yeah you're right actually. It's nice to keep things encapsulated :)

PR #175, opened 2020-06-24
Set the kernel VMM start address to the virtual address offset

### LINKED ISSUE #174: The kernel VMM starts at 4096 when it should start at the virtual address offset

SamTebbs33:
Closes #174

### DIFF
--- a/src/kernel/vmm.zig
+++ b/src/kernel/vmm.zig
@@ -108,6 +108,10 @@ pub const VmmError = error{
     InvalidPhysAddresses,
 };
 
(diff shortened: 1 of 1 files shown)


src/kernel/vmm.zig
```diff
@@ -108,6 +108,8 @@ pub const VmmError = error{
     InvalidPhysAddresses,
 };
 
+extern var KERNEL_ADDR_OFFSET: *u32;
```
DrDeano, line 111: Maybe can add a comment :P

PR #179, opened 2020-06-30
Insert the correct address into the allocations map

### LINKED ISSUE #178: VMM.set doesn't put the correct virtual address into the allocations list
vmm.set puts the last virtual address allocated into the allocations list, rather than the first. This can result in duplicate virtual address allocation.

SamTebbs33:
Closes #178﻿


SamTebbs33:
> :) Could add a test for allocating same addresses like in the other PR

I've added a check for the correct address being in the map since duplicate address allocation was a side effect of that not happening properly. Let me know if I should change the approach.

PR #181, opened 2020-07-06
Update for latest zig

SamTebbs33:


test/mock/kernel/mock_framework.zig
```diff
@@ -377,11 +393,11 @@ fn Mock() type {
         pub fn addAction(self: *Self, comptime fun_name: []const u8, data: var, action_type: ActionType) void {
             // Add a new mapping if one doesn't exist.
             if (!self.named_actions.contains(fun_name)) {
-                expect(self.named_actions.put(fun_name, TailQueue(Action).init()) catch unreachable == null);
```
SamTebbs33, line 380: I had to change this because `put` now returns void so there's nothing to compare with `null`.

PR #188, opened 2020-07-13
A comptime bitmap

DrDeano:
This allows for a bitmap that doesn't need an allocator.


DrDeano:
So we could init a bitmap as a global, I would like to use it for the PIDs as there is no runtime length of the bitmap and is hard coded, so if there was no allocations, then I can remove the tasking init function.
I shall see if I can reduce duplicate code.

PR #191, opened 2020-07-16
Latest zig - rename keyword var to anytype

markfirmware:


SamTebbs33:
This was actually done in c164f5ee4d5d71ba80f901e2c4b2211977d0a470, but I just forgot to rebase aarch64-port onto develop! I did miss out the change in vfs.zig though, so you're welcome to make the change there, submit a PR against develop and then we can rebase aarch64-port on top of that.

markfirmware:
I think you already have update vfs.zig.

PR #202, opened 2020-07-23
Feature/initrd

### LINKED ISSUE #201: Ramdisk
Create a initrd ramdisk that can be used to load drivers and other files that won't be part of the main kernel. This will be loaded as a grub module and will be using a custom file format for storing the files on the ramdisk. The files will only need to be read only so the file format can be very simple.
Making the ramdisk can be done using a build step in the build file.

DrDeano:
Closes #201

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -2,10 +2,13 @@ const std = @import("std");
 const builtin = @import("builtin");
 const rt = @import("test/runtime_test.zig");
 const RuntimeStep = rt.RuntimeStep;
+const Allocator = std.mem.Allocator;
 const Builder = std.build.Builder;
+const Step = std.build.Step;
 const Target = std.Target;
 const CrossTarget = std.zig.CrossTarget;
 const fs = std.fs;
+const File = fs.File;
 const Mode = builtin.Mode;
 const TestMode = rt.TestMode;
 const ArrayList = std.ArrayList;
@@ -38,6 +41,7 @@ pub fn build(b: *Builder) !void {
     const iso_dir_path = try fs.path.join(b.allocator, &[_][]const u8{ b.exe_dir, "iso" });
     const boot_path = try fs.path.join(b.allocator, &[_][]const u8{ b.exe_dir, "iso", "boot" });
     const modules_path = try fs.path.join(b.allocator, &[_][]const u8{ b.exe_dir, "iso", "modules" });
+    const ramdisk_path = try fs.path.join(b.allocator, &[_][]const u8{ b.install_path, "initrd.ramdisk" });
 
     const build_mode = b.standardReleaseOptions();
     comptime var test_mode_desc: []const u8 = "\n                         ";
@@ -59,10 +63,22 @@ pub fn build(b: *Builder) !void {
     exec.setTarget(target);
 
     const make_iso = switch (target.getCpuArch()) {
-        .i386 => b.addSystemCommand(&[_][]const u8{ "./makeiso.sh", boot_path, modules_path, iso_dir_path, exec.getOutputPath(), output_iso }),
+        .i386 => b.addSystemCommand(&[_][]const u8{ "./makeiso.sh", boot_path, modules_path, iso_dir_path, exec.getOutputPath(), ramdisk_path, output_iso }),
         else => unreachable,
     };
     make_iso.step.dependOn(&exec.step);
+
+    var ramdisk_files_al = ArrayList([]const u8).init(b.allocator);
+    defer ramdisk_files_al.deinit();
+
+    // Add some test files for the ramdisk runtime tests
+    if (test_mode == .Initialisation) {
+        try ramdisk_files_al.append("test/ramdisk_test1.txt");
+        try ramdisk_files_al.append("test/ramdisk_test2.txt");
+    }
+
+    const ramdisk_step = RamdiskStep.create(b, target, ramdisk_files_al.toOwnedSlice(), ramdisk_path);
+    make_iso.step.dependOn(&ramdisk_step.step);
 
     b.default_step.dependOn(&make_iso.step);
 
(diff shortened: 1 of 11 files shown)


build.zig
```diff
@@ -59,11 +63,28 @@ pub fn build(b: *Builder) !void {
...
-        .i386 => b.addSystemCommand(&[_][]const u8{ "./makeiso.sh", boot_path, modules_path, iso_dir_path, exec.getOutputPath(), output_iso }),
+        .i386 => b.addSystemCommand(&[_][]const u8{ "./makeiso.sh", boot_path, modules_path, iso_dir_path, exec.getOutputPath(), ramdisk_path, output_iso }),
         else => unreachable,
     };
     make_iso.step.dependOn(&exec.step);
 
+    const int_size: u16 = switch (target.getCpuArch()) {
+        .i386 => 32,
+        else => unreachable,
+    };
```
SamTebbs33: Can this be replaced by `@sizeOf(usize)`?

build.zig
```diff
@@ -149,3 +170,110 @@ pub fn build(b: *Builder) !void {
...
+        defer ramdisk.close();
+
+        // First write the number of files/headers
+        switch (self.int_size) {
+            32 => {
+                std.debug.assert(self.files.len < std.math.maxInt(u32));
+                try ramdisk.writer().writeIntNative(u32, @truncate(u32, self.files.len));
+            },
+            else => unreachable,
+        }
```
SamTebbs33: Perhaps passing the integer type to the RuntimeStep rather than the integer size would be nice? That would remove the need for all the switch statements.

build.zig
```diff
@@ -149,3 +170,110 @@ pub fn build(b: *Builder) !void {
...
+
+        // Open the out file
+        var ramdisk = try fs.cwd().createFile(self.out_file_path, .{});
+        defer ramdisk.close();
+
+        // First write the number of files/headers
+        switch (self.int_size) {
+            32 => {
+                std.debug.assert(self.files.len < std.math.maxInt(u32));
+                try ramdisk.writer().writeIntNative(u32, @truncate(u32, self.files.len));
```
SamTebbs33: Is it possible to write the integer in the target's endianness rather than the native endianness?

src/kernel/initrd.zig
```diff
@@ -0,0 +1,673 @@
...
+
+    /// See vfs.FileSystem.open
+    fn open(fs: *const vfs.FileSystem, dir: *const vfs.DirNode, name: []const u8, flags: vfs.OpenFlags) (Allocator.Error || vfs.Error)!*vfs.Node {
+        var self = @fieldParentPtr(RamdiskFS, "instance", fs.instance);
+        switch (flags) {
+            .CREATE_DIR, .CREATE_FILE => return vfs.Error.InvalidFlags,
+            .NO_CREATION => {
+                for (self.files) |*file| {
+                    if (std.mem.eql(u8, file.name, name)) {
+                        // Opening 2 files of the same name, will create 2 different Nodes
```
SamTebbs33, line 99: To avoid creating two nodes for the same file, you could create a list of nodes when initialising the ramdisk and then fetch from that when opening a node. That would also make destroying it upon closing unnecessary.

src/kernel/initrd.zig
```diff
@@ -0,0 +1,673 @@
...
+        const buff = try self.allocator.alloc(u8, length);
+        std.mem.copy(u8, buff, file_header.content[0..length]);
+        return buff;
+    }
+
+    /// See vfs.FileSystem.write
+    fn write(fs: *const vfs.FileSystem, node: *const vfs.FileNode, bytes: []const u8) (Allocator.Error || vfs.Error)!void {}
+
+    /// See vfs.FileSystem.open
+    fn open(fs: *const vfs.FileSystem, dir: *const vfs.DirNode, name: []const u8, flags: vfs.OpenFlags) (Allocator.Error || vfs.Error)!*vfs.Node {
```
SamTebbs33, line 92: It doesn't look like this uses the `dir` argument. Is it intended that the ramdisk doesn't support directories? That's fine, I'm just asking to be up to date.

src/kernel/initrd.zig
```diff
@@ -0,0 +1,673 @@
...
+    /// Return: *RamdiskFS
+    ///     A pointer to the ram disk file system.
+    ///
+    /// Error: Allocator.Error || Error
+    ///     error.OutOfMemory    - If there isn't enough memory for initialisation. Any memory
+    ///                            allocated will be freed.
+    ///     error.InvalidRamDisk - If the provided raw ramdisk is invalid. This can be due to a
+    ///                            mis-match of the number of files to the length of the raw
+    ///                            ramdisk or the wrong length provided to cause undefined parsed
+    ///                            lengths for other pars of the ramdisk.
```
SamTebbs33: s/pars/parts

src/kernel/initrd.zig
```diff
@@ -0,0 +1,673 @@
...
+    ///
+    pub fn init(rd_module: mem.Module, allocator: *Allocator) (Allocator.Error || Error)!*RamdiskFS {
+        log.logInfo("Init Ramdisk\n", .{});
+        defer log.logInfo("Done Ramdisk\n", .{});
+
+        const rd_len: usize = rd_module.region.end - rd_module.region.start;
+        const ramdisk_bytes = @intToPtr([*]u8, rd_module.region.start)[0..rd_len];
+        // First @sizeOf(usize) bytes is the number of files
+        const num_of_files = std.mem.readIntSlice(usize, ramdisk_bytes[0..], builtin.endian);
+        var headers = try allocator.alloc(RamdiskHeader, num_of_files);
```
SamTebbs33: Modifying the ramdisk structure to start with the number of headers, then all of the headers themselves and after that the file content would put the headers in a packed array meaning you can just cast the ramdisk start address (plus the size of the field for the number of headers) to a `[*]RamdiskHeader` and make a slice of that, instead of allocating memory and copying everything over (it's already in memory so let's take advantage of that). What do you think?

src/kernel/vfs.zig
```diff
@@ -383,6 +387,16 @@ pub fn isAbsolute(path: []const u8) bool {
     return path.len > 0 and path[0] == SEPARATOR;
 }
 
+///
+/// Initialise the virtual file system with a root Node. This will be a Directory node.
+///
+/// Arguments:
+///     IN node: *Node - The node to initialise the root node.
+///
+pub fn init(node: *Node) void {
```
SamTebbs33: Since we foresee this being used to switch root from the initrd to a disk filesystem at runtime, perhaps it should be called `setRoot` or similar?

SamTebbs33 (changes requested):
Looking great :) I was thinking about the name, and what do you think about the file extension being "ramdisk" or "rd" since not all ramdisks will be the "initial" ramdisk (which initrd is short for?)?

build.zig
SamTebbs33: Scratch that. Obviously that would use the host system's `usize`!

DrDeano: I wanted this, but the build file is ran as 64bit so `@sizeOf(usize)` return 8 instead of 4

DrDeano: I could pass the target to the ramdisk step to make this easier, plus can use the native endianness

build.zig
DrDeano: Will pass the target to the ramdisk step

build.zig
DrDeano: I did try this, but need comptime arch which isn't provided, instead will pass the target to the ramdisk step

src/kernel/initrd.zig
DrDeano: I wanted to create the *Nodes on the fly so not, this would save memory but would be slower. Having all the *Nodes created at init would increase memory, but would be faster. I don't mind having 2 different nodes for the same file. I can iterate through the hash map and search for any node that is opened (I did have this) but this would make opening even slower. What are your thoughts on this?

src/kernel/initrd.zig
DrDeano: It isn't used no, I wasn't sure how to use it anyway as yes, there are no directories and from your example, it is used for searching, but if there is a use for it, I would love to hear

src/kernel/initrd.zig
DrDeano: That is an idea. But we would need to know the name and content length to create the []u8 string. So just casting wouldn't work.

src/kernel/initrd.zig
SamTebbs33: To be honest, at the moment I can't think of any downsides to have two nodes per file (other than the extra memory allocated) so I'm gonna leave it up to you :)

src/kernel/initrd.zig
SamTebbs33: Well the directory node is the directory under which the file (`name` is a file name and extension, not a path) is created/opened, so it can be ignored if you don't plan for the ramdisk to support directories at the moment.

DrDeano: I see, the initrd will probably not have any directories, just to make it easier. In that case, shall I rename RamdiskFS to InitrdFS.

src/kernel/initrd.zig
DrDeano: I guess as the initrd will only be used once and load all files in one go, I could create all the Nodes and add a `free` or `deinit` to clear the initrd and reuse the memory after everything is loaded. This will also free the raw ramdisk so have all the memories.

src/kernel/initrd.zig
SamTebbs33: I'm cool with either

SamTebbs33 (approved):

src/kernel/initrd.zig
SamTebbs33: That sounds good :)

PR #208, opened 2020-07-25
Bugfix/paging map 4 mb boundary

### LINKED ISSUE #207: IndexOutOfBounds when mapping pages across 4MB boundaries
If we have an address that crosses a 4MB (0x40 0000) boundary for example: 0xC030 0000 - 0xC050 0000, the `mapDirEntry` will be given the address range: 0xC030 0000 - 0xC050 0000 which results in a index out of bounds. As this address range spans 2 tables, `mapDirEntry` will need to be give the address ranges 0xC030 0000 - 0xC040 0000, then 0xC040 0000 - 0xC050 0000.

DrDeano:
Closes #207


SamTebbs33 (approved):
Nice :+1:

PR #215, opened 2020-08-01
Feature/device interface

### LINKED ISSUE #214: Add a device interface that abstract the underlying hardware for the file system
As we now have the basis of the virtual file system, it would be good to hide the underlying device the the file system will sit on. This can include a RAM device, hard drive device, etc. A device will only need a read and write operation to a given offset and length.

DrDeano:
This also makes some changed to the heap resize so when shrinking won't shrink past a point: `@sizeOf(Header)`;

Closes #214


src/kernel/filesystem/ram_device.zig
```diff
@@ -0,0 +1,228 @@
...
+    ///     IN allocator: *Allocator - The allocator when needing to allocate memory.
+    ///
+    /// Return: *RamDevice
+    ///     A new RAM device.
+    ///
+    /// Error: Allocator.Error
+    ///     error.OutOfMemory - If there isn't enough memory when initialising the RAM device.
+    ///                         Any memory allocated will be freed on return.
+    ///
+    pub fn init(memory: []u8, allocator: *Allocator) Allocator.Error!*RamDevice {
```
SamTebbs33, line 73: Could this return a struct instead to allow the caller to decide if it should be on the heap or stack?

src/kernel/filesystem/ram_device.zig
```diff
@@ -0,0 +1,228 @@
...
+        errdefer allocator.destroy(device);
+
+        ram_device.* = .{
+            .device = device,
+            .allocator = allocator,
+            // QUESTION: Should we dupe the memory? We could just have the memory with the same life time as the pointer passed (caller)
+            //           Then just ass a comment in the doc comment about the life time.
+            //           This would improve speed and memory usage, but the device should control the memory.
+            //           Could add 2 init function, one with a size to allocate and one with a memory pointer (this one).
+            .memory = try allocator.dupe(u8, memory),
```
SamTebbs33, line 86: My understanding is that this RAM device receives a region of memory to manage, rather than creating its own region? I think taking control of the given slice would make this more flexible and would cut down on allocations.

src/kernel/filesystem/ram_device.zig
```diff
@@ -0,0 +1,228 @@
...
+        const device = try allocator.create(Device);
+
+        ram_device.* = .{
+            .device = device,
+            .allocator = allocator,
+            // QUESTION: Should we dupe the memory? We could just have the memory with the same life time as the pointer passed (caller)
+            //           Then just ass a comment in the doc comment about the life time.
+            //           This would improve speed and memory usage, but the device should control the memory.
+            //           Could add 2 init function, one with a size to allocate and one with a memory pointer (this one).
+            .memory = memory,
```
SamTebbs33, line 112: I think this is a better solution than in `init`.

src/kernel/heap.zig
```diff
@@ -381,12 +388,12 @@ const FreeListAllocator = struct {
             }
 
             // If there is enough unused space to the right of this node then create a smaller node
-            if ((@sizeOf(Header) + header.size) - alignment_padding - real_size > @sizeOf(Header)) {
-                header.next_free = insertFreeHeader(@ptrToInt(header) + real_size + alignment_padding, header.size + @sizeOf(Header) - real_size - alignment_padding - @sizeOf(Header), header.next_free);
+            if (header.size - alignment_padding > real_size) {
```
SamTebbs33, line 391: Nice. I had the extra `@sizeOf(Header)` in there before to avoid underflow, but that doesn't seem to be a problem.

src/kernel/heap.zig
```diff
@@ -244,10 +244,16 @@ const FreeListAllocator = struct {
             return Allocator.Error.OutOfMemory;
         } else {
             // Shrinking
-            const size_diff = old_mem.len - real_size;
+            var size_diff = old_mem.len - real_size;
             // If shrinking would leave less space than required for a new header,
             // or if shrinking would make the buffer too small, don't shrink
-            if (size_diff < @sizeOf(Header) or real_size < @sizeOf(Header)) return Allocator.Error.OutOfMemory;
+            if (size_diff < @sizeOf(Header)) {
+                return old_mem.len;
```
SamTebbs33, line 251: I believe that the `Allocator` API says that `OutOfMemory` should be returned if no resizing is happening, but that may be incorrect since there is some inconsistency in the documentation.

src/kernel/kmain.zig
```diff
@@ -106,6 +107,14 @@ export fn kmain(boot_payload: arch.BootPayload) void {
...
     if (rd_module) |module| {
         // Load the ram disk
+        // Create a RamDevice
+        // const rd_len: usize = module.region.end - module.region.start;
+        // const ramdisk_bytes = @intToPtr([*]u8, module.region.start)[0..rd_len];
+        // Tbh, there is a lot, A LOT of memory coping. It would be faster to just use the ram disk in memory already.
+        // It would be nice to use the ram device for the initrd but would but unnecessary.
+        // var ram_disk_device = ram_device.RamDevice.init(ramdisk_bytes, &kernel_heap.allocator) catch |e| {
+        //     panic_root.panic(@errorReturnTrace(), "Failed to initialise ramdisk device: {}\n", .{e});
+        // };
```
SamTebbs33, line 117: Here there be commented code.

src/kernel/filesystem/device.zig
```diff
@@ -0,0 +1,61 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+
+/// The functions for a hardware device. A device can be a RAM device where the all content is in
+/// RAM, or can be a hard disk drive device for a floppy or CD etc.
+pub const Device = struct {
```
SamTebbs33, line 6: I was thinking about the name and wondered if something like `Resource` would be more appropriate to avoid confusion with hardware devices (like mice, graphics cards, controllers etc.)? If we can't think of a better name then `Device` is fine.

SamTebbs33 (changes requested):
This is a great idea!

src/kernel/filesystem/ram_device.zig
DrDeano: Sure

src/kernel/filesystem/ram_device.zig
DrDeano: Yeh, but my concern was that if the caller frees the memory and still tries to use the device will cause undefined behavior. So will have a comment saying not to free the memory.

src/kernel/heap.zig
DrDeano: I had to do this, bc I have a shrink that is lees than the size of a header, and there is nothing else I can do but do this.

src/kernel/kmain.zig
DrDeano: I was going to use the new RamDevice, but there just seems to be too much allocations for a RAM device. Should the read function just return a slice of the memory instead of allocating it?

src/kernel/filesystem/device.zig
DrDeano: How about `StorageDevice`

src/kernel/kmain.zig
SamTebbs33: The initrd read function or the RamDevice read function?

src/kernel/filesystem/device.zig
SamTebbs33: Hmm "storage" typically means persistent storage, like disks, flash drives etc.

src/kernel/kmain.zig
DrDeano: The RamDevice read function, as it allocated memory for the read.

src/kernel/filesystem/device.zig
DrDeano: I guess `Resource` is fine

DrDeano:
This PR is now not needed as a ram device can be implemented as a FixedBufferStream and other devices/resources can be implemented in the same way.

PR #218, opened 2020-08-05
rpi led and serial

markfirmware:
zig build -Dtarget=aarch64-freestanding

Then copy rpi-sdcard.img to an unused sd card and insert in 3b+

Will flash rpi3b+ activity led quickly on exception. Pins 14/15 have serial tx/rx.


build.zig
```diff
@@ -160,4 +160,49 @@ pub fn build(b: *Builder) !void {
         "target remote localhost:1234",
     });
     debug_step.dependOn(&debug_cmd.step);
+
+    const zip_folder = "rpi-sdcard";
```
SamTebbs33: Instead of zipping, could a .img file not be created that has the kernel, config file et.c inside it? Then that could be copied directly to an SD card instead of having to unzip.

build.zig
```diff
@@ -160,4 +160,49 @@ pub fn build(b: *Builder) !void {
...
+        firmware_url ++ "start.elf",
+    });
+    const zip_do_zip_cmd = b.addSystemCommand(&[_][]const u8{
+        "zip",
+        "-jqr",
+        zip_folder ++ ".zip",
+        zip_folder,
+    });
+
+    const zip_step = b.step("zip", "Create product zip file");
```
SamTebbs33: I think it would be best to do all of this under the default step, rather than in a specialised zip step.

build.zig
SamTebbs33: I think it would be cleaner to put the build zip/image file into the output directory (`b.cache_root`). This will keep built files in one place.

src/kernel/arch/aarch64/arch.zig
```diff
@@ -35,50 +36,38 @@ pub fn initTTY(boot_payload: BootPayload) TTY {
...
+        mmio.write(mmio_addr, .AUX_ENABLES, 1);
+        mmio.write(mmio_addr, .AUX_MU_IER_REG, 0);
+        mmio.write(mmio_addr, .AUX_MU_CNTL_REG, 0);
+        mmio.write(mmio_addr, .AUX_MU_LCR_REG, 3);
+        mmio.write(mmio_addr, .AUX_MU_MCR_REG, 0);
+        mmio.write(mmio_addr, .AUX_MU_IER_REG, 0);
+        mmio.write(mmio_addr, .AUX_MU_IIR_REG, 0xc6);
+        mmio.write(mmio_addr, .AUX_MU_BAUD_REG, 270);
+        mmio.write(mmio_addr, .AUX_MU_CNTL_REG, 3);
+    }
```
SamTebbs33: Why did all of this need to be changed? All the comments, system clock configuration and such has been removed.

src/kernel/arch/aarch64/arch.zig
```diff
@@ -35,50 +36,38 @@ pub fn initTTY(boot_payload: BootPayload) TTY {
...
 }
 
-fn uartWriteByte(byte: u8) void {
-    // Wait until the UART is ready to transmit
-    while ((mmio.read(mmio_addr, mmio.Register.UART_FLAGS) & (1 << 5)) != 0) {}
-    mmio.write(mmio_addr, mmio.Register.UART_DATA, byte);
+pub fn uartWriteByte(byte: u8) void {
+    if (byte == 10) {
+        uartWriteByte(13);
+    }
```
SamTebbs33, line 72: What's the purpose of this check?

src/kernel/arch/aarch64/arch.zig
```diff
@@ -120,5 +109,133 @@ pub fn haltNoInterrupts() noreturn {
 
 // TODO: implement
 pub fn spinWait() noreturn {
-    while (true) {}
+    spinLed(500);
```
SamTebbs33: I think this should still be a while true loop (it should probably include a `wfe` but that can wait) since you won't always want to make the LED flash when spinning.

src/kernel/arch/aarch64/arch.zig
```diff
@@ -120,5 +109,133 @@ pub fn haltNoInterrupts() noreturn {
...
+        \\ mov x2, #0x08000000
+        \\ orr x1, x1, x2
+        \\ str w1, [x0, #0x08]
+        \\ mov x1, #0x20000000
+        \\ str w1, [x0, #0x28]
+        :
+        :
+        : "x0", "x1", "x2"
+    );
+}
```
SamTebbs33: Could these use more zig code instead? We like to keep raw assembly to the minimum.

src/kernel/arch/aarch64/arch.zig
```diff
@@ -120,5 +109,133 @@ pub fn haltNoInterrupts() noreturn {
...
+    AlternateFunction3 = 7,
+    AlternateFunction4 = 3,
+    AlternateFunction5 = 2,
+};
+
+const PinPull = enum {
+    None,
+    Down,
+    Up,
+};
```
SamTebbs33: The raspberry pi-specific things (like LEDs and pins) should be moved to the rpi.zig file since they are specific to the rpi.

src/kernel/arch/aarch64/arch.zig
```diff
@@ -120,5 +109,133 @@ pub fn haltNoInterrupts() noreturn {
...
+        \\ msr ttbr0_el3, x1
+        \\ isb
+        \\ mrs x0, sctlr_el3
+        \\ orr x0, x0, #1
+        \\ msr sctlr_el3, x0
+        \\ isb
+        :
+        :
+        : "x0", "x1", "x2", "x3"
+    );
```
SamTebbs33: Same again regarding the raw assembly code.

src/kernel/arch/aarch64/boot.zig
```diff
@@ -8,32 +9,39 @@ extern var KERNEL_STACK_END: *usize;
...
+    asm volatile (
+        \\ mrs x0, mpidr_el1
+        \\ mov x1, #3
+        \\ ands x0, x0, x1
+        \\ beq cpu0
+        \\hang:
+        \\ wfe
+        \\ b hang
+        \\cpu0:
+    );
```
SamTebbs33: Raw assembly again :)

src/kernel/arch/aarch64/boot.zig
```diff
@@ -8,32 +9,39 @@ extern var KERNEL_STACK_END: *usize;
...
+        \\ msr vbar_el3, %[table_addr]
         :
         : [table_addr] "r" (@ptrToInt(&interrupts.exception_table))
     );
 
+    arch.enableFlatMmu();
+
     // The rpi puts the board part number in midr_el1
-    const board = detectBoard();
+    const board = rpi.RaspberryPiBoard.RaspberryPi3; // detectBoard();
```
SamTebbs33: Why not `detectBoard()`?

src/kernel/arch/aarch64/mmio.zig
```diff
@@ -1,31 +1,87 @@
-const GPIO_OFFSET: usize = 0x200000;
-const UART_OFFSET: usize = GPIO_OFFSET + 0x1000;
+const log = @import("../../log.zig");
+const std = @import("std");
+
+const MMIO_OFFSET: usize = 0x200000;
+const GPIO_OFFSET: usize = MMIO_OFFSET + 0x0000;
+const UART_OFFSET: usize = MMIO_OFFSET + 0x1000;
+const AUX_OFFSET: usize = MMIO_OFFSET + 0x15000;
```
SamTebbs33, line 7: Nice. Making them relative to mmio_offset is more correct.

src/kernel/arch/aarch64/mmio.zig
```diff
@@ -1,31 +1,87 @@
...
+    GPIO_CLR0 = GPIO_OFFSET + 0x28,
+    GPIO_CLR1 = GPIO_OFFSET + 0x2c,
+    GPIO_PUD = GPIO_OFFSET + 0x94,
+    GPIO_PUDCLK0 = GPIO_OFFSET + 0x98,
+    GPIO_PUDCLK1 = GPIO_OFFSET + 0x9c,
     MBOX_STATUS = MBOX_OFFSET + 0x18,
     MBOX_WRITE = MBOX_OFFSET + 0x20,
     MBOX_READ = MBOX_OFFSET + 0x00,
+    UART_DR = UART_OFFSET + 0x00,
+    UART_FR = UART_OFFSET + 0x18,
```
SamTebbs33: For the registers that existed before, could we keep old names? They're a bit more descriptive.

src/kernel/arch/aarch64/mmio.zig
```diff
@@ -1,31 +1,87 @@
...
 };
 
 pub fn write(mmio_base: usize, reg: Register, data: u32) void {
     const addr = mmio_base + @enumToInt(reg);
-    const ptr = @intToPtr(*u32, addr);
+    const ptr = @intToPtr(*volatile u32, addr);
```
SamTebbs33, line 52: Good spot :+1:

SamTebbs33 (changes requested):
Good job on getting it working in hardware! You're not going to like me for requesting this but could you add doc comments above each function, struct, struct field and public variable? Documentation is important :) Have a look at src/kernel/filesystem/vfs.zig for inspiration.

build.zig
SamTebbs33: Of course these zipping commands and such shouldn't be added when building for x86 so a check for aarch64 would be needed too.

src/kernel/arch/aarch64/arch.zig
markfirmware: This sends a carriage return before a line feed. Should I add a comment or is there a better place to do this?

src/kernel/arch/aarch64/arch.zig
markfirmware: Yes. I wanted to get your eyes on this before proceeding further. Updating tcr and ttbr0 is delicate and will probably still be asm copied from the arm manual.

src/kernel/arch/aarch64/arch.zig
markfirmware: Done: b86e4cdb62bf8cd9c6ebd743986baca6d3e839ce

src/kernel/arch/aarch64/boot.zig
markfirmware: This is necessary because the disassembly of the zig version showed the use of the sp which is not yet initialized. Each core will need its own sp so sp cannot be initialized yet.

I have added a comment: a161e9f4352fd8f183017710984af0c6335c1e79

src/kernel/arch/aarch64/boot.zig
markfirmware: ~~I suspect a compiler defect when returning an optional struct from a function. I intend to reduce and isolate this when I am at my lab today.~~

67dfa7a7e650cb8295daee71b9337a082df720f5

Turns out the problem was returning the address of a const on the stack.

src/kernel/arch/aarch64/mmio.zig
markfirmware: Thanks.

build.zig
markfirmware: It can be done but would still require something like https://www.balena.io/etcher/ to do the copying.

Could we open an issue to improve this later?

markfirmware: > Good job on getting it working in hardware! You're not going to like me for requesting this but could you add doc comments above each function, struct, struct field and public variable? Documentation is important :) Have a look at src/kernel/filesystem/vfs.zig for inspiration.

Will do.

src/kernel/arch/aarch64/mmio.zig
markfirmware: Thanks.

src/kernel/arch/aarch64/arch.zig
markfirmware: Done: 313b85716e147c5a5393fb938d40fedf4932d8f4

src/kernel/arch/aarch64/mmio.zig
markfirmware: The reference for these is https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2835/BCM2835-ARM-Peripherals.pdf which also applies generally to the bcm2837 which is on the pi3b and pi3b+.

The names I used are from this manual but using your prefix (GP instead of GPIO.) What I would like to do is use the ones straight from the manual, unmodified. They are not as descriptive, but they are what everyone else is using.

Your decision.

src/kernel/arch/aarch64/arch.zig
markfirmware: Uart0 is the full uart and uart1 is the mini (aux) uart. Uart0 is used for the bluetooth chip on 3b and above, leaving uart1 for anything else like the serial console.

However, qemu does not implement uart1, it only implements uart0, and of course it does not implement the bluetooth chip.

Therefore uart1 is used for the console except when under qemu. Under qemu uart0 is used for the console. Also qemu uart0 does not require initialization so uart0 constants were removed. Uart0 constants and initialization would be added back, probably differently, for bluetooth when we get to it.

By the way, the uart0 interface to the bluetooth chip is quite elegant and easy to program.

src/kernel/arch/aarch64/arch.zig
markfirmware: These are the instructions for updating tcr and ttbr0. This is the 1.0 manual. There is a 1.1 manual but it does not have an online link - the 1.1 version can only be used locally. This particukar section of the manual is the same.

https://static.docs.arm.com/100940/0100/armv8_a_address%20translation_100940_0100_en.pdf#%5B%7B%22num%22%3A53%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C54%2C456%2C0%5D

markfirmware (commented):
Some of the comments can be resolved. @SamTebbs33

build.zig
SamTebbs33: Oh I don't mean that the SD card should be flashed with this step, just that an image file is created (that can be put on an SD card manually), just like we do with `grub-mkrescue` for x86.

src/kernel/arch/aarch64/arch.zig
SamTebbs33: I see. So does uart1 not depend on the system clock? This code could do with some comments to explain what is happening.

src/kernel/arch/aarch64/arch.zig
SamTebbs33: Hmm sending a carriage return as well sounds very windows-like. Is this standard in serial communication?

src/kernel/arch/aarch64/arch.zig
SamTebbs33: > Yes. I wanted to get your eyes on this before proceeding further. Updating tcr and ttbr0 is delicate and will probably still be asm copied from the arm manual.

Sure, if something has to use inline assembly that's fine. Register moves, arithmetic and the like should preferably use Zig.

src/kernel/arch/aarch64/boot.zig
SamTebbs33: Oh well spotted. It's stupid use of the stack too since the stored value is loaded immediately afterwards. Perhaps you could read `mpidr_el1` in assembly then do the comparison with 0 and loop in Zig?

src/kernel/arch/aarch64/boot.zig
SamTebbs33: Nice!

src/kernel/arch/aarch64/mmio.zig
SamTebbs33: I see your reasoning. I still think that more descriptive names are better though. Perhaps we can convince the rest of the world too :)

SamTebbs33 (changes requested):

src/kernel/arch/aarch64/arch.zig
markfirmware: Somewhere along the line a carriage return is needed for the terminal to move the print head to the left position. Line feed only advances the paper. This is real teletype stuff.

This is needed for an attached tty or a terminal emulator like minicom.

src/kernel/arch/aarch64/boot.zig
markfirmware: Possibly, but that loop might be using the stack.

Can we keep it assembly in the pr and have an issue to improve it?

src/kernel/arch/aarch64/mmio.zig
markfirmware: Ok, the hex offsets are an unambiguous link between the source and the manual, so more descriptive names will work.

src/kernel/arch/aarch64/arch.zig
SamTebbs33: OK that makes sense. You could add a brief comment to explain that.

markfirmware: Done: 43da14b643f868eb9a12dc80e4c4ddcb8cd21117

src/kernel/arch/aarch64/boot.zig
SamTebbs33: Have a look and see if looping does end up using the stack in the debug build mode. If it does then I reckon you can keep it as it is.

build.zig
markfirmware: https://forums.balena.io/t/how-do-i-create-an-sd-disk-image-file/171673

src/kernel/arch/aarch64/boot.zig
FireFox317: I think an idea might be to keep the `_start` function really small and only setup a temporary stack (with asm volatile) and call another function. That way the `_start` function will not use the stack and the next function can do the checking of the core number etc and afterwards setting up the correct stack. The small temporary stack only has to be like 256 bytes or something small.

Or just setup a small temporary stack as the first step in the `_start` function.

SamTebbs33: > I think an idea might be to keep the `_start` function really small and only setup a temporary stack (with asm volatile) and call another function. That way the `_start` function will not use the stack and the next function can do the checking of the core number etc and afterwards setting up the correct stack. The small temporary stack only has to be like 256 bytes or something small.
> 
> Or just setup a small temporary stack as the first step in the `_start` function.

@FireFox317 That sounds ideal to me. The only consideration would be if the cores using the same stack would cause a conflict.

src/kernel/arch/aarch64/rpi.zig
```diff
@@ -47,6 +48,7 @@ pub inline fn turnOnLed() void {
 }
 
 pub inline fn turnOffLed() void {
+    // do not use zig here in order to side-step occasional compiler defects
```
SamTebbs33: What compiler defects have you encountered in `turnOffLed` and `turnOnLed`?

src/kernel/arch/aarch64/boot.zig
```diff
@@ -20,6 +20,9 @@ export fn _start() linksection(".text.boot") callconv(.Naked) noreturn {
         \\ wfe
         \\ b hang
         \\cpu0:
+        \\ mov x0, #0x3f200000
+        \\ mov w1, #67
+        \\ str w1, [x0, 0x1000]
```
SamTebbs33: What is this doing? It's also best to avoid using unclear constants.

SamTebbs33 (changes requested):

build.zig
SamTebbs33: I believe that something like this would work https://unix.stackexchange.com/questions/90793/create-iso-image-from-folder-via-terminal-commands/90794

src/kernel/arch/aarch64/boot.zig
FireFox317: @SamTebbs33 yeah obviously you want different stacks per core, otherwise the cores would overwrite the data of the other core. I would do something like the following:
```asm
mrs x0, mpidr_el1 // get system register
and x0, x0, #3 // mask out bits not belonging to core id
mul x1, x0, #255 // 255 bytes per stack
mov x2, <stack_address> // store the lowest stack address in x2
sub x2, x2, x1 
mov sp, x2
```
This can probably written in smaller assembly.

src/kernel/arch/aarch64/boot.zig
markfirmware: This was residue from development. I have removed it.

src/kernel/arch/aarch64/mmio.zig
markfirmware: > For the registers that existed before, could we keep old names? They're a bit more descriptive.

Done: 0a034d3c956f98a70001b24ed4453ccb3520fe9d

src/kernel/arch/aarch64/boot.zig
markfirmware: @SamTebbs33 I will change to this approach.

markfirmware: @SamTebbs33 @FireFox317

boot.zig: 

https://github.com/markfirmware/pluto/blob/1b04bd80a27c560a7777542e38d07f2844ff8ecd/src/kernel/arch/aarch64/boot.zig#L1-L50

cpu registers:

https://github.com/markfirmware/pluto/blob/1b04bd80a27c560a7777542e38d07f2844ff8ecd/src/kernel/arch/aarch64/arch.zig#L118-L178

src/kernel/arch/aarch64/arch.zig
```diff
@@ -113,6 +115,74 @@ pub fn spinWait() noreturn {
     while (true) {}
 }
 
+pub const cpu = struct {
```
SamTebbs33: This is really nice. We could consider using this in x86 in the future! Don't forget to capitalise `Cpu`.

src/kernel/arch/aarch64/boot.zig
```diff
@@ -1,49 +1,57 @@
...
+            cpu.wfe();
+        }
+    }
+
+    // Give all exception levels the same vector table
+    cpu.vbar.el(1).write(@ptrToInt(&interrupts.exception_table));
+    cpu.vbar.el(2).write(@ptrToInt(&interrupts.exception_table));
+    cpu.vbar.el(3).write(@ptrToInt(&interrupts.exception_table));
+
+    interrupts.exception_handler_depth = 0;
```
SamTebbs33: Very nice

src/kernel/arch/aarch64/boot.zig
```diff
@@ -1,49 +1,57 @@
...
 
     arch.enableFlatMmu();
 
-    // The rpi puts the board part number in midr_el1
     board = detectBoard();
+    arch.initMmioAddress(&board);
+
+    // invoke alignment exception
+    // var addr: usize = 1;
+    // @intToPtr(*volatile u32, addr).* = 0;
```
SamTebbs33: This looks like it's for testing so remember to remove it before we merge.

SamTebbs33 (changes requested):

src/kernel/arch/aarch64/boot.zig
SamTebbs33: Cheers

SamTebbs33:
Other then the comments from my latest review, it looks like all that is needed is some documentation and comments in `initSerial`.

src/kernel/arch/aarch64/boot.zig
markfirmware: Done.

src/kernel/arch/aarch64/boot.zig
markfirmware: Thanks

src/kernel/arch/aarch64/arch.zig
markfirmware: Thanks. Done.

src/kernel/arch/aarch64/arch.zig
markfirmware: This is no longer assembly but the constants still need explanation.

markfirmware:
> Other then the comments from my latest review, it looks like all that is needed is some documentation and comments in `initSerial`.

Yes including a few questions that you made that I have not yet answered.

build.zig
markfirmware: I took a look but it did not look like a complete solution.

build.zig
markfirmware: Done: https://github.com/markfirmware/pluto/blob/da14037897a64676ef5fcc4d63e23a102690f842/build.zig#L70-L85

build.zig
markfirmware: Done: https://github.com/markfirmware/pluto/blob/da14037897a64676ef5fcc4d63e23a102690f842/build.zig#L70

build.zig
```diff
@@ -67,9 +67,20 @@ pub fn build(b: *Builder) !void {
     exec.setLinkerScriptPath(linker_script_path);
     exec.setTarget(target);
 
+    const zip_folder = try fs.path.join(b.allocator, &[_][]const u8{ b.cache_root, "rpi-sdcard" });
+    const kernel = try fs.path.join(b.allocator, &[_][]const u8{ zip_folder, "kernel8.img" });
+    const firmware_version_tag = "1.20200601+arm64";
+    const firmware_url = "https://github.com/raspberrypi/firmware/raw/" ++ firmware_version_tag ++ "/boot/";
+    const image_file_name = try fs.path.join(b.allocator, &[_][]const u8{ exec.output_dir.?, "kernel8.img" });
```
SamTebbs33: Can these be declared inside the .aarch64 branch of the switch?

build.zig
```diff
@@ -67,9 +67,20 @@ pub fn build(b: *Builder) !void {
...
     const make_iso = switch (target.getCpuArch()) {
         .i386 => b.addSystemCommand(&[_][]const u8{ "./makeiso.sh", boot_path, modules_path, iso_dir_path, exec.getOutputPath(), output_iso }),
-        .aarch64 => b.addSystemCommand(&[_][]const u8{ "aarch64-linux-gnu-objcopy", exec.getOutputPath(), "-O", "binary", try fs.path.join(b.allocator, &[_][]const u8{ exec.output_dir.?, "kernel8.img" }) }),
+        .aarch64 =>
+        // armstub not yet working, therefore: b 0x80000 (which is 0x14020000) followed by 0 filler until 0x80000 followed by kernel that starts at 0x80000
```
SamTebbs33: What's wrong with armstub?

build.zig
```diff
@@ -67,9 +67,20 @@ pub fn build(b: *Builder) !void {
...
+        b.addSystemCommand(&[_][]const u8{
+            "/bin/bash",
+            "-c",
+            try std.fmt.allocPrint(b.allocator, "aarch64-linux-gnu-objcopy {} -O binary {} && mkdir --parents {} && cp --archive src/kernel/arch/aarch64/rpi-sdcard/config.txt {} && echo -ne \"\\x00\\x00\\x02\\x14\" > {} && dd status=none bs=1 count=524284 if=/dev/zero >> {} && cat {}/kernel8.img >> {} && wget --directory-prefix={} --quiet --timestamp {}/bootcode.bin {}/fixup.dat {}/start.elf && zip --junk-paths --quiet --recurse-paths {}.zip {}", .{ exec.getOutputPath(), image_file_name, zip_folder, zip_folder, kernel, kernel, b.cache_root, kernel, zip_folder, firmware_url, firmware_url, firmware_url, zip_folder, zip_folder }),
```
SamTebbs33: It would be cleaner for these to be separate system commands that each depends on the previous one.

SamTebbs33 (changes requested):

build.zig
markfirmware: Done: https://github.com/markfirmware/pluto/blob/da9b1aaef9ef6261e1f6d4001b510201e248a44c/build.zig#L72-L77

@SamTebbs33

build.zig
markfirmware: Done: https://github.com/markfirmware/pluto/blob/da9b1aaef9ef6261e1f6d4001b510201e248a44c/build.zig#L72-L142

@SamTebbs33

build.zig
markfirmware: I do not know. I was unable to get it to work. I can open  an issue to improve this.

@SamTebbs33

build.zig
```diff
@@ -67,20 +67,79 @@ pub fn build(b: *Builder) !void {
...
+            const cp_config = b.addSystemCommand(&[_][]const u8{
+                "cp",
+                "--archive",
+                "src/kernel/arch/aarch64/rpi-sdcard/config.txt",
+                zip_folder,
+            });
+            cp_config.step.dependOn(&objcopy.step);
+
+            const emulate_armstub = armstubSequence: {
+                // armstub not yet working, therefore: b 0x80000 (which is 0x14020000)
```
SamTebbs33: What's wrong with armstub?

build.zig
```diff
@@ -67,20 +67,79 @@ pub fn build(b: *Builder) !void {
...
+                const create_b_0x80000 = bash(b, "echo -ne \"\\x00\\x00\\x02\\x14\" > {}", .{kernel});
+                create_b_0x80000.step.dependOn(&cp_config.step);
+
+                // followed by 0 filler until 0x80000
+                const dd = bash(b, "dd status=none bs=1 count=524284 if=/dev/zero >> {}", .{kernel});
+                dd.step.dependOn(&create_b_0x80000.step);
+
+                // followed finally by kernel that starts at 0x80000
+                const cat = bash(b, "cat {}/kernel8.img >> {}", .{ b.cache_root, kernel });
```
SamTebbs33: Why is this necessary? Shouldn't it just be a case of packaging the kernel image, config etc. into a .img file?

build.zig
```diff
@@ -67,20 +67,79 @@ pub fn build(b: *Builder) !void {
...
+            const wget = b.addSystemCommand(&[_][]const u8{
+                "wget",
+                try std.fmt.allocPrint(b.allocator, "--directory-prefix={}", .{zip_folder}),
+                "--quiet",
+                "--timestamp",
+                try std.fmt.allocPrint(b.allocator, "{}/bootcode.bin", .{firmware_url}),
+                try std.fmt.allocPrint(b.allocator, "{}/fixup.dat", .{firmware_url}),
+                try std.fmt.allocPrint(b.allocator, "{}/start.elf", .{firmware_url}),
+            });
+            wget.step.dependOn(&emulate_armstub.step);
```
SamTebbs33: Can we store these in the repo instead of downloading them? If licensing doesn't allow this then we could request the user download them and put them in a specific place.

build.zig
```diff
@@ -67,20 +67,79 @@ pub fn build(b: *Builder) !void {
...
+
+            const zip = b.addSystemCommand(&[_][]const u8{
+                "zip",
+                "--junk-paths",
+                "--quiet",
+                "--recurse-paths",
+                try std.fmt.allocPrint(b.allocator, "{}.zip", .{zip_folder}),
+                zip_folder,
+            });
+            zip.step.dependOn(&wget.step);
```
SamTebbs33: Why do we need a zip now that we have the image file that we should be able to burn directly to the SD card?

build.zig
```diff
@@ -172,3 +231,11 @@ pub fn build(b: *Builder) !void {
     });
     debug_step.dependOn(&debug_cmd.step);
 }
+
+fn bash(b: *Builder, comptime fmt: []const u8, args: anytype) *std.build.RunStep {
+    return b.addSystemCommand(&[_][]const u8{
+        "bash",
```
SamTebbs33: As discovered by marler8997, not all systems have bash (#131). I suggest using `b.addSystemCommand` instead.

SamTebbs33 (changes requested):

SamTebbs33:
Thanks for cleaning up the image creation code in build.zig :+1:

build.zig
markfirmware: This is part of putting the image at 0x80000 with a jump instruction at 0. The bash has been replaced by zig.

https://github.com/markfirmware/pluto/blob/0fb94376e5622874046f0d874f15966c373f59fa/build.zig#L214-L242

markfirmware: The entire disk is now packaged into rpi-sdcard.img.

markfirmware: @SamTebbs33

SamTebbs33:
Thanks for replacing the wget and bash :+1: Including the firmware files here means that we need to replicate their copyright notices in the README, but I can do that before finishing the port. I think there are still a few unresolved comments left then we can go ahead and merge.

src/kernel/arch/aarch64/rpi.zig
markfirmware: There were some alignment faults some time ago, but they have not recurred since enabling the mmu. I have removed this unneeded code.

src/kernel/arch/aarch64/arch.zig
markfirmware: I have added comments regarding the use of uart0/uart1 in relation to qemu and also to the system clock. enable_uart=1 is now included in config.txt to ensure the baud rate is stable.

markfirmware:
I have added the initSerial comments and removed the now unnecessary assembly code for the led.

I am now starting the /// commentary.

markfirmware:
@SamTebbs33 Please check  997d157 for satisfaction.

SamTebbs33 (approved):
Thanks @markfirmware :+1:. It's great to have your changes to make it work on real hardware!

PR #219, opened 2020-08-06
Feature/initrd use fixed buffer stream

DrDeano:
This moved the initrd ramdisk to use a FixedbufferStream to read the raw ramdisk in memory.
Additional fixed allocations of sized that are not aligned to `Header`.


SamTebbs33 (approved):
Noice

PR #222, opened 2020-08-15
Add new parameters for resize and alloc

SamTebbs33:
Zig update.﻿

### DIFF
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -246,6 +246,7 @@ fn traversePath(path: []const u8, flags: OpenFlags) (Allocator.Error || Error)!*
         child: []const u8,
 
         const Self = @This();
+
         fn func(split: *std.mem.SplitIterator, node: *Node, rec_flags: OpenFlags) (Allocator.Error || Error)!Self {
             // Get segment string. This will not be unreachable as we've made sure the spliterator has more segments left
             const seg = split.next() orelse unreachable;
@@ -560,8 +561,8 @@ test "mount" {
 
     // The root fs
     var testfs = try testInitFs(allocator);
-    defer testfs.deinit();
     defer allocator.destroy(testfs);
+    defer testfs.deinit();
 
     testfs.instance = 1;
     root = testfs.tree.val;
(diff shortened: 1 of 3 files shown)


src/kernel/filesystem/vfs.zig
```diff
@@ -560,8 +561,8 @@ test "mount" {
 
     // The root fs
     var testfs = try testInitFs(allocator);
-    defer testfs.deinit();
     defer allocator.destroy(testfs);
+    defer testfs.deinit();
```
DrDeano, line 565: I think it still needs to be `deinit` then `free` as when freeing first, will be deiniting a free pointer

src/kernel/filesystem/vfs.zig
```diff
@@ -594,8 +595,8 @@ test "mount" {
 test "traversePath" {
     var allocator = testing.allocator;
     var testfs = try testInitFs(allocator);
-    defer testfs.deinit();
     defer allocator.destroy(testfs);
+    defer testfs.deinit();
```
DrDeano, line 599: Same here (5 in total)

src/kernel/heap.zig
```diff
@@ -182,16 +182,18 @@ const FreeListAllocator = struct {
     /// Arguments:
     ///     IN allocator: *std.Allocator - The allocator to resize within.
     ///     IN old_mem: []u8 - The buffer to resize.
+    ///     IN mem_align: u29 - The original alignment for old_mem.
```
DrDeano: Maybe `old_align` would be better

src/kernel/filesystem/vfs.zig
SamTebbs33: I was getting seg faults when using the previous order and upon looking at the docs I read that the deferred statements are run in reverse order. which is why I defer deinitting the testfs last.

src/kernel/heap.zig
SamTebbs33: Sure. This name was copied straight from std lib.

src/kernel/filesystem/vfs.zig
DrDeano: Ah, ok, I thought it was the other way, so I will have to re-order some of mine then

SamTebbs33: Yeah me too! It actually makes sense to have it reversed though.

DrDeano (approved):
:)

PR #228, opened 2020-08-25
Add user mode

### LINKED ISSUE #50: User mode
We'll need to support user mode so that some processes can be run in a less privileged state.

Depends on #35

SamTebbs33:
This patch adds user mode for tasks. Each user task is given its own VMM and its code is copied from kernel memory into its address space. This patch also stops deinitialisation of the initrd so that testing files can be loaded (I think deinitialisation should be done just before setting the root to the hard disk, once that is implemented).

I still have some tests and doc comments to add but the implementation can be reviewed in its current state.

Closes #50

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -78,10 +78,22 @@ pub fn build(b: *Builder) !void {
     var ramdisk_files_al = ArrayList([]const u8).init(b.allocator);
     defer ramdisk_files_al.deinit();
 
-    // Add some test files for the ramdisk runtime tests
     if (test_mode == .Initialisation) {
+        // Add some test files for the ramdisk runtime tests
         try ramdisk_files_al.append("test/ramdisk_test1.txt");
         try ramdisk_files_al.append("test/ramdisk_test2.txt");
+    } else if (test_mode == .Scheduler) {
+        // Add some test files for the user mode runtime tests
+        const user_program = b.addAssemble("user_program", "test/user_program.s");
+        user_program.setOutputDir(b.cache_root);
+        user_program.setTarget(target);
+        user_program.setBuildMode(build_mode);
+        user_program.strip = true;
+
+        const copy_user_program = b.addSystemCommand(&[_][]const u8{ "objcopy", "-O", "binary", "zig-cache/user_program.o", "zig-cache/user_program" });
+        copy_user_program.step.dependOn(&user_program.step);
+        try ramdisk_files_al.append("zig-cache/user_program");
+        exec.step.dependOn(&copy_user_program.step);
     }
 
     const ramdisk_step = RamdiskStep.create(b, target, ramdisk_files_al.toOwnedSlice(), ramdisk_path);
--- a/src/kernel/arch/x86/arch.zig
+++ b/src/kernel/arch/x86/arch.zig
@@ -24,6 +24,7 @@ const Serial = @import("../../serial.zig").Serial;
 const panic = @import("../../panic.zig").panic;
 const TTY = @import("../../tty.zig").TTY;
 const Keyboard = @import("../../keyboard.zig").Keyboard;
+const Task = @import("../../task.zig").Task;
 const MemProfile = mem.MemProfile;
 
 /// The type of a device.
@@ -53,8 +54,9 @@ extern var KERNEL_STACK_END: *u32;
 /// The interrupt context that is given to a interrupt handler. It contains most of the registers
 /// and the interrupt number and error code (if there is one).
 pub const CpuState = packed struct {
+    // Page directory
+    cr3: usize,
     // Extra segments
-    ss: u32,
     gs: u32,
     fs: u32,
     es: u32,
@@ -101,9 +103,6 @@ pub const VMM_MAPPER: vmm.Mapper(VmmPayload) = vmm.Mapper(VmmPayload){ .mapFn = 
 
 /// The size of each allocatable block of memory, normally set to the page size.
 pub const MEMORY_BLOCK_SIZE: usize = paging.PAGE_SIZE_4KB;
-
-/// The default stack size of a task. Currently this is set to a page size.
-pub const STACK_SIZE: u32 = MEMORY_BLOCK_SIZE / @sizeOf(u32);
 
 ///
 /// Assembly that reads data from a given port and returns its value.
@@ -499,50 +498,71 @@ pub fn initKeyboard(allocator: *Allocator) Allocator.Error!*Keyboard {
 }
 
 ///
-/// Initialise a 32bit kernel stack used for creating a task.
+/// Initialise a stack used for creating a task.
 /// Currently only support fn () noreturn functions for the entry point.
 ///
 /// Arguments:
+///     IN task: *Task           - The task to be initialised. The function will only modify whatever
+///                                is required by the architecture. In the case of x86, it will put
+///                                the initial CpuState on the kernel stack.
 ///     IN entry_point: usize    - The pointer to the entry point of the function. Functions only
 ///                                supported is fn () noreturn
 ///     IN allocator: *Allocator - The allocator use for allocating a stack.
 ///
-/// Return: struct { stack: []u32, pointer: usize }
-///     The stack and stack pointer with the stack initialised as a 32bit kernel stack.
-///
 /// Error: Allocator.Error
 ///     OutOfMemory - Unable to allocate space for the stack.
 ///
-pub fn initTaskStack(entry_point: usize, allocator: *Allocator) Allocator.Error!struct { stack: []u32, pointer: usize } {
+pub fn initTask(task: *Task, entry_point: usize, allocator: *Allocator) Allocator.Error!void {
+    const data_offset = if (task.kernel) gdt.KERNEL_DATA_OFFSET else gdt.USER_DATA_OFFSET | 0b11;
+    // Setting the bottom two bits of the code offset designates that this is a ring 3 task
+    const code_offset = if (task.kernel) gdt.KERNEL_CODE_OFFSET else gdt.USER_CODE_OFFSET | 0b11;
+    // Ring switches push and pop two extra values on interrupt: user_esp and user_ss
+    const kernel_stack_bottom = if (task.kernel) task.kernel_stack.len - 18 else task.kernel_stack.len - 20;
+
+    var stack = &task.kernel_stack;
+
     // TODO Will need to add the exit point
     // Set up everything as a kernel task
-    var stack = try allocator.alloc(u32, STACK_SIZE);
-    stack[STACK_SIZE - 18] = gdt.KERNEL_DATA_OFFSET; // ss
-    stack[STACK_SIZE - 17] = gdt.KERNEL_DATA_OFFSET; // gs
-    stack[STACK_SIZE - 16] = gdt.KERNEL_DATA_OFFSET; // fs
-    stack[STACK_SIZE - 15] = gdt.KERNEL_DATA_OFFSET; // es
-    stack[STACK_SIZE - 14] = gdt.KERNEL_DATA_OFFSET; // ds
+    stack.*[kernel_stack_bottom] = mem.virtToPhys(@ptrToInt(&paging.kernel_directory));
+    stack.*[kernel_stack_bottom + 1] = data_offset; // gs
+    stack.*[kernel_stack_bottom + 2] = data_offset; // fs
+    stack.*[kernel_stack_bottom + 3] = data_offset; // es
+    stack.*[kernel_stack_bottom + 4] = data_offset; // ds
 
-    stack[STACK_SIZE - 13] = 0; // edi
-    stack[STACK_SIZE - 12] = 0; // esi
+    stack.*[kernel_stack_bottom + 5] = 0; // edi
+    stack.*[kernel_stack_bottom + 6] = 0; // esi
     // End of the stack
-    stack[STACK_SIZE - 11] = @ptrToInt(&stack[STACK_SIZE - 1]); // ebp
-    stack[STACK_SIZE - 10] = 0; // esp (temp) this won't be popped by popa bc intel is dump XD
+    stack.*[kernel_stack_bottom + 7] = @ptrToInt(&stack.*[stack.len - 1]); // ebp
+    stack.*[kernel_stack_bottom + 8] = 0; // esp (temp) this won't be popped by popa bc intel is dump XD
 
-    stack[STACK_SIZE - 9] = 0; // ebx
-    stack[STACK_SIZE - 8] = 0; // edx
-    stack[STACK_SIZE - 7] = 0; // ecx
-    stack[STACK_SIZE - 6] = 0; // eax
+    stack.*[kernel_stack_bottom + 9] = 0; // ebx
+    stack.*[kernel_stack_bottom + 10] = 0; // edx
+    stack.*[kernel_stack_bottom + 11] = 0; // ecx
+    stack.*[kernel_stack_bottom + 12] = 0; // eax
 
-    stack[STACK_SIZE - 5] = 0; // int_num
-    stack[STACK_SIZE - 4] = 0; // error_code
+    stack.*[kernel_stack_bottom + 13] = 0; // int_num
+    stack.*[kernel_stack_bottom + 14] = 0; // error_code
 
-    stack[STACK_SIZE - 3] = entry_point; // eip
-    stack[STACK_SIZE - 2] = gdt.KERNEL_CODE_OFFSET; // cs
-    stack[STACK_SIZE - 1] = 0x202; // eflags
+    stack.*[kernel_stack_bottom + 15] = entry_point; // eip
+    stack.*[kernel_stack_bottom + 16] = code_offset; // cs
+    stack.*[kernel_stack_bottom + 17] = 0x202; // eflags
 
-    const ret = .{ .stack = stack, .pointer = @ptrToInt(&stack[STACK_SIZE - 18]) };
-    return ret;
+    if (!task.kernel) {
+        // Put the extra values on the kernel stack needed when chaning privilege levels
+        stack.*[kernel_stack_bottom + 18] = @ptrToInt(&task.user_stack[task.user_stack.len - 1]); // user_esp
+        stack.*[kernel_stack_bottom + 19] = data_offset; // user_ss
+
+        if (!builtin.is_test) {
+            // Create a new page directory for the user task by mirroring the kernel directory
+            // We need kernel mem mapped so we don't get a page fault when entering kernel code from an interrupt
+            task.vmm.payload = &(try allocator.allocAdvanced(paging.Directory, paging.PAGE_SIZE_4KB, 1, .exact))[0];
+            task.vmm.payload.* = paging.kernel_directory.copy();
+            stack.*[kernel_stack_bottom] = vmm.kernel_vmm.virtToPhys(@ptrToInt(task.vmm.payload)) catch |e| {
+                panic(@errorReturnTrace(), "Failed to get the physical address of the user task's page directory: {}\n", .{e});
+            };
+        }
+    }
+    task.stack_pointer = @ptrToInt(&stack.*[kernel_stack_bottom]);
 }
 
 pub fn getDevices(allocator: *Allocator) Allocator.Error![]Device {
@@ -575,6 +595,19 @@ pub fn init(mem_profile: *const MemProfile) void {
     // arch init test messages
     vga.init();
     tty.init();
+}
+
+///
+/// Check the state of the user task used for runtime testing for the expected values. These should mirror those in test/user_program.s
+///
+/// Arguments:
+///     IN ctx: *const CpuState - The task's saved state
+///
+/// Return: bool
+///     True if the expected values were found, else false
+///
+pub fn runtimeTestCheckUserTaskState(ctx: *const CpuState) bool {
+    return ctx.eax == 0xCAFE and ctx.ebx == 0xBEEF;
 }
 
 test "" {
--- a/src/kernel/arch/x86/interrupts.zig
+++ b/src/kernel/arch/x86/interrupts.zig
@@ -32,7 +32,8 @@ export fn commonStub() callconv(.Naked) void {
         \\push  %%es
         \\push  %%fs
         \\push  %%gs
-        \\push  %%ss
+        \\mov %%cr3, %%eax
+        \\push %%eax
         \\mov   $0x10, %%ax
         \\mov   %%ax, %%ds
         \\mov   %%ax, %%es
@@ -42,13 +43,30 @@ export fn commonStub() callconv(.Naked) void {
         \\push  %%eax
         \\call  handler
         \\mov   %%eax, %%esp
-        \\pop   %%ss
+    );
+
+    // Pop off the new cr3 then check if it's the same as the previous cr3
+    // If so don't change cr3 to avoid a TLB flush
+    asm volatile (
+        \\pop   %%eax
+        \\mov   %%cr3, %%ebx
+        \\cmp   %%eax, %%ebx
+        \\je    same_cr3
+        \\mov   %%eax, %%cr3
+        \\same_cr3:
         \\pop   %%gs
         \\pop   %%fs
         \\pop   %%es
         \\pop   %%ds
         \\popa
-        \\add   $0x8, %%esp
+    );
+    // The Tss.esp0 value is the stack pointer used when an interrupt occurs. This should be the current process' stack pointer
+    // So skip the rest of the CpuState, set Tss.esp0 then un-skip the last few fields of the CpuState
+    asm volatile (
+        \\add   $0x1C, %%esp
+        \\.extern main_tss_entry
+        \\mov   %%esp, (main_tss_entry + 4)
+        \\sub   $0x14, %%esp
         \\iret
     );
 }
--- a/src/kernel/arch/x86/paging.zig
+++ b/src/kernel/arch/x86/paging.zig
@@ -42,6 +42,19 @@ pub const Directory = packed struct {
             }
         }
     }
+
+    ///
+    /// Copy the page directory. Changes to one copy will not affect the other
+    ///
+    /// Arguments:
+    ///     IN self: *const Directory - The directory to copy
+    ///
+    /// Return: Directory
+    ///     The copy
+    ///
+    pub fn copy(self: *const Directory) Directory {
+        return self.*;
+    }
 };
 
 /// An array of table entries. Forms the second level of paging and covers a 4MB memory space.
@@ -596,6 +609,26 @@ test "map and unmap" {
         const table = dir.tables[entry_idx] orelse unreachable;
         checkDirEntry(entry, virt, virt + PAGE_SIZE_4MB, phys, attrs, table, false);
     }
+}
+
+test "copy" {
+    // Create a dummy page dir
+    var dir: Directory = Directory{ .entries = [_]DirectoryEntry{0} ** ENTRIES_PER_DIRECTORY, .tables = [_]?*Table{null} ** ENTRIES_PER_DIRECTORY, .allocator = std.testing.allocator };
+    dir.entries[0] = 123;
+    dir.entries[56] = 794;
+    var table0 = Table{ .entries = [_]TableEntry{654} ** ENTRIES_PER_TABLE };
+    var table56 = Table{ .entries = [_]TableEntry{987} ** ENTRIES_PER_TABLE };
+    dir.tables[0] = &table0;
+    dir.tables[56] = &table56;
+    var dir2 = dir.copy();
+    const dir_slice = @ptrCast([*]const u8, &dir)[0..@sizeOf(Directory)];
+    const dir2_slice = @ptrCast([*]const u8, &dir2)[0..@sizeOf(Directory)];
+    testing.expectEqualSlices(u8, dir_slice, dir2_slice);
+
+    // Changes to one should not affect the other
+    dir2.tables[1] = &table0;
+    dir.tables[0] = &table56;
+    testing.expect(!std.mem.eql(u8, dir_slice, dir2_slice));
 }
 
 // The labels to jump to after attempting to cause a page fault. This is needed as we don't want to cause an
--- a/src/kernel/kmain.zig
+++ b/src/kernel/kmain.zig
@@ -83,6 +83,12 @@ export fn kmain(boot_payload: arch.BootPayload) void {
     arch.init(&mem_profile);
     logger.info("Arch init done\n", .{});
 
+    // The VMM runtime tests can't happen until the architecture has initialised itself
+    switch (build_options.test_mode) {
+        .Initialisation => vmm.runtimeTests(arch.VmmPayload, kernel_vmm, &mem_profile),
+        else => {},
+    }
+
     // Give the kernel heap 10% of the available memory. This can be fine-tuned as time goes on.
     var heap_size = mem_profile.mem_kb / 10 * 1024;
     // The heap size must be a power of two so find the power of two smaller than or equal to the heap_size
@@ -101,10 +107,6 @@ export fn kmain(boot_payload: arch.BootPayload) void {
         keyboard.addKeyboard(kb) catch |e| panic_root.panic(@errorReturnTrace(), "Failed to add architecture keyboard: {}\n", .{e});
     }
 
-    scheduler.init(&kernel_heap.allocator) catch |e| {
-        panic_root.panic(@errorReturnTrace(), "Failed to initialise scheduler: {}\n", .{e});
-    };
-
     // Get the ramdisk module
     const rd_module = for (mem_profile.modules) |module| {
         if (std.mem.eql(u8, module.name, "initrd.ramdisk")) {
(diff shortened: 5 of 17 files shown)


src/kernel/arch/x86/paging.zig
```diff
@@ -380,6 +384,12 @@ pub fn unmap(virtual_start: usize, virtual_end: usize, dir: *Directory) (Allocat
     }
 }
 
+pub fn mirrorDir(src: *Directory, allocator: *Allocator) Allocator.Error!*Directory {
+    var dir = try allocator.create(Directory);
+    @memcpy(@ptrCast([*]u8, dir), @ptrCast([*]u8, src), @sizeOf(Directory));
+    return dir;
+}
+
```
SamTebbs33: This has been superseded by `copy`.

src/kernel/arch/x86/pit.zig
```diff
@@ -372,8 +372,6 @@ pub fn init() void {
         panic(@errorReturnTrace(), "Invalid frequency: {}\n", .{freq});
     };
 
-    log.debug("Set frequency at: {}Hz, real frequency: {}Hz\n", .{ freq, getFrequency() });
-
```
SamTebbs33: This was removed by mistake.

SamTebbs33:
Bump @DrDeano

build.zig
```diff
@@ -71,10 +71,22 @@ pub fn build(b: *Builder) !void {
...
         try ramdisk_files_al.append("test/ramdisk_test2.txt");
+    } else if (test_mode == .Scheduler) {
+        // Add some test files for the user mode runtime tests
+        const user_program = b.addAssemble("user_program", "test/user_program.s");
+        user_program.setOutputDir(b.cache_root);
+        user_program.setTarget(target);
+        user_program.setBuildMode(build_mode);
+        user_program.strip = true;
+
+        const copy_user_program = b.addSystemCommand(&[_][]const u8{ "objcopy", "-O", "binary", "zig-cache/user_program.o", "test/user_program" });
```
DrDeano: You can use NASM:
```
const user_program = b.addSystemCommand(&[_][]const u8{ "nasm", user_program_src, "-o", user_program_bin });
```
So all in one line. What do you think?

src/kernel/arch/x86/interrupts.zig
```diff
@@ -42,13 +43,30 @@ export fn commonStub() callconv(.Naked) void {
...
         \\popa
-        \\add   $0x8, %%esp
+    );
+    // The Tss.esp0 value is the stack pointer used when an interrupt occurs. This should be the current process' stack pointer
+    // So skip the rest of the CpuState, set Tss.esp0 then un-skip the last few fields of the CpuState
+    asm volatile (
+        \\add   $0x1C, %%esp
+        \\.extern main_tss_entry
+        \\mov %%esp, (main_tss_entry + 4)
+        \\sub $0x14, %%esp
```
DrDeano: Could align the ASM to match the other ASM

src/kernel/arch/x86/paging.zig
```diff
@@ -579,6 +583,21 @@ test "map and unmap" {
...
+    dir.entries[56] = 794;
+    var table0 = Table{ .entries = [_]TableEntry{654} ** ENTRIES_PER_TABLE };
+    var table56 = Table{ .entries = [_]TableEntry{987} ** ENTRIES_PER_TABLE };
+    dir.tables[0] = &table0;
+    dir.tables[56] = &table56;
+    const dir2 = dir.copy();
+    const dir_slice = @ptrCast([*]const u8, &dir)[0..@sizeOf(Directory)];
+    const dir2_slice = @ptrCast([*]const u8, &dir2)[0..@sizeOf(Directory)];
+    testing.expectEqualSlices(u8, dir_slice, dir2_slice);
+}
```
DrDeano: Could add if you change `dir` and does it change `dir2`

test/mock/kernel/vmm_mock.zig
```diff
@@ -45,9 +46,11 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
         pub fn free(self: *Self, vaddr: usize) (bitmap.Bitmap(u32).BitmapError || VmmError)!void {
             return VmmError.NotAllocated;
         }
+
+        pub fn copyDataToVMM(self: *Self, to: *const Self, data: []const u8, dest: usize) (bitmap.Bitmap(usize).BitmapError || VmmError || Allocator.Error)!void {}
     };
 }
 
-pub fn init(mem_profile: *const mem.MemProfile, allocator: *std.mem.Allocator) std.mem.Allocator.Error!VirtualMemoryManager(arch.VmmPayload) {
+pub fn init(mem_profile: *const mem.MemProfile, allocator: *std.mem.Allocator) std.mem.Allocator.Error!*VirtualMemoryManager(arch.VmmPayload) {
```
DrDeano: Can replace `std.mem.Allocator` with `Allocator` as you added it at the top

src/kernel/kmain.zig
```diff
@@ -126,25 +121,27 @@ export fn kmain(boot_payload: arch.BootPayload) void {
...
     logger.info("Init\n", .{});
 
     // Main initialisation finished so can enable interrupts
-    arch.enableInterrupts();
 
     logger.info("Creating init2\n", .{});
 
     // Create a init2 task
-    var idle_task = task.Task.create(initStage2, &kernel_heap.allocator) catch |e| {
+    var idle_task = task.Task.create(@ptrToInt(initStage2), true, kernel_vmm, &kernel_heap.allocator) catch |e| {
```
DrDeano: Have noticed, this isn't a idle task :P

src/kernel/kmain.zig
```diff
@@ -126,25 +121,27 @@ export fn kmain(boot_payload: arch.BootPayload) void {
...
         panic_root.panic(@errorReturnTrace(), "Failed to create init stage 2 task: {}\n", .{e});
     };
     scheduler.scheduleTask(idle_task, &kernel_heap.allocator) catch |e| {
         panic_root.panic(@errorReturnTrace(), "Failed to schedule init stage 2 task: {}\n", .{e});
     };
+    arch.enableInterrupts();
```
DrDeano: Also, how come this was moved lower? As the runtime test for the scheduler won't run until this point.

src/kernel/scheduler.zig
```diff
@@ -348,14 +355,47 @@ fn rt_variable_preserved(allocator: *Allocator) void {
     log.info("SUCCESS: Scheduler variables preserved\n", .{});
 }
 
+fn rt_user_task(allocator: *Allocator, mem_profile: *const mem.MemProfile) void {
```
DrDeano, line 372: Doc comment

test/mock/kernel/task_mock.zig
```diff
@@ -8,16 +10,19 @@ pub const addTestParams = mock_framework.addTestParams;
...
     pid: u32,
-    stack: []u32,
+    kernel_stack: []u32,
+    user_stack: []u32,
     stack_pointer: usize,
+    kernel: bool,
+    vmm: vmm.VirtualMemoryManager(arch.VmmPayload),
 
-    pub fn create(entry_point: EntryPointFn, allocator: *Allocator) Allocator.Error!*Task {
+    pub fn create(entry_point: EntryPointFn, kernel: bool, task_vmm: *vmm.VirtualMemoryManager(arch.VmmPayload), allocator: *Allocator) Allocator.Error!*Task {
```
DrDeano: You will need to add the new parameters to the `performAction` function.

src/kernel/scheduler.zig
```diff
@@ -83,6 +86,7 @@ pub fn pickNextTask(ctx: *arch.CpuState) usize {
         current_task = next_task;
     }
 
+    const str: []const u8 = if (current_task.kernel) "kernel" else "user";
```
DrDeano: Doesn't look like this is used

src/kernel/scheduler.zig
```diff
@@ -348,14 +355,47 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+    };
+    task_vmm.* = vmm.VirtualMemoryManager(arch.VmmPayload).init(0, @ptrToInt(mem_profile.vaddr_start), allocator, arch.VMM_MAPPER, undefined) catch unreachable;
+    // 2. Create user task. The code will be loaded at address 0
+    var user_task = task.Task.create(0, false, task_vmm, allocator) catch |e| {
+        panic(@errorReturnTrace(), "Failed to create user task: {}\n", .{e});
+    };
+    // 3. Read the user program file from the filesystem
+    const user_program_file = fs.openFile("/user_program", .NO_CREATION) catch |e| {
+        panic(@errorReturnTrace(), "Failed to open /user_program: {}\n", .{e});
+    };
```
DrDeano, line 385: `defer user_program_file.close();`?

src/kernel/scheduler.zig
```diff
@@ -348,14 +355,47 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+    const code = user_program_file.read(1024) catch |e| {
+        panic(@errorReturnTrace(), "Failed to read user program file: {}\n", .{e});
+    };
+    defer allocator.free(code);
+    // 4. Allocate space in the vmm for the user_program
+    const code_start = task_vmm.alloc(std.mem.alignForward(code.len, vmm.BLOCK_SIZE) / vmm.BLOCK_SIZE, .{ .kernel = false, .writable = true, .cachable = true }) catch |e| {
```
DrDeano: Not sure code should be writable.

src/kernel/scheduler.zig
```diff
@@ -348,14 +355,47 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+        panic(@errorReturnTrace(), "Failed to allocate VMM memory for user program code: {}\n", .{e});
+    } orelse unreachable;
+    if (code_start != 0) unreachable;
```
DrDeano: I think instead of `unreachable` have kernel panics to explain where the error was and some debug info

src/kernel/scheduler.zig
```diff
@@ -348,14 +355,47 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+    // 5. Copy user_program code over
+    vmm.kernel_vmm.copyDataToVMM(task_vmm, code, code_start) catch |e| {
+        panic(@errorReturnTrace(), "Failed to copy user code: {}\n", .{e});
+    };
+    // 6. Schedule it
+    scheduleTask(user_task, allocator) catch unreachable;
+    // 7. Profit!(?)
```
DrDeano: As you are writing 0xCAFE and 0xBEEF to the registers, in the scheduler check the stack for these values. Might need to set up an additional runtime option for `user_program1` for this, or a within the `scheduler` runtime option, have a bool for testing this case.

src/kernel/scheduler.zig
```diff
@@ -308,7 +315,7 @@ fn rt_variable_preserved(allocator: *Allocator) void {
     defer allocator.destroy(is_set);
     is_set.* = true;
 
-    var test_task = Task.create(task_function, allocator) catch unreachable;
+    var test_task = Task.create(@ptrToInt(task_function), true, undefined, allocator) catch unreachable;
```
DrDeano, line 325: Could add some tests that create user tasks

src/kernel/task.zig
```diff
@@ -8,14 +8,15 @@ const mock_path = build_options.mock_path;
...
+const vmm = @import("vmm.zig");
 const Allocator = std.mem.Allocator;
 
 /// The kernels main stack start as this is used to check for if the task being destroyed is this stack
 /// as we cannot deallocate this.
 extern var KERNEL_STACK_START: *u32;
 
 /// The function type for the entry point.
-const EntryPointFn = fn () void;
+pub const EntryPointFn = usize;
```
DrDeano: This is used both in scheduler and here, so just need one.

src/kernel/arch/x86/paging.zig
```diff
@@ -25,6 +25,10 @@ pub const Directory = packed struct {
 
     /// The tables allocated for the directory. This is ignored by the CPU.
     tables: [ENTRIES_PER_DIRECTORY]?*Table,
+
+    pub fn copy(self: *const Directory) Allocator.Error!Directory {
```
DrDeano: Doc comment

src/kernel/task.zig
```diff
@@ -57,16 +72,31 @@ pub const Task = struct {
...
     ///                   be freed on return.
     ///
-    pub fn create(entry_point: EntryPointFn, allocator: *Allocator) Allocator.Error!*Task {
+    pub fn create(entry_point: EntryPointFn, kernel: bool, task_vmm: *vmm.VirtualMemoryManager(arch.VmmPayload), allocator: *Allocator) Allocator.Error!*Task {
         var task = try allocator.create(Task);
         errdefer allocator.destroy(task);
 
-        task.pid = allocatePid();
+        const pid = allocatePid();
         errdefer freePid(task.pid);
```
DrDeano: Will need to move the `errdefer` down below the `task.* =`... or change it to `errdefer(freePid(pid))`.

src/kernel/task.zig
```diff
@@ -57,16 +72,31 @@ pub const Task = struct {
...
-        const task_stack = try arch.initTaskStack(@ptrToInt(entry_point), allocator);
-        task.stack = task_stack.stack;
-        task.stack_pointer = task_stack.pointer;
+        var k_stack = try allocator.alloc(usize, STACK_SIZE);
+        errdefer allocator.free(k_stack);
+
+        var u_stack = if (kernel) &[_]usize{} else try allocator.alloc(usize, STACK_SIZE);
+        if (!kernel) {
+            errdefer allocator.free(u_stack);
+        }
```
DrDeano: Could be turned into: `errdefer if (!kernel) allocator.free(u_stack);`

src/kernel/task.zig
```diff
@@ -155,22 +188,22 @@ test "create expected setup" {
...
+    expectEqual(false, std.testing.allocator_instance.detectLeaks());
 
     // All PIDs were freed
     expectEqual(all_pids.bitmap, 1);
 }
 
 test "Multiple create" {
-    var task1 = try Task.create(test_fn1, std.testing.allocator);
-    var task2 = try Task.create(test_fn1, std.testing.allocator);
+    var task1 = try Task.create(@ptrToInt(test_fn1), true, undefined, std.testing.allocator);
```
DrDeano, line 212: Could create some tests for creating user tasks

src/kernel/vmm.zig
```diff
@@ -180,6 +183,18 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
             };
         }
 
+        pub fn mirror(self: *const Self) Allocator.Error!Self {
```
DrDeano: Doc comment

src/kernel/vmm.zig
```diff
@@ -360,6 +375,7 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
                         self.mapper.mapFn(vaddr, vaddr + BLOCK_SIZE, addr, addr + BLOCK_SIZE, attrs, self.allocator, self.payload) catch |e| {
                             panic(@errorReturnTrace(), "Failed to map virtual memory: {X}\n", .{e});
                         };
+                        log.debug("Mapping {X} to {X}\n", .{ vaddr, addr });
```
DrDeano: Left over debug?

src/kernel/vmm.zig
```diff
@@ -369,6 +385,82 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
+                        // Cannot be out of bounds as it has been set above
+                        self.bmp.clearEntry((v - self.start) / BLOCK_SIZE) catch unreachable;
+                    }
+                }
+                for (blocks.items) |block, i| {
+                    const v = v_start + i * BLOCK_SIZE;
+                    const v_end = v + BLOCK_SIZE;
+                    const p = block;
+                    const p_end = p + BLOCK_SIZE;
+                    log.debug("Copy: Mapping {X} to {X}\n", .{ v, p });
```
DrDeano: Left over debug?

src/kernel/vmm.zig
```diff
@@ -369,6 +385,82 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
+                                // If we can't unmap then just panic
+                                panic(@errorReturnTrace(), "Failed to unmap region 0x{X} -> 0x{X}: {}\n", .{ v_start, v_end, e2 });
+                            };
+                        }
+                        panic(@errorReturnTrace(), "Failed to map vrutal region 0x{X} -> 0x{X} to 0x{X} -> 0x{X}: {}\n", .{ v, v_end, p, p_end, e });
+                    };
+                }
+                // Copy to vaddr from above
+                const align_offset = dest - start_addr;
+                log.debug("Copying to {X}\n", .{v_start + align_offset});
```
DrDeano: Same here

src/kernel/vmm.zig
```diff
@@ -734,4 +861,47 @@ fn runtimeTests(comptime Payload: type, vmm: VirtualMemoryManager(Payload), mem_
     }
 
     log.info("Tested allocations\n", .{});
+
+    rt_copyDataToVMM(Payload, vmm);
+}
+
+fn rt_copyDataToVMM(comptime Payload: type, vmm: *VirtualMemoryManager(Payload)) void {
```
DrDeano: Doc comment

src/kernel/vmm.zig
```diff
@@ -699,7 +826,7 @@ fn testUnmap(vstart: usize, vend: usize, payload: u8) (Allocator.Error || Mapper
 ///     IN mem_profile: *const mem.MemProfile - The mem profile with details about all the memory regions that should be reserved
 ///     IN mb_info: *multiboot.multiboot_info_t - The multiboot info struct that should also be reserved
 ///
-fn runtimeTests(comptime Payload: type, vmm: VirtualMemoryManager(Payload), mem_profile: *const mem.MemProfile) void {
+fn runtimeTests(comptime Payload: type, vmm: *VirtualMemoryManager(Payload), mem_profile: *const mem.MemProfile) void {
```
DrDeano: As there are 2 runtime tests, can move the exiting test to a new function

src/kernel/vmm.zig
```diff
@@ -369,6 +385,82 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
+        ///     IN data: []const u8 - The data being copied. Must be within memory mapped in `self`
+        ///     IN dest: usize - The address within `to` to copy the data to. The space covered by `dest` and `dest` + `dest.len` must be mapped in `to`
+        ///
+        /// Error: VmmError || pmm.PmmError || Allocator.Error
+        ///     VmmError.NotAllocated - Some or all of the destination isn't mapped
+        ///     VmmError.OutOfMemory - There wasn't enough space in the VMM to use for temporary mapping
+        ///     Bitmap(u32).Error.OutOfBounds - The address given is outside of the memory managed
+        ///     Allocator.Error.OutOfMemory - There wasn't enough memory available to fulfill the request
+        ///
+        pub fn copyDataToVMM(self: *Self, to: *const Self, data: []const u8, dest: usize) (bitmap.Bitmap(usize).BitmapError || VmmError || Allocator.Error)!void {
```
DrDeano, line 438: Could this be unit tested

src/kernel/scheduler.zig
```diff
@@ -348,14 +355,47 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+}
+
 ///
 /// The scheduler runtime tests that will test the scheduling functionality.
 ///
 /// Arguments:
 ///     IN allocator: *Allocator - The allocator to use when needing to allocate memory.
 ///
-fn runtimeTests(allocator: *Allocator) void {
+fn runtimeTests(allocator: *Allocator, mem_profile: *const mem.MemProfile) void {
```
DrDeano, line 413: Need to add the additional argument to the doc comment

DrDeano (commented):
Looking good. Could try some syscalls maybe

build.zig
SamTebbs33: I'm not sure about adding dependencies on a program like NASM. Most people are guaranteed to have objcopy installed, or not have any trouble installing it, since it's a GNU tool.

src/kernel/kmain.zig
SamTebbs33: Indeed :)

src/kernel/kmain.zig
SamTebbs33: I don't think it was intended so I'll move it back.

src/kernel/scheduler.zig
SamTebbs33: Indeed, this was left over from testing.

src/kernel/scheduler.zig
SamTebbs33: It will need to be when copying the code from kernel space to user space though

src/kernel/scheduler.zig
SamTebbs33: Ah yes, thanks for reminding me about this

build.zig
DrDeano: I though objcopy would be a dependency anyway. But if not could make this this into a little helper function for compiling a flat binary asm file as I use nasm for the FAT heater (for the sub bootloader), so can use this instead.

src/kernel/scheduler.zig
DrDeano: yes, but once copied, could it be marked as not writable. Idk if we have this kinda of security in place yet?

SamTebbs33: I agree and I don't think we do. Perhaps this should come in a separate PR that makes changing the attributes of an allocation possible.

src/kernel/scheduler.zig
SamTebbs33: Since the cpu state is arch-dependent, I've added a function to arch.zig that will check the saved state for the expected values. Let me know what you think of this.

src/kernel/scheduler.zig
SamTebbs33: There is one in `rt_user_task`.

src/kernel/task.zig
SamTebbs33: Good spot!

src/kernel/task.zig
SamTebbs33: I think you're right. Like this it will errdefer at the end of the if statement which is not what we want.

src/kernel/task.zig
SamTebbs33: I've added user task testing to all existing tests.

src/kernel/vmm.zig
SamTebbs33: Maybe :)

src/kernel/scheduler.zig
DrDeano: Yes, I think it can come under loading an ELF file when setting up the section permitions

src/kernel/scheduler.zig
DrDeano: Yes, but unit test coverage :P And unit tests are faster to run

src/kernel/scheduler.zig
SamTebbs33: Good idea

src/kernel/scheduler.zig
SamTebbs33: Ah you wrote this comment in one of the rt functions. I added unit tests in a commit after your review :+1:

src/kernel/vmm.zig
SamTebbs33: We won't be able to test the functionality itself, only the function's clean-up and such. I'll give it a go.

build.zig
SamTebbs33: Yeah, I think objcopy is a better dependency to have, if we're choosing between objcopy and nasm, since most linux distributions and even WSL should have it installed by default, and those that don't shouldn't have any issue since it's a GNU tool. I can rebase on top of your fat32 builder patch and make a nice helper function :+1:

DrDeano: Rebase? I assume this will be merged before my fat32 builder

src/kernel/vmm.zig
DrDeano: Sure, Also if possible try to reach all the Error.

build.zig
SamTebbs33: I doubt it, I'm having real trouble figuring out why the copyDataToVmm rt-test isn't working

DrDeano: Also, I have seen you can use `addInstallRaw` like in #240

SamTebbs33: Yep, that's much better :)

SamTebbs33: I tried to use `addInstallRaw` but an `.Obj` step (which is what's returned by `addAssemble()`) isn't supported atm. I'd like to use `addInstallRaw` once that is supported.

DrDeano: No prob, what you have is good

SamTebbs33:
> Looking good. Could try some syscalls maybe

I thought it would make the PR a bit easier to manage if we add syscalls in later PRs.

src/kernel/kmain.zig
```diff
@@ -83,6 +83,12 @@ export fn kmain(boot_payload: arch.BootPayload) void {
     arch.init(&mem_profile);
     logger.info("Arch init done\n", .{});
 
+    // The VMM runtime tests can't happen until the architecture has initialised itself
+    switch (build_options.test_mode) {
+        .Initialisation => vmm.runtimeTests(arch.VmmPayload, kernel_vmm, &mem_profile),
+        else => {},
+    }
+
```
DrDeano, line 91: Can't the `arch.init()` be moved up to above the `pmm.init()`?

src/kernel/kmain.zig
```diff
@@ -129,10 +130,12 @@ export fn kmain(boot_payload: arch.BootPayload) void {
...
         // Need to init the vfs after the ramdisk as we need the root node from the ramdisk filesystem
         vfs.setRoot(ramdisk_filesystem.root_node);
-
-        // Load all files here
     }
 
+    scheduler.init(&kernel_heap.allocator, &mem_profile) catch |e| {
+        panic_root.panic(@errorReturnTrace(), "Failed to initialise scheduler: {}\n", .{e});
+    };
+
```
DrDeano, line 138: How come the scheduler was moved down, I don't think it makes a difference, am just curious.

src/kernel/kmain.zig
```diff
@@ -120,7 +122,6 @@ export fn kmain(boot_payload: arch.BootPayload) void {
         var ramdisk_filesystem = initrd.InitrdFS.init(&initrd_stream, &kernel_heap.allocator) catch |e| {
             panic_root.panic(@errorReturnTrace(), "Failed to initialise ramdisk: {}\n", .{e});
         };
-        defer ramdisk_filesystem.deinit();
```
DrDeano, line 123: How come the `deinit` is removed?

src/kernel/scheduler.zig
```diff
@@ -154,20 +166,20 @@ fn test_fn2() void {}
 
 var test_pid_counter: u7 = 1;
 
-fn task_create(entry_point: EntryPointFn, allocator: *Allocator) Allocator.Error!*Task {
+fn task_create(entry_point: EntryPoint, allocator: *Allocator, kernel: bool, task_vmm: *vmm.VirtualMemoryManager(u8)) Allocator.Error!*Task {
```
DrDeano: Have noticed that I forgot the correct function case, so should be tastCreate(). So this can be updated. Same for task_destroy.

src/kernel/scheduler.zig
```diff
@@ -348,14 +360,55 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+    const code_start = task_vmm.alloc(std.mem.alignForward(code_len, vmm.BLOCK_SIZE) / vmm.BLOCK_SIZE, .{ .kernel = false, .writable = true, .cachable = true }) catch |e| {
+        panic(@errorReturnTrace(), "Failed to allocate VMM memory for user program code: {}\n", .{e});
+    } orelse panic(null, "User task VMM didn't allocate space for the user program\n", .{});
+    if (code_start != 0) panic(null, "User program start address was {} instead of 0\n", .{code_start});
+    // 5. Copy user_program code over
+    vmm.kernel_vmm.copyDataToVMM(task_vmm, code[0..code_len], code_start) catch |e| {
+        panic(@errorReturnTrace(), "Failed to copy user code: {}\n", .{e});
+    };
+    // 6. Schedule it
+    scheduleTask(user_task, allocator) catch unreachable;
```
DrDeano: Could also add a panic for failing to schedule the task.

src/kernel/scheduler.zig
```diff
@@ -129,17 +139,19 @@ pub fn init(allocator: *Allocator) Allocator.Error!void {
     errdefer allocator.destroy(current_task);
     // PID 0
     current_task.pid = 0;
-    current_task.stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..4096];
+    current_task.kernel_stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..4096];
```
DrDeano: Could try and use `KERNEL_STACK_END` in case we change the stack size later for the length.

src/kernel/task.zig
```diff
@@ -145,32 +178,42 @@ test "create out of memory for stack" {
...
+    var task = try Task.create(@ptrToInt(test_fn1), true, undefined, allocator);
+    var user_task = try Task.create(@ptrToInt(test_fn1), false, undefined, allocator);
 
-    task.destroy(&fa.allocator);
+    task.destroy(allocator);
+    user_task.destroy(allocator);
 
     // Make sure any memory allocated is freed
-    expectEqual(fa.allocated_bytes, fa.freed_bytes);
+    expectEqual(false, std.testing.allocator_instance.detectLeaks());
```
DrDeano: Can remove this line as the testing allocator does this check for us

src/kernel/arch/x86/arch.zig
```diff
@@ -499,50 +498,71 @@ pub fn initKeyboard(allocator: *Allocator) Allocator.Error!*Keyboard {
...
+
+    if (!task.kernel) {
+        // Put the extra values on the kernel stack needed when chaning privilege levels
+        stack.*[kernel_stack_bottom + 18] = @ptrToInt(&task.user_stack[task.user_stack.len - 1]); // user_esp
+        stack.*[kernel_stack_bottom + 19] = data_offset; // user_ss
+
+        if (!builtin.is_test) {
+            // Create a new page directory for the user task by mirroring the kernel directory
+            // We need kernel mem mapped so we don't get a page fault when entering kernel code from an interrupt
+            task.vmm.payload = &(try allocator.allocAdvanced(paging.Directory, 4096, 1, .exact))[0];
```
DrDeano: Is there a constant for the 4096 bytes to allocate for a page

src/kernel/vmm.zig
```diff
@@ -627,7 +792,8 @@ const test_vaddr_start: usize = 0xC0000000;
 ///
 fn testInit(num_entries: u32) Allocator.Error!VirtualMemoryManager(u8) {
     if (test_allocations == null) {
-        test_allocations = try bitmap.Bitmap(u64).init(num_entries, std.heap.page_allocator);
+        test_allocations = try std.heap.page_allocator.create(bitmap.Bitmap(u64));
+        test_allocations.?.* = try bitmap.Bitmap(u64).init(num_entries, std.heap.page_allocator);
```
DrDeano: Could use the `std.testing.allocator` instead

src/kernel/vmm.zig
```diff
@@ -699,7 +869,7 @@ fn testUnmap(vstart: usize, vend: usize, payload: u8) (Allocator.Error || Mapper
 ///     IN mem_profile: *const mem.MemProfile - The mem profile with details about all the memory regions that should be reserved
 ///     IN mb_info: *multiboot.multiboot_info_t - The multiboot info struct that should also be reserved
 ///
-fn runtimeTests(comptime Payload: type, vmm: VirtualMemoryManager(Payload), mem_profile: *const mem.MemProfile) void {
+pub fn runtimeTests(comptime Payload: type, vmm: *VirtualMemoryManager(Payload), mem_profile: *const mem.MemProfile) void {
```
DrDeano, line 918: I think it would be nice to move the existing test to a new function

src/kernel/kmain.zig
SamTebbs33: The thing is that the arch initialisation depends on the vmm being set up.

src/kernel/kmain.zig
SamTebbs33: That's because the scheduler rt tests use a file from the ramdisk so need to be run after that is set up.

src/kernel/kmain.zig
SamTebbs33: Since if the ramdisk was deinitialised then the scheduler wouldn't be able to read the user program from the vfs. I think we discussed privately about leaving the ramdisk in memory until we initialise a FAT system and change the root (at which point we won't need the ramdisk).

src/kernel/scheduler.zig
SamTebbs33: Good idea.

src/kernel/arch/x86/arch.zig
SamTebbs33: Yep, this is the alignment so I can change it to PAGE_SIZE_4KB.

src/kernel/vmm.zig
SamTebbs33: _Should_ be using that instead! :)

src/kernel/vmm.zig
SamTebbs33: Sure thing.

src/kernel/scheduler.zig
SamTebbs33: I've changed it to `createTestTask` and `destroyTestTask` to differentiate them from other functions, what do you think?

DrDeano: Sound good

DrDeano (approved):
Bam, almost there to a usable OS :)

SamTebbs33:
> Bam, almost there to a usable OS :)

POGGERS

Issue #216, opened 2020-08-01
Add the FAT32 filesystem


DrDeano:
A good starting point for a filesystem is the FAT32. Even though this is old, it is simple (tm) to implement. This will use the device interface so can abstract what the FAT32 is sitting on, hard drive, ram, ...
We would also like for this to create empty FAT32 images as creating these are difficult via scripts (unless someone can find one that will play nicely). This maybe part of a separate issue.

PR #238, opened 2020-09-22
Create a blank FAT32 image

DrDeano:
This will be used for testing the FAT32 driver for the kernel and will be integrated into the OS as a mkfs.fat32 program.
Plus typos

Part of #216

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const logger = std.log.scoped(.builder);
 const builtin = @import("builtin");
 const rt = @import("test/runtime_test.zig");
 const RuntimeStep = rt.RuntimeStep;
@@ -12,6 +13,7 @@ const File = fs.File;
 const Mode = builtin.Mode;
 const TestMode = rt.TestMode;
 const ArrayList = std.ArrayList;
+const Fat32 = @import("mkfat32.zig").Fat32;
 
 const x86_i686 = CrossTarget{
     .cpu_arch = .i386,
@@ -28,6 +30,7 @@ pub fn build(b: *Builder) !void {
 
     const fmt_step = b.addFmt(&[_][]const u8{
         "build.zig",
+        "mkfat32.zig",
         "src",
         "test",
     });
@@ -42,6 +45,7 @@ pub fn build(b: *Builder) !void {
     const boot_path = try fs.path.join(b.allocator, &[_][]const u8{ b.exe_dir, "iso", "boot" });
     const modules_path = try fs.path.join(b.allocator, &[_][]const u8{ b.exe_dir, "iso", "modules" });
     const ramdisk_path = try fs.path.join(b.allocator, &[_][]const u8{ b.install_path, "initrd.ramdisk" });
+    const fat32_image_path = try fs.path.join(b.allocator, &[_][]const u8{ b.install_path, "fat32.img" });
 
     const build_mode = b.standardReleaseOptions();
     comptime var test_mode_desc: []const u8 = "\n                         ";
@@ -67,6 +71,9 @@ pub fn build(b: *Builder) !void {
         else => unreachable,
     };
     make_iso.step.dependOn(&exec.step);
+
+    var fat32_builder_step = Fat32BuilderStep.create(b, .{}, fat32_image_path);
+    make_iso.step.dependOn(&fat32_builder_step.step);
 
     var ramdisk_files_al = ArrayList([]const u8).init(b.allocator);
     defer ramdisk_files_al.deinit();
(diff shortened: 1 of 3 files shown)


build.zig
```diff
@@ -28,20 +30,24 @@ pub fn build(b: *Builder) !void {
...
         "build.zig",
+        "mkfat32.zig",
         "src",
         "test",
     });
     b.default_step.dependOn(&fmt_step.step);
 
     const main_src = "src/kernel/kmain.zig";
     const arch_root = "src/kernel/arch";
+    const fat32_bootstrap_src = "fat32_bootstrap.asm";
```
SamTebbs33: What's the purpose of the fat32_bootstrap?

build.zig
```diff
@@ -166,6 +178,226 @@ pub fn build(b: *Builder) !void {
...
+    builder: *Builder,
+
+    /// The file path to the bootstrap compiled code to be used in the FAT32 header.
+    fat32_bootstrap_bin: []const u8,
+
+    /// The path to where the ramdisk will be written to.
+    out_file_path: []const u8,
+
+    /// Options for creating the FAt32 image.
+    fat32_options: Options,
```
SamTebbs33: I think the `fat32_` prefixes here is unnecessary.

build.zig
```diff
@@ -166,6 +178,226 @@ pub fn build(b: *Builder) !void {
...
+            /// A user provided value. This can be a unsigned integer or a string.
+            value: union {
+                int: usize,
+                str: []const u8,
+            },
+        };
+
+        /// The FAT32 image size in bytes. This is not the formatted size. Default see
+        /// `Fat32.getDefaultImageSize`
+        image_size: DefaultOrValue = .default,
```
SamTebbs33: Instead of setting these to `.default`, could they just be set to the default value? Feel free to explain the reason why there's an extra level of indirection.

mkfat32.zig
```diff
@@ -0,0 +1,361 @@
...
+        minute: u8,
+        second: u8,
+        millisecond_10: u8,
+    };
+
+    /// The FAT32 boot sector header (without the jmp and boot code) including the extended
+    /// parameter block for FAT32 with the extended signature 0x29.
+    pub const Header = struct {
+        /// The OEM.
+        oem: [8]u8 = "ZystemOS".*,
```
SamTebbs33, line 60: It's almost as if we planned the name to be the exact length needed :D

mkfat32.zig
```diff
@@ -0,0 +1,361 @@
...
+    /// Error File.WriteError || File.SeekError
+    ///     File.WriteError - If there is an error when writing to the image. See File.WriteError
+    ///     File.SeekError  - If there is an error when seeking the image. See File.SeekError
+    ///
+    pub fn writeFSInfo(image: File, fat32_header: Header, free_cluster_num: u32, next_free_cluster: u32) (File.WriteError || File.SeekError)!void {
+        // Seek to the correct location
+        try image.seekableStream().seekTo(fat32_header.fsinfo_sector * fat32_header.bytes_per_sector);
+
+        // First signature
+        try image.writer().writeIntLittle(u32, 0x41615252);
```
SamTebbs33: Can `image.writer()` be stored in a variable instead of being called every time?

mkfat32.zig
```diff
@@ -0,0 +1,361 @@
...
+    /// Create the FAT32 serial number. This is generated from the date. Reference:
+    /// https://www.digital-detective.net/documents/Volume%20Serial%20Numbers.pdf
+    ///
+    /// Arguments:
+    ///     IN current_time: SerialTime - The current date and time.
+    ///
+    /// Return: u32
+    ///     Serial number for a new FAT32 image.
+    ///
+    pub fn createSerialNumber(current_time: SerialTime) u32 {
```
SamTebbs33: This could be moved to the SerialTime struct for nice encapsulation.

SamTebbs33 (changes requested):

build.zig
DrDeano: It is part of the FAT32 header. Is it used if it is loaded as a bootloader and will just print "please insert a bootable media, press any key to restart." It is taken from my bootloader. I guess it isn't 100% needed, but it part of the specification so may not be able to mount in Linux. There could just be a random flat binary, but the code must come from somewhere.

build.zig
DrDeano: That is a good point lol. I just went with it because I kinda like it ;)

mkfat32.zig
DrDeano: I know right :P

mkfat32.zig
DrDeano: In the std lib, when multiple writes are made, `writer()` is called each time. But it didn't make a difference when storing the write in a variable. ??

mkfat32.zig
DrDeano: I don't think it should be moved as it is a static function and there isn't a need for `Self` for `SerialTime`. It would encapsulate it but would make the use of it longer: `Fat32.createSerialNumber()` =>`Fat32.SerialTime.createSerialNumber()`. Also will need to make the struct public.

SamTebbs33: It's static for now, but with my suggestion, if you have a `SerialTime` instance you'd just do `serial_time.createSerialNumber()` instead of `Fat32.createSerialNumber(serial_time)`, so it would actually be shorter!

DrDeano: There wouldn't be any other need for SerialTime other than creating the serial number. The only reason I have the SerialTime struct was to reduce the number of parameters being passed to the `createSerialNumber`. It would just create code blot and be more complex than needed. Once there is std lib get human readable time, I can have another look at changing how this is done. Probable removing the the argument for `createSerialNumber`, removing the `SerialTime` struct, and having the function call the std lib function directly.

build.zig
SamTebbs33: That makes sense. I think it would be nice to move it somewhere under arch/x86 though.

mkfat32.zig
SamTebbs33: I just think it would be a little cleaner.

mkfat32.zig
SamTebbs33: Hmm, I still think that putting the function in the struct is better than having it in Fat32, since that's how we and the std lib do things in other structs. I don't agree with the code bloat argument since it won't mean adding more lines and it would actually be cleaner! Alternatively I don't mind having 6 parameters for the function.

DrDeano: I shall do parameters

mkfat32.zig
DrDeano: ok

build.zig
DrDeano: It shouldn't go into the x86 folder as it can be used for any arch. Maybe could add the pre-build binary to the git repo and remove the build step for this, so removing the nasm dependency. And if people want to build it, they can separetly

SamTebbs33: Building the image separately makes a lot of sense 👍 Building a FAT image shouldn't be a dependency for the kernel. Although, I don't see how this can be used for any architecture since it uses x86 instructions.

SamTebbs33 (approved):
LGTM

PR #240, opened 2020-09-26
build.zig use addInstallRaw instead of objcopy

markfirmware:


SamTebbs33 (approved):
Nice, this makes it much cleaner. We can use this elsewhere too!

PR #243, opened 2020-09-29
Feature/aarch64 port improved linking

markfirmware:
I was confused about armstub - it is not used at all when using kernel_old=1 (which loads the kernel at 0.)

This simplifies build.zig.

Changes to the mmu page table so that it is not allocated at 0x50000, etc.


markfirmware:
not yet

PR #248, opened 2020-09-30
aarch64 improved linking

markfirmware:
I was confused about armstub - it is not used at all when using kernel_old=1 (which loads the kernel at 0.)

This simplifies build.zig.

Changes to the mmu page table so that it is a var and  not allocated at fixed addresses like 0x50000, etc.

@SamTebbs33

### DIFF
--- a/src/kernel/arch/aarch64/link.ld
+++ b/src/kernel/arch/aarch64/link.ld
@@ -1,8 +1,7 @@
 ENTRY(_start)
 
 SECTIONS {
-	/* The binary is loaded at 0x80000 for aarch64 and 0x8000 for arm */
-	. = 0x80000;
+	. = 0;
 	KERNEL_PHYSADDR_START = .;
 
 	.text ALIGN(4K) : {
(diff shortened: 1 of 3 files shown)


src/kernel/arch/aarch64/link.ld
```diff
@@ -1,8 +1,7 @@
 ENTRY(_start)
 
 SECTIONS {
-	/* The binary is loaded at 0x80000 for aarch64 and 0x8000 for arm */
-	. = 0x80000;
+	. = 0;
```
SamTebbs33, line 4: Is this correct? I read that the Pi loads the kernel at 0x80000.

markfirmware: config.txt kernel_old=1 will load it at  0
https://www.raspberrypi.org/documentation/configuration/config-txt/boot.md#kernel_old

SamTebbs33 (approved):
Nice cleanup. Thanks!

PR #249, opened 2020-10-03
Initial PCI interface

### LINKED ISSUE #244: Initial PCI interface
Create a basic interface for PCI: https://wiki.osdev.org/PCI. This allow the use of hard drives, network cards and other PCI devices. This issue will only relate to an initial set up where the kernel can list the current devices on the PCI bus.
Future issues will add functionality for attaching drivers to devices and providing a more details description of attached devices and handle PCI-to-PCI devices.

DrDeano:
Closes #244


src/kernel/arch/x86/arch.zig
```diff
@@ -132,6 +132,37 @@ pub fn inb(port: u16) u8 {
...
 }
 
+///
+/// Assembly to write to a given port with a double word of data.
+///
+/// Arguments:
+///     IN port: u16 - The port to write to.
+///     IN data: u32 - The byte of data that will be sent.
+///
+pub fn outl(port: u16, data: u32) void {
```
SamTebbs33: I think something like `out32` would be a better name.

src/kernel/arch/x86/arch.zig
```diff
@@ -132,6 +132,37 @@ pub fn inb(port: u16) u8 {
...
+///
+/// Assembly that reads data from a given port and returns its value.
+///
+/// Arguments:
+///     IN port: u16 - The port to read data from.
+///
+/// Return: u32
+///     The data that the port returns.
+///
+pub fn inl(port: u16) u32 {
```
SamTebbs33: Like `out32`, I think `in32` would be better.

src/kernel/pci.zig
```diff
@@ -0,0 +1,310 @@
...
+        return PciAddress{
+            .bus = self.bus,
+            .device = self.device,
+            .function = function,
+            .register_offset = @enumToInt(pci_reg),
+        };
+    }
+
+    ///
+    /// Read the configuration register date from this device, function and register. This will
```
SamTebbs33: s/date/data

src/kernel/pci.zig
```diff
@@ -0,0 +1,310 @@
...
+    /// return the types width of the register, as PCI reads can only be done on a u32 basis.
```
SamTebbs33: I don't quite understand the role of last part of this sentence. It says that reads can only be done with u32s, which I think is a bit confusing since it returns a integer of a type depending on the register width.

src/kernel/pci.zig
```diff
@@ -0,0 +1,310 @@
...
+            u32 => 0,
+            else => @compileError("Invalid read size. Only u8, u16 and u32 allowed."),
+        };
+        comptime const mask = switch (pci_reg.getWidth()) {
+            u8 => 0xFF,
+            u16 => 0xFFFF,
+            u32 => 0xFFFFFFFF,
+            else => @compileError("Invalid read size. Only u8, u16 and u32 allowed."),
+        };
+        return @intCast(pci_reg.getWidth(), (result >> shift) & mask);
```
SamTebbs33: Can you use @truncate or some of the other builtins to do any of this work? You should be able to replace the masking with @truncate.

src/kernel/pci.zig
```diff
@@ -0,0 +1,310 @@
...
+    }
+};
+
+///
+/// List all the PCI device. This is printed to the serial output.
+///
+fn listAllPCI() void {
```
SamTebbs33: I think it would be better to find all of the devices in `init` or some other helper funciton, return them and list them in kmain. That way they can be printed multiple times without having to re-find them and they can be stored and processed later on.

src/kernel/kmain.zig
```diff
@@ -169,6 +170,8 @@ fn initStage2() noreturn {
 
     tty.print("Hello Pluto from kernel :)\n", .{});
 
+    pci.init();
```
SamTebbs33: PCI is specific to x86 so we should make the device system more abstract in the future.

SamTebbs33 (changes requested):

src/kernel/arch/x86/arch.zig
DrDeano: Or could have one `out` and pass a comptime type that is either u8, u16 or u32

src/kernel/pci.zig
DrDeano: Yes. I was using `@intcast` as it has runtime safety, but `@truncate` would be better.

src/kernel/kmain.zig
DrDeano: I wasn't sure of this. Does ARM have PCI?

src/kernel/arch/x86/arch.zig
SamTebbs33: That's an option. An alternative would be to make `data` be of type `var` and use `@TypeOf' to get the type.

src/kernel/kmain.zig
SamTebbs33: So the Arm architecture doesn't specify how devices are detected and such, that's all up to the board itself. The Raspberry Pi, which is what we're porting to at the moment, doesn't have PCI, so we'll need to make the device system abstract so each arch and board port decides how it finds the devices itself. I'm happy for this abstraction to come in another PR though.

src/kernel/pci.zig
DrDeano: I still need the masking as 0x2 can't be truncated. Plus it is comptime, so there isn't any runtime cost.

src/kernel/arch/x86/arch.zig
DrDeano: I don't think you can return a anytype. So will need to pass a comtime type. (This is for the `in` function)

SamTebbs33: You can return `@TypeOf(data)`.

src/kernel/pci.zig
SamTebbs33: Even if there's no runtime cost, it's cleaner to use what the language provides. What do you mean by 0x2 not being truncatable?

src/kernel/arch/x86/arch.zig
DrDeano: ~~Ah yes :P~~

src/kernel/pci.zig
DrDeano: Actually, I'm not sure how you would use `@truncate` over `@intcast`, could you give an example?

src/kernel/arch/x86/arch.zig
DrDeano: For `in` there is no data to pass in so can't use `@TypeOf(data)` for the return

src/kernel/pci.zig
DrDeano: Have improved the comment

src/kernel/arch/x86/arch.zig
SamTebbs33: True, `in` could have a type given then.

src/kernel/pci.zig
SamTebbs33: `@truncate(pci_reg.getWidth(), result >> shift)` Then you don't need the mask.

DrDeano: Ah, I see, this would only be for the last `@intCast`

SamTebbs33: Yeah sorry :)

src/kernel/arch/x86/arch.zig
```diff
@@ -102,41 +106,67 @@ pub const MEMORY_BLOCK_SIZE: usize = paging.PAGE_SIZE_4KB;
...
-pub fn inb(port: u16) u8 {
-    return asm volatile ("inb %[port], %[result]"
-        : [result] "={al}" (-> u8)
-        : [port] "N{dx}" (port)
-    );
+///     IN comptime Type: type - The type of the data. This can only be u8, u16 or u32.
+///     IN port: u16           - The port to write to.
+///     IN data: Type          - The data that will be sent.
+///
+pub fn out(comptime Type: type, port: u16, data: Type) void {
```
SamTebbs33: We don't need the first type argument if data becomes a `var` type.

SamTebbs33 (changes requested):

SamTebbs33 (approved):
Nice one.

PR #250, opened 2020-10-05
Assert valid aarch64 exception levels

markfirmware:
System registers that are separated by exception level are only available in certain exception levels. This change asserts that only the correct levels are used.
@SamTebbs33


SamTebbs33 (approved):
Looks good 👍

PR #251, opened 2020-10-05
Generate the mocking framework (a bit)

DrDeano:
Instead of adding new types in 5 different places, this adds a generation tool that you add the type once and it will be populated in the framework for you.
This is defo one of those over engineered projects, but I had fun making it.

### DIFF
--- a/test/gen_types.zig
+++ b/test/gen_types.zig
@@ -1,0 +1,258 @@
+const std = @import("std");
+const Allocator = std.mem.Allocator;
+const File = std.fs.File;
+
+// Check duplicate types
+comptime {
+    @setEvalBranchQuota(types.len * types.len * 7);
+    inline for (types) |t1, i| {
+        inline for (types) |t2, j| {
+            if (i != j) {
+                if (std.mem.eql(u8, t1[0], t2[0])) {
+                    @compileError("Duplicate types: " ++ t1[0]);
+                } else if (std.mem.eql(u8, t1[1], t2[1])) {
+                    @compileError("Duplicate enum literal: " ++ t1[1]);
+                }
+            }
+        }
+    }
+}
+
+/// The types needed for mocking
+/// The format is as follows:
+///     1. The type represented as a string. This is because @typeName doesn't play nicely with
+///        all types so this way, what is put here is what you get when generated. There can only
+///        be one of each type.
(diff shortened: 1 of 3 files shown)


test/gen_types.zig
```diff
@@ -0,0 +1,260 @@
...
+// All the function generation parts are the same apart from 3 things
+fn genGenericFunc(comptime intermediate: []const u8, comptime trail: []const u8, comptime end: []const u8) []const u8 {
+    comptime var str: []const u8 = "";
+    inline for (types) |t, i| {
+        if (i == 0) {
+            // Want nice formatting
+            str = str ++ "    " ++ t[0] ++ intermediate ++ t[1] ++ trail;
+        } else {
+            str = str ++ "                " ++ t[0] ++ intermediate ++ t[1] ++ trail;
+        }
```
SamTebbs33: We could probably remove the duplicate code by assigning the big ol' block of spaces to a const and using that.

SamTebbs33 (changes requested):
This is indeed one of the most beautifully over-engineered solutions I have ever seen :D

test/gen_types.zig
```diff
@@ -145,12 +147,12 @@ fn genGenericFunc(comptime intermediate: []const u8, comptime trail: []const u8,
     inline for (types) |t, i| {
         if (i == 0) {
             // Want nice formatting
-            str = str ++ "    " ++ t[0] ++ intermediate ++ t[1] ++ trail;
+            str = str ++ tab_space ++ t[0] ++ intermediate ++ t[1] ++ trail;
         } else {
-            str = str ++ "                " ++ t[0] ++ intermediate ++ t[1] ++ trail;
+            str = str ++ tab_space ** 4 ++ t[0] ++ intermediate ++ t[1] ++ trail;
         }
     }
```
SamTebbs33: I meant something a bit more like
```
const spaces = if (i == 0) " " ** 4 else " " ** 16;
str = str ++ spaces ++ t[0] ++ intermediate ++ t[1] ++ trail;
```

SamTebbs33 (changes requested):

test/gen_types.zig
DrDeano: Sure

SamTebbs33 (approved):
lgtm

PR #253, opened 2020-10-10
Improved gitignore

DrDeano:
Remove some redundant ignores and just have the Zig ignores plus for the generated files.


SamTebbs33:
I'm not sure. This change makes it possible to accidentally add a binary or object file from the project root or source dir. This doesn't cover people that use a build dir other than zig-cache, build etc.

DrDeano:
We could catch binary files being accidentally added in the PR, so if binary files need to be ignored, they can be added to the .gitignore later, but for now, this isn't a problem. The build directory is set in the build script, so idk how people would have different directories.

SamTebbs33:
> We could catch binary files being accidentally added in the PR, so if binary files need to be ignored, they can be added to the .gitignore later, but for now, this isn't a problem. The build directory is set in the build script, so idk how people would have different directories.

Yeah but we can miss things during review. The build dircetory is set to the cache directory, which people can change with an option to `zig build`. I'm not too sure these changes are necessary.

DrDeano:
I can re-add the some of the ignore then, but there are some that we defiantly don't need, plus this also remove the mock file as this is generated anyway

SamTebbs33:
Yeah it would be great to clean up some stuff, like removing the idea one, but I think we should keep ignoring the binaries, elf files, object files and such.

.gitignore
```diff
@@ -1,51 +1,15 @@
...
-
-# Build dir
-bin/*
-*.pyc
-__pycache__
+# Binaries
+**/*.elf
+**/*.bin
+**/*.o
+**/*.iso
```
SamTebbs33, line 5: Are the "**" necessary?

DrDeano: It so it can find these files in side other directories. So will cover all files with these extentions

SamTebbs33: Alright, as long as this also ignores files in the project root, this is ok.

SamTebbs33 (approved):

PR #255, opened 2020-10-18
Feature/code page 437

DrDeano:
This is for the FAT32 unicode to OEM character conversion. Mode FAT32 implementations will use code page 437, but later improvements could allow the user to have different code pages.


SamTebbs33:
What is a code page?

DrDeano:
It is a character encoding that are single byte. There are many for different languages. It is what was before unicode. But FAT32 still uses it. And there is google ;) It is used by Windows and Linux to encode the short name.

SamTebbs33 (approved):
Oh cool! Well my only suggestion is that this goes in the eventual FAT32 file rather than a new "res" directory.

PR #257, opened 2020-10-26
Add symlink support

### LINKED ISSUE #194: VFS symlinks
The virtual filesystem should support files that mirror other files, similar to the way that directory mounts work.

SamTebbs33:
Closes #194﻿

### DIFF
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -10,8 +10,16 @@ pub const OpenFlags = enum {
     CREATE_DIR,
     /// Create a file if it doesn't exist
     CREATE_FILE,
+    /// Create a symlink if it doesn't exist
+    CREATE_SYMLINK,
     /// Do not create a file or directory
     NO_CREATION,
+};
+
+/// The args used when opening new or existing fs nodes
+pub const OpenArgs = struct {
+    /// What to set the target to when creating a symlink
+    symlink_target: ?[]const u8 = null,
 };
 
 /// A filesystem node that could either be a directory or a file
@@ -20,6 +28,9 @@ pub const Node = union(enum) {
     File: FileNode,
     /// The dir node if this represents a directory
     Dir: DirNode,
+    /// The absolute path that this symlink is linked to
+    Symlink: []const u8,
+
     const Self = @This();
 
     ///
@@ -34,7 +45,7 @@ pub const Node = union(enum) {
     pub fn isDir(self: Self) bool {
         return switch (self) {
             .Dir => true,
-            .File => false,
+            else => false,
         };
     }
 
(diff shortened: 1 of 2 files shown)


src/kernel/filesystem/vfs.zig
```diff
@@ -261,15 +298,23 @@ fn traversePath(path: []const u8, flags: OpenFlags) (Allocator.Error || Error)!*
             }
             return switch (node.*) {
                 .File => Error.NotADirectory,
-                .Dir => |*dir| try func(split, try dir.open(seg, rec_flags), rec_flags),
+                .Dir => |*dir| blk: {
+                    var n = try dir.open(seg, rec_flags, .{});
```
DrDeano: Maybe `n` isn't the best variable name lol

src/kernel/filesystem/vfs.zig
```diff
@@ -370,19 +428,34 @@ pub fn openFile(path: []const u8, flags: OpenFlags) (Allocator.Error || Error)!*
...
             file.close();
             break :blk Error.NotADirectory;
         },
+        // We instructed open to folow symlinks above, so this is impossible
+        .Symlink => unreachable,
         .Dir => &node.Dir,
     };
 }
 
+pub fn openSymlink(path: []const u8, target: ?[]const u8, flags: OpenFlags) (Allocator.Error || Error)![]const u8 {
```
DrDeano, line 465: Doc comment

DrDeano (commented):
For the initrd, you could add `.CREATE_SYMLINK` test for opening files and directories in the `open fail with invalid flags` test. Also, in the vfs.open implemention, the follow_sym is true, but in the initFS it is false, is there a reason?

SamTebbs33:
> Also, in the vfs.open implemention, the follow_sym is true, but in the initFS it is false, is there a reason?

Do you mean that it is true in `openSymlink`? What is `initFS`?

DrDeano:
Sorry, I mean `InitrdFS`, and in the tests for InitrdFS, where `open` is used, the `follow_sym` is set to `false` but in the `OpenFile` and `OpenDir` in `vfs.zig`, the `follow_sym` is set to `true`.

SamTebbs33:
> Sorry, I mean `InitrdFS`, and in the tests for InitrdFS, where `open` is used, the `follow_sym` is set to `false` but in the `OpenFile` and `OpenDir` in `vfs.zig`, the `follow_sym` is set to `true`.

Oh right. I used `false` since I knew there wouldn't be any symlinks and so it didn't _have_ to be true, but the behaviour is the same if `true` is used, so I can change it to that. That could make it more reliable in the future too.

DrDeano (approved):

PR #258, opened 2020-10-28
Check for memory leaks in all tests that allocate memory

### LINKED ISSUE #223: Standardise allocator usage and struct freeing
There are currently a few different allocators used in our tests and with the recent addition of the `GeneralPurposeAllocator` into the std lib, I propose that this is made the standard testing allocator to make our tests more uniform (available with `std.testing.allocator`). This allocator allows leak detection with the `detectLeaks` function and is available with `std.testing.allocator_instance`.

In addition we should make sure that each struct whose `init` function takes an allocator also has a corresponding `deinit` function which frees all of the struct's allocations (not the struct itself, just the fields). A test should be added to verify this behaviour.

SamTebbs33:
This standardises the allocator used in tests and adds leak detection to each that allocates memory. Deinitialisation functions have been added when missing. Closes #223﻿

### DIFF
--- a/src/kernel/arch/x86/paging.zig
+++ b/src/kernel/arch/x86/paging.zig
@@ -25,6 +25,23 @@ pub const Directory = packed struct {
 
     /// The tables allocated for the directory. This is ignored by the CPU.
     tables: [ENTRIES_PER_DIRECTORY]?*Table,
+
+    /// The allocator used to allocate and free tables
+    allocator: *std.mem.Allocator,
+
+    ///
+    /// Free the state occupied by the directory. It will be unusable afterwards.
+    ///
+    /// Arguments:
+    ///     IN self: *Self - The directory to deinitialise.
+    ///
+    pub fn deinit(self: *@This()) void {
+        for (self.tables) |table| {
+            if (table) |t| {
+                self.allocator.destroy(t);
+            }
+        }
+    }
 };
 
 /// An array of table entries. Forms the second level of paging and covers a 4MB memory space.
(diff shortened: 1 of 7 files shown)


src/kernel/filesystem/vfs.zig
```diff
@@ -582,39 +582,41 @@ fn testInitFs(allocator: *Allocator) !*TestFS {
...
+        // Create a file within the mounted directory
+        var test_file = try openFile("/mnt/123.txt", .CREATE_FILE);
+        testing.expectEqual(@ptrCast(*const FileSystem, testfs2.fs), test_file.fs);
+        // This shouldn't be in the root fs
+        testing.expectEqual(@as(usize, 1), testfs.tree.children.items.len);
+        testing.expectEqual(@as(usize, 0), testfs.tree.children.items[0].children.items.len);
+        // It should be in the mounted fs
+        testing.expectEqual(@as(usize, 1), testfs2.tree.children.items.len);
+        testing.expectEqual(test_file, &testfs2.tree.children.items[0].val.File);
+    }
```
DrDeano: Can remove the `{}` as there is only one test

src/kernel/arch/x86/paging.zig
```diff
@@ -25,6 +25,22 @@ pub const Directory = packed struct {
 
     /// The tables allocated for the directory. This is ignored by the CPU.
     tables: [ENTRIES_PER_DIRECTORY]?*Table,
+
+    /// The allocator used to allocate and free tables
+    allocator: *std.mem.Allocator,
+
+    /// Free the state occupied by the directory. It will be unusable afterwards.
```
DrDeano, line 33: Missing the leading `///`

DrDeano (approved):

Issue #230, opened 2020-08-29
Load and execute ELF files


SamTebbs33:
Now that the kernel supports multitasking, we should be able to load an elf executable from memory and execute it. The standard library has some ELF parsing code, but reads from files rather than directly from memory, so this depends on #229.

SamTebbs33:
On second thoughts, this shouldn't use the OS abstraction layer as that would require doing many syscalls from the kernel to the kernel, which is inefficient.

As part of this ticket we should also make it possible to change the attributes of mapped memory, if that is necessary for elf loading and execution.

PR #261, opened 2020-11-06
Load ELF files

SamTebbs33:
This PR adds the ability to load elf files from memory. I have chosen not to use the std lib ELF construct since it uses a lot of magic constants rather than enums, and isn't packed so requires a lot of extra processing to load. This implementation just needs a simple bit cast to load from memory. Tests will come in a later update.﻿

Part of #230

### DIFF
--- a/src/kernel/elf.zig
+++ b/src/kernel/elf.zig
@@ -1,0 +1,693 @@
+const std = @import("std");
+const builtin = @import("builtin");
+const Arch = std.Target.Cpu.Arch;
+const Endian = builtin.Endian;
+const log = std.log.scoped(.elf);
+const testing = std.testing;
+
+/// The data sizes that ELF files support. The int value corresponds to the value used in the file
+pub const DataSize = enum(u8) {
+    /// 32-bit
+    ThirtyTwoBit = 1,
+    /// 64-bit
+    SixtyFourBit = 2,
+
+    ///
+    /// Get the number of bits taken by the data size
+    ///
+    /// Arguments:
+    ///     IN self: DataSize - The data size to get the number of bits for
+    ///
+    /// Return: usize
+    ///     The number of bits
+    ///
+    pub fn toNumBits(self: @This()) usize {
+        return switch (self) {
+            .ThirtyTwoBit => 32,
+            .SixtyFourBit => 64,
+        };
+    }
+};
+
+/// The endiannesses supported by Elf files. The int value corresponds to the value used in the file
+pub const Endianness = enum(u8) {
+    /// Little-endian
+    Little = 1,
+    /// Big-endian
+    Big = 2,
+
+    ///
+    /// Translate into the corresponding std lib Endian
+    ///
+    /// Arguments:
+    ///     IN self: Endianness - The endianness to translate
+    ///
+    /// Return: Endian
+    ///     The corresponding std lib Endian value
+    ///
+    pub fn toEndian(self: @This()) Endian {
+        return switch (self) {
+            .Big => .Big,
+            .Little => .Little,
+        };
+    }
+};
+
+/// The type of the elf file. The int value corresponds to the value used in the file
+pub const Type = enum(u16) {
+    /// Unused
+    None = 0,
+    /// Relocatable
+    Rel = 1,
+    /// Executable
+    Executable = 2,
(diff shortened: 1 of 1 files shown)


src/kernel/elf.zig
```diff
@@ -0,0 +1,449 @@
...
+    ///
+    /// Error: Allocator.Error || Error
+    ///     Allocator.Error - There wasn't enough memory free to allocate the required state
+    ///     Error.InvalidMagicNumber - The ELF file magic number wasn't as expected
+    ///     Error.InvalidArchitecture - The ELF file wasn't built for the expected architecture
+    ///     Error.InvalidDataSize - The ELF file wasn't built for the data size supported by the given architecture
+    ///     Error.InvalidEndianness - The ELF file wasn't built with the endianness supported by the given architecture
+    ///
+    pub fn init(elf_data: []const u8, arch: Arch, allocator: *std.mem.Allocator) (std.mem.Allocator.Error || Error)!Self {
+        const header = @ptrCast(*const Header, elf_data.ptr).*;
```
DrDeano: Maybe use `std.mem.bytesToValue`

src/kernel/elf.zig
```diff
@@ -0,0 +1,449 @@
...
+        for (program_segments) |*segment| {
+            segment.* = @ptrCast(*const ProgramHeader, elf_data.ptr + seg_offset).*;
+            seg_offset += header.program_header_entry_size;
+        }
+
+        var section_headers = try allocator.alloc(SectionHeader, header.section_header_entries);
+        var section_data = try allocator.alloc(?[]const u8, header.section_header_entries);
+        var sec_offset = header.section_header_offset;
+        for (section_headers) |*section, i| {
+            section.* = @ptrCast(*const SectionHeader, elf_data.ptr + sec_offset).*;
```
DrDeano: Same here, could use `std.mem.bytesToValue`

src/kernel/elf.zig
```diff
@@ -0,0 +1,449 @@
...
+            .Aarch64 => .aarch64,
+            .RISC_V => .riscv32,
+        };
+    }
+};
+
+/// The header describing the entire ELF file
+pub const Header = packed struct {
+    /// Should be 0x7f | 0x45 | 0x4C | 0x46
+    magic_number: usize,
```
DrDeano: In 64 bit, `usize` will be 8 bytes and not 4 bytes wide. So would need to use u32. As the magic part of the header is 4 bytes.

DrDeano (commented):
A few minor changed. As you say the tests will come in a later PR, could make sure the packet structs are behaving properly for you, as for my FAT32, I can't use packet structs yet.

src/kernel/elf.zig
SamTebbs33: This mistake caused me so much trouble when testing!

SamTebbs33:
> A few minor changed. As you say the tests will come in a later PR, could make sure the packet structs are behaving properly for you, as for my FAT32, I can't use packet structs yet.

I meant that tests will come in a later push, which they now finally have! Packed structs have been working well for me so far.

src/kernel/elf.zig
```diff
@@ -175,6 +178,10 @@ pub const Header = packed struct {
     section_name_index: u16,
 };
 
+comptime {
+    std.debug.assert(@sizeOf(Header) == if (@bitSizeOf(usize) == 32) 0x32 else 0x40);
+}
+
```
DrDeano: I think these should be grouped together some where.

src/kernel/elf.zig
```diff
@@ -446,4 +458,187 @@ pub const Error = error{
...
+    data_offset += 1;
+
+    std.mem.copy(u8, data[data_offset .. data_offset + string_section_name.len], string_section_name);
+    data_offset += string_section_name.len;
+    data[data_offset] = 0;
+    data_offset += 1;
+    return data[0..data_size];
+}
+
+test "init" {
```
DrDeano, line 557: Could add tests where you corrupt/modify parts of the ELF header and test the error conditions and memory leaks when there is an error.

SamTebbs33: Done!

DrDeano (approved):
🚀

PR #262, opened 2020-11-06
Rename the llog to logger to make is all consistent

DrDeano:


SamTebbs33 (commented):
I'm not really sure why this is necessary. Wasn't `log` the consistently used name before?

DrDeano:
> 
> 
> I'm not really sure why this is necessary. Wasn't `log` the consistently used name before?

I can use `log` instead of `logger` as I was using logger for some of my stuff, but can move to `log`. I wasn't sure which one was the original, but log is shorter ;P

SamTebbs33:
> > I'm not really sure why this is necessary. Wasn't `log` the consistently used name before?
> 
> I can use `log` instead of `logger` as I was using logger for some of my stuff, but can move to `log`. I wasn't sure which one was the original, but log is shorter ;P

Oh I see! I've only seen `log` in the source so I thought that was the only word used. Awesome.

SamTebbs33 (approved):

PR #263, opened 2020-11-09
mkFAT32 use a anytype stream

DrDeano:
This allows mkFAT32 to work on a File or a fixedBufferStream or any streram that allows for reader(), writer() or seekableStream() interfaces.

This will be part of #216


SamTebbs33 (approved):
Nice :)

PR #264, opened 2020-11-11
Feature/initial fat32

DrDeano:
This initial FAT32 filesystem will parse a stream, which could be a HDD or a in memory filesystem and produce a FAT32 filesystem. This will check if the raw bytes is a valid FAT32 filesystem as described in the spec.

This is only a skeleton with no open, read or write functionality.

Part of #216

p.s. have fun, this is a long one, lots more to come in later PR :P


SamTebbs33:
Sorry, I had a quick look a couple of days ago but got distracted. I'll have a proper look tomorrow :)

src/kernel/code_page.zig
```diff
@@ -0,0 +1,88 @@
...
+    /// Return: u8
+    ///     The converted character.
+    ///
+    /// Error: Error
+    ///     error.InvalidChar - The character to be converted is not in the code page table.
+    ///
+    pub fn toCodePage(code_page: CodePages, char: u16) Error!u8 {
+        const table = switch (code_page) {
+            .CP437 => cp437.table,
+        };
```
SamTebbs33: This could be made nice and abstract by adding a function to `CodePages` that returns the table. This could then be used in `toWideChar` as well.

src/kernel/code_page.zig
```diff
@@ -0,0 +1,88 @@
+const std = @import("std");
```
SamTebbs33: Since this is more of a supporting file, it could be moved to src/kernel/code_page

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+const vfs = @import("vfs.zig");
+const mem = @import("../mem.zig");
+const CodePage = @import("../code_page.zig").CodePage;
+const mkfat32 = @import("../../../mkfat32.zig");
+
+/// The boot record for FAT32. This is use for parsing the initial boot sector to extract the
+/// relevant information for todays FAT32.
+const BootRecord = struct {
+    /// The jump bytes that begin the boot code. This will jump over the FAT32 header.
+    jmp: [3]u8,
```
SamTebbs33, line 20: Do we need to keep this around if the disk is only used for storage and not booting?

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+    /// The number of reserved sectors at the beginning of the image. This is where the fat
+    /// header, FSInfo and FAT is stored.
+    reserved_sectors: u16,
+
+    /// The number of FAT's.
+    fat_count: u8,
+
+    /// The size in bytes of the root directory. This is only used by FAT12 and FAT16, so is
+    /// always 0 for FAT32.
+    zero_root_directory_size: u16,
```
SamTebbs33: The `zero` part of the name is a bit confusing. I think that documentation saying that it's 0 for fat32 is enough. Same for the other fields with `zero` in their name.

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+    signature: u8,
+
+    /// The serial number of the FAT image at format. This is a function of the current
+    /// timestamp of creation.
+    serial_number: u32,
+
+    /// The partitioned volume label.
+    volume_label: [11]u8,
+
+    /// The file system type.
```
SamTebbs33: I think it would be nice to have some documentation as to why these are arrays.

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+    lead_signature: u32,
+
+    /// Reserved bytes
+    reserved0: [480]u8,
+
+    /// The middle or struct signature: 0x61417272
+    struct_signature: u32,
+
+    /// The number of free clusters in the image
+    number_free_cluster: u32,
```
SamTebbs33: cluster**s**?

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+    /// Check the attributes and check if the entry is a directory.
+    ///
+    /// Arguments:
+    ///     IN self: *const ShortName - The short name entry to get if this is a directory.
+    ///
+    /// Return: bool
+    ///     Whether the file is a directory or not.
+    ///
+    pub fn isDir(self: *const ShortName) bool {
+        return self.attributes & @enumToInt(ShortName.Attributes.Directory) == @enumToInt(ShortName.Attributes.Directory);
```
SamTebbs33, line 354: `@enumToInt(ShortName.Attributes.Directory) == @enumToInt(ShortName.Attributes.Directory);` will never be false.

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+    ///
+    pub fn clusterToSector(self: *const FATConfig, cluster: u32) u32 {
+        // FAT count will be 2 as this is checked in the init function
+        return (self.sectors_per_fat * 2) + self.reserved_sectors + ((cluster - 2) * self.sectors_per_cluster);
+    }
+};
+
+///
+/// Initialise a struct from bytes.
+/// TODO: Once packed structs are good to go, then use std.mem.bytesAsValue.
```
SamTebbs33, line 466: I've had success with packed structs in my ELF work, have you tried using them recently?

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+        ///     IN self: *Fat32Self - Self to free.
+        ///
+        pub fn deinit(self: *Fat32Self) void {
+            // Make sure we have closed all files
+            // TODO: Should this deinit close any open files instead?
+            std.debug.assert(self.opened_files.count() == 0);
+            self.opened_files.deinit();
+            self.allocator.destroy(self.root_node.node);
+            self.allocator.destroy(self.fs);
+            self.allocator.destroy(self);
```
SamTebbs33: In general I don't think deinit functions should destroy the struct itself since that would mean you couldn't deinit stack-allocated structs.

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+        /// Error: Allocator.Error || ReadError || SeekError || Error
+        ///     Allocator.Error - If there is no more memory. Any memory allocated will be freed.
+        ///     ReadError       - If there is an error reading from the stream.
+        ///     SeekError       - If there si an error seeking the stream.
+        ///     Error           - If there is an error when parsing the stream to set up a fAT32
+        ///                       filesystem. See Error for the list of possible errors.
+        ///
+        pub fn init(allocator: *Allocator, stream: StreamType) (Allocator.Error || ReadError || SeekError || Error)!*Fat32Self {
+            log.info("Init\n", .{});
+            defer log.info("Done\n", .{});
```
SamTebbs33: Fat32 initialisation will be happening after boot initialisation too, so I don't think these log messages are appropriate. We don't want them appearing when the user plugs in a memory stick, for example (we'd want device related log messages instead).

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+///     IN stream: anytype       - A stream this is used to read and seek a raw disk or memory to
+///                                be parsed as a FAT32 filesystem. E.g. a FixedBufferStream.
+///
+/// Return: *Fat32FS(@TypeOf(stream))
+///     A pointer to a FAT32 filesystem.
+///
+/// Error: Allocator.Error || Fat32FS(@TypeOf(stream)).Error
+///     Allocator.Error - If there isn't enough memory to create the filesystem.
+///
+pub fn initialiseFAT32(allocator: *Allocator, stream: anytype) (Allocator.Error || ErrorSet(@TypeOf(stream)) || Fat32FS(@TypeOf(stream)).Error)!*Fat32FS(@TypeOf(stream)) {
```
SamTebbs33: Do we need this when we have Fat32FS.init?

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+    return Fat32FS(@TypeOf(stream)).init(allocator, stream);
+}
+
+/// The number of allocations that the init function make.
+const init_allocations: usize = 5;
+
+/// The test buffer for the test filesystem stream.
+var stream_buff: [if (builtin.is_test) 34090496 else 0]u8 = undefined;
```
SamTebbs33: It would be good to prefix these variables used for testing with `test`.

src/kernel/filesystem/fat32.zig
```diff
@@ -0,0 +1,1756 @@
...
+
+    const test_path_prefix = "test/fat32";
+    const files = &[_][]const u8{
+        "/....leading_dots.txt",
+        "/[nope].txt",
+        "/A_verY_Long_File_namE_With_normal_Extension.tXt",
+        "/dot.in.file.txt",
+        "/file.long_ext",
+        "/file.t x t",
+        "/insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long.txt",
```
SamTebbs33, line 1029: nice xD

SamTebbs33 (changes requested):
Looking good! Nice job on getting so much done.

src/kernel/filesystem/fat32.zig
DrDeano: This is only here for parsing the boot record. This struct isn't used in the main FAT32FS.

src/kernel/filesystem/fat32.zig
DrDeano: `self.attributes & @enumToInt(ShortName.Attributes.Directory)` will evaluate first then the equality. I have a test for this.

src/kernel/filesystem/fat32.zig
DrDeano: Shall i change it to debug logs

src/kernel/filesystem/fat32.zig
DrDeano: It just makes the API easier to use. Plus this is how I have seen it in the std lib.

src/kernel/filesystem/fat32.zig
DrDeano: If you can think of other test files, let me know

src/kernel/code_page.zig
DrDeano: Not really sure what you mean. is it like:
```Zig
pub fn getTable(code_page: CodePage) []u8 {
    return switch (code_page) {
            .CP437 => cp437.table,
        };
}
```

src/kernel/filesystem/fat32.zig
DrDeano: I have did an update and when I try using packed structs, it doesn't compile and not error returned, just exit code 1. Plus the sizes are not expected. So will have to wait a bit longer.

src/kernel/filesystem/fat32.zig
SamTebbs33: Oh of course, I was reading this after writing some C++ at work and read the `&` as `&&`! Perhaps some parentheses would be nice anyway.

src/kernel/filesystem/fat32.zig
SamTebbs33: Yeah that's a good idea

src/kernel/code_page.zig
SamTebbs33: Exactly. That could go in the struct.

DrDeano: what would the use be of it though? but can add it

SamTebbs33: Well you have this logic in a couple of places so it would be good to extract it into the enum itself. Plus, encapsulation is nice :)

DrDeano: I see now

SamTebbs33 (approved):
:rocket:

PR #266, opened 2020-11-20
Rework copyDataToVMM to work in both directions

SamTebbs33:
This patch reworks vmm.copyDataToVMM to allow it to either copy to or from another VMM. This new capability will be used in syscalls when copying data from user memory to kernel memory.﻿

### DIFF
--- a/src/kernel/vmm.zig
+++ b/src/kernel/vmm.zig
@@ -424,10 +424,11 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
         /// Copy data from an address in a virtual memory manager to an address in another virtual memory manager
         ///
         /// Arguments:
-        ///     IN self: *Self - The virtual memory that owns the data being copied. This must be the VMM currently in use
-        ///     IN to: *const Self - The virtual memory manager that owns the address that the data is being copied to
-        ///     IN data: []const u8 - The data being copied. Must be within memory mapped in `self`
-        ///     IN dest: usize - The address within `to` to copy the data to. The space covered by `dest` and `dest` + `dest.len` must be mapped in `to`
+        ///     IN self: *Self - One of the VMMs to copy between. This should be the currently active VMM
+        ///     IN other: *Self - The second of the VMMs to copy between
+        ///     IN data: []u8 - The being copied from or written to (depending on `from`). Must be mapped within the VMM being copied from/to
+        ///     IN address: usize - The address within `other` that is to be copied from or to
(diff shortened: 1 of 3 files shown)


src/kernel/vmm.zig
```diff
@@ -777,7 +782,7 @@ test "copy" {
     std.testing.expectError(VmmError.NotAllocated, mirrored.virtToPhys(alloc4));
 }
 
-test "copyDataToVMM" {
+test "copyData" {
```
DrDeano, line 785: Probs worth added some test for copying the other way around.

src/kernel/vmm.zig
```diff
@@ -424,29 +424,30 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
         /// Arguments:
-        ///     IN self: *Self - The virtual memory that owns the data being copied. This must be the VMM currently in use
-        ///     IN to: *const Self - The virtual memory manager that owns the address that the data is being copied to
-        ///     IN data: []const u8 - The data being copied. Must be within memory mapped in `self`
-        ///     IN dest: usize - The address within `to` to copy the data to. The space covered by `dest` and `dest` + `dest.len` must be mapped in `to`
+        ///     IN self: *Self - One of the VMMs to copy between. This should be the currently active VMM
+        ///     IN other: *Self - The second of the VMMs to copy between
+        ///     IN data: []u8 - The being copied from or written to (depending on `from`). Must be mapped within the VMM being copied from/to
+        ///     IN address: usize - The address within `other` that is to be copied from or to
+        ///     IN from: bool - Whether the date should be copied from `self` to `other, or the other way around
```
DrDeano, line 431: Should the comment be: `Whether the date should be copied from 'data' to 'other' or the the other way around`

src/kernel/vmm.zig
SamTebbs33: That is actually below this. Have a look at the lines starting with `try vmm2.copyData(&vmm, buff2, alloc, false);`.

src/kernel/vmm.zig
SamTebbs33: This is talking about which VMM is being copied from/to, so that's what `self` refers to. It's not referring to the data itself.

src/kernel/vmm.zig
DrDeano: Are yes, I am blind

DrDeano (approved):
:)

PR #267, opened 2020-11-22
Add error handling for syscalls

SamTebbs33:
This patch makes it possible for a syscall handler to return an error that is returned to the caller via an error code. This error code is passed in ebx for x86. This has a decent number of implementation changes so I thought I'd submit the PR before adding tests and docs so you can critique it.

### DIFF
--- a/src/kernel/arch/x86/syscalls.zig
+++ b/src/kernel/arch/x86/syscalls.zig
@@ -9,6 +9,7 @@ const testing = std.testing;
 const expect = std.testing.expect;
 const isr = @import("isr.zig");
 const panic = @import("../../panic.zig").panic;
+const syscalls = @import("../../syscalls.zig");
 
 /// The isr number associated with syscalls
 pub const INTERRUPT: u16 = 0x80;
@@ -17,16 +18,20 @@ pub const INTERRUPT: u16 = 0x80;
 pub const NUM_HANDLERS: u16 = 256;
 
 /// A syscall handler
-pub const SyscallHandler = fn (ctx: *arch.CpuState, arg1: u32, arg2: u32, arg3: u32, arg4: u32, arg5: u32) u32;
+pub const Handler = fn (ctx: *arch.CpuState, arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) syscalls.Error!usize;
 
 /// Errors that syscall utility functions can throw
-pub const SyscallError = error{
+pub const Error = error{
     SyscallExists,
     InvalidSyscall,
 };
 
+comptime {
+    std.debug.assert(@typeInfo(syscalls.Syscall).Enum.fields.len <= NUM_HANDLERS);
+}
+
 /// The array of registered syscalls
-var handlers: [NUM_HANDLERS]?SyscallHandler = [_]?SyscallHandler{null} ** NUM_HANDLERS;
+var handlers: [NUM_HANDLERS]?Handler = [_]?Handler{null} ** NUM_HANDLERS;
 
 ///
 /// Returns true if the syscall is valid, else false.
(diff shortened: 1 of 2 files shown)


src/kernel/arch/x86/syscalls.zig
```diff
@@ -70,151 +81,199 @@ fn handle(ctx: *arch.CpuState) u32 {
 /// Register a syscall so it can be called by triggering interrupt 128.
 ///
 /// Arguments:
-///     IN syscall: u8 - The syscall number to register the handler with.
-///     IN handler: SyscallHandler - The handler to register the syscall with.
+///     IN syscall: Syscall - The syscall to register the handler with.
```
SamTebbs33: I still have to fix up some of the doc comment changes such as this one.

src/kernel/syscalls.zig
```diff
@@ -0,0 +1,62 @@
...
+            const error2_instance = @field(Error, err2.name);
+            if (error_instance != error2_instance and toErrorCode(error_instance) == toErrorCode(error2_instance)) {
+                @compileError("Syscall errors '" ++ err.name ++ "' and '" ++ err2.name ++ "' share the same error code\n");
+            }
+        }
+    }
+}
+
+pub const Syscall = enum {
+    Void,
```
SamTebbs33: This is just here as an example so the compiler doesn't complain that the enum is empty. I'll be replacing it with proper syscalls once this PR is merged.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Done

src/kernel/syscalls.zig
SamTebbs33: Done

src/kernel/arch/x86/syscalls.zig
```diff
@@ -257,78 +359,115 @@ pub fn init() void {
...
 
+fn testHandler6(ctx: *arch.CpuState, arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) syscalls.Error!usize {
+    return syscalls.Error.OutOfMemory;
+}
+
 test "registerSyscall returns SyscallExists" {
-    registerSyscall(123, testHandler0) catch unreachable;
-    registerSyscall(123, testHandler0) catch |err| {
+    registerSyscall(122, testHandler0) catch unreachable;
+    registerSyscall(122, testHandler0) catch |err| {
```
DrDeano: Realized these should be try's and can use `expectError` instead of returning. Oh the joys of looking as old code and seeing how young we once were ;)

src/kernel/syscalls.zig
```diff
@@ -0,0 +1,122 @@
...
+        return switch (self) {
+            .Test1 => handleTest1,
+            .Test2 => handleTest2,
+            .Test3 => handleTest3,
+        };
+    }
+};
+
+/// A function that can handle a syscall and return a result or an error
+pub const Handler = fn (arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) Error!usize;
```
DrDeano, line 81: This has the same name as the handler in the arch syscall file. Maybe these could have different names

src/kernel/arch/x86/syscalls.zig
```diff
@@ -257,78 +359,115 @@ pub fn init() void {
...
     }
 
-    if (syscall0(123) != 0 or test_int != 1) {
-        panic(@errorReturnTrace(), "FAILURE syscall0\n", .{});
+    if (syscall0(122)) |res| {
+        if (res != 0 or test_int != 1) {
+            panic(@errorReturnTrace(), "FAILURE syscall0\n", .{});
+        }
+    } else |e| {
+        panic(@errorReturnTrace(), "FAILURE syscall0 errored\n", .{});
```
DrDeano: Could also add the error to the print. Same for the others

src/kernel/syscalls.zig
```diff
@@ -0,0 +1,122 @@
+const std = @import("std");
+const scheduler = @import("scheduler.zig");
+const panic = @import("panic.zig").panic;
+const log = std.log.scoped(.syscalls);
+
+/// A compilation of all errors that syscall handlers could return.
+pub const Error = error{OutOfMemory};
```
DrDeano, line 7: An idea:
As OutOfMemory is an `Allocator` thing, make `Error` a union of `Allocator.Error`. Then for `fromErrorCode` and `toErrorCode` use `error.OutOfMemory` (notice the lowercase 'e') or use the full `Allocator.Error.OutOfMemory` so if the Allocator interface changes, then the comptime code will fail so can update the error set

DrDeano (commented):
Getting excited to implement some syscalls for real. Keep some for me ;)

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Good point :D We'll never be as young and ignorant as before

src/kernel/syscalls.zig
SamTebbs33: So I considered making this an error union, but decided against it when I found out that errors of the same name can be coerced. For example, if error set `A` contains `a, b, c` and error set `B` contains `b, c`, you can return B.b from a function that returns A and it will be coerced (see [here](https://ziglang.org/documentation/master/#Errors)). Not using an error union here and instead using error coercion means that we only need to handle the errors that are actually thrown, rather than the whole set from the union. For example, if we add `d` to the `B` from above, but the functions that a syscall uses never actually return it, then we don't have to handle it here and add it to `A`, and we avoid unnecessary code. A more concrete example is that if e.g. the vfs couldn't return a `.NotAFile`, even if it is part of the vfs error set, then using an error union means we'd have to account for it anyway, but using error coercion means we wouldn't because it's never going to be returned!

src/kernel/syscalls.zig
SamTebbs33: I don't see there being a problem with the duplicate names since they're in different namespaces.

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Good idea :+1:

src/kernel/syscalls.zig
DrDeano: If it wouldn't be a problem, then should be fine, just didn't want confusion of the 2 types as are different types.

src/kernel/syscalls.zig
DrDeano: Fab, didn't consider the error coerction. I shall brush up on my error sets ;)

src/kernel/syscalls.zig
SamTebbs33: Yeah I think we can keep them separated nicely.

DrDeano (approved):
👍

PR #271, opened 2020-11-30
Added the ability to close a directory node

### LINKED ISSUE #270: VFS close directories
At the moment, there isn't a way to close a directory, this can result in lost memory when opening a directory without the ability to close it.

DrDeano:
Closes #270

### DIFF
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -87,13 +87,13 @@ pub const FileSystem = struct {
     const Self = @This();
 
     ///
-    /// Close an open file, performing any last operations required to save data etc.
+    /// Close an open node, performing any last operations required to save data etc.
     ///
     /// Arguments:
     ///     IN self: *const FileSystem - The filesystem in question being operated on.
-    ///     IN node: *const FileNode - The file being closed
+    ///     IN node: *const Node - The node being closed.
     ///
-    const Close = fn (self: *const Self, node: *const FileNode) void;
+    const Close = fn (self: *const Self, node: *const Node) void;
 
     ///
     /// Read from an open file
@@ -191,7 +191,8 @@ pub const FileNode = struct {
 
     /// See the documentation for FileSystem.Close
     pub fn close(self: *const FileNode) void {
-        return self.fs.close(self.fs, self);
+        // TODO: Use @fieldParentPtr() once implemented for unions
+        return self.fs.close(self.fs, @ptrCast(*const Node, self));
     }
 
     /// See the documentation for FileSystem.Write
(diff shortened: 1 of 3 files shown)


src/kernel/filesystem/vfs.zig
```diff
@@ -87,13 +87,13 @@ pub const FileSystem = struct {
     const Self = @This();
 
     ///
-    /// Close an open file, performing any last operations required to save data etc.
+    /// Close an open file or directory, performing any last operations required to save data etc.
```
SamTebbs33: Maybe change this to say "an open node" since any node, not just directories and files, can be closed with these changes.

src/kernel/filesystem/vfs.zig
```diff
@@ -191,7 +191,7 @@ pub const FileNode = struct {
 
     /// See the documentation for FileSystem.Close
     pub fn close(self: *const FileNode) void {
-        return self.fs.close(self.fs, self);
+        return self.fs.close(self.fs, @ptrCast(*const Node, self));
```
SamTebbs33, line 195: Instead of casting it would make more sense if this function becomes part of the `Node` struct.

src/kernel/filesystem/vfs.zig
```diff
@@ -218,6 +218,17 @@ pub const DirNode = struct {
         }
         return fs.open(fs, node, name, flags, args);
     }
+
+    /// See the documentation for FileSystem.Close
+    pub fn close(self: *const DirNode) void {
```
SamTebbs33, line 223: As with `FileNode.close`, we should make the close function be part of the `Node` struct instead.

SamTebbs33 (changes requested):

src/kernel/filesystem/vfs.zig
DrDeano: This wouldn't work as you can't close a symlink. Plus openFile and openDir return FileNode and DirNode so there will need to be a cast at some point.

SamTebbs33: I reckon we could close a symlink, after all it's going to be kept around in memory by the FS and we'll want to get rid of it from memory once it isn't used. openFile and openDir do indeed return specialised nodes, but when converting a `Node` to a `DirNode` or `FileNode` we switch on them rather than cast since it's much safer and allows us to detect invalid nodes being passed.

DrDeano: How would you do back from a `FileNode` or `DirNode` to a `Node`?

SamTebbs33: Well you wouldn't have to if the function is part of `Node` and takes a `*Node` or `*const Node` as `self`.

DrDeano: I had a look, but changes a lot of types from FileNode and DirNode to just Node and mean that I will need to change the mount type to Node instead of DirNode. I'm not sure how to do this. Could you show me da way sensei.
Was thinking of having 2 close functions, one for FileNode and one for DirNode.

SamTebbs33: Oh I see that the problem is that with this a close function that takes a `Node` we can't do
```
var file = vfs.openFile(...)
file.close()
```

A solution is to have a `close` function in in `FIleNode` and `DirNode` like you do now, then have a `close` function for each of those in `FileSystem` (i.e. `closeFile`, `closeDir`). This avoids the casting that could be dodgy. I'm not sure how to tackle closing symlinks, that could be tended to later on.

SamTebbs33 (approved):
Noice

SamTebbs33 (approved):

PR #272, opened 2020-12-01
Feature/fat32 open

DrDeano:
Needed to return a pointer to the file system as `@fieldParentPtr(Fat32Self, "instance", fs.instance)` doesn't work as the the pointer to `instance` is on the stack, then goes out of scope when the init is returned. So will be using free stack space.
Part of #216

### DIFF
--- a/mkfat32.zig
+++ b/mkfat32.zig
@@ -557,13 +557,16 @@ pub const Fat32 = struct {
 
     ///
     /// Make a FAT32 image. This will either use the default options or modified defaults from the
-    /// user. The file will be saved to the path specified.
+    /// user. The file will be saved to the path specified. If quick format is on, then the entire
+    /// stream is zeroed else the reserved and FAT sectors are zeroed.
     ///
     /// Argument:
-    ///     IN options: Options - The FAT32 options that the user can provide to change the
-    ///                           parameters of a FAT32 image.
-    ///     IN stream: anytype  - The stream to create a new FAT32 image. This stream must support
-    ///                           reader(), writer() and seekableStream() interfaces.
+    ///     IN options: Options   - The FAT32 options that the user can provide to change the
+    ///                             parameters of a FAT32 image.
+    ///     IN stream: anytype    - The stream to create a new FAT32 image. This stream must
+    ///                             support reader(), writer() and seekableStream() interfaces.
+    ///     IN quick_format: bool - Whether to completely zero the stream initially or zero just
+    ///                             the important sectors.
     ///
     /// Error:  @TypeOf(stream).WriteError ||  @TypeOf(stream).SeekError || Error
     ///     @TypeOf(stream).WriteError       - If there is an error when writing. See the relevant error for the stream.
@@ -572,14 +575,19 @@ pub const Fat32 = struct {
     ///     Error.TooLarge                   - The stream size is too small. < 17.5KB.
     ///     Error.TooSmall                   - The stream size is to large. > 2TB.
     ///
-    pub fn make(options: Options, stream: anytype) (ErrorSet(@TypeOf(stream)) || Error)!void {
+    pub fn make(options: Options, stream: anytype, quick_format: bool) (ErrorSet(@TypeOf(stream)) || Error)!void {
         // First set up the header
         const fat32_header = try Fat32.createFATHeader(options);
-        // Get the total image size again. As the above has a check for the size, we don't need one here again
-        const image_size = std.mem.alignBackward(options.image_size, fat32_header.bytes_per_sector);
 
         // Initialise the stream with all zeros
-        try stream.writer().writeByteNTimes(0x00, image_size);
+        try stream.seekableStream().seekTo(0);
+        if (quick_format) {
+            // Zero just the reserved and FAT sectors
+            try stream.writer().writeByteNTimes(0x00, (fat32_header.reserved_sectors + (fat32_header.sectors_per_fat * 2)) * fat32_header.bytes_per_sector);
+        } else {
+            const image_size = std.mem.alignBackward(options.image_size, fat32_header.bytes_per_sector);
+            try stream.writer().writeByteNTimes(0x00, image_size);
+        }
 
         // Write the boot sector with the bootstrap code and header and the backup boot sector.
         try Fat32.writeBootSector(stream, fat32_header);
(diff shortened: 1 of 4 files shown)


mkfat32.zig
```diff
@@ -557,13 +557,16 @@ pub const Fat32 = struct {
...
     /// Argument:
-    ///     IN options: Options - The FAT32 options that the user can provide to change the
-    ///                           parameters of a FAT32 image.
-    ///     IN stream: anytype  - The stream to create a new FAT32 image. This stream must support
-    ///                           reader(), writer() and seekableStream() interfaces.
+    ///     IN options: Options   - The FAT32 options that the user can provide to change the
+    ///                             parameters of a FAT32 image.
+    ///     IN stream: anytype    - The stream to create a new FAT32 image. This stream must
+    ///                             support reader(), writer() and seekableStream() interfaces.
+    ///     IN quick_format: bool - Weather to completely zero the stream initially or zero just
```
SamTebbs33: s/weather/whether/

src/kernel/filesystem/fat32.zig
```diff
@@ -1195,16 +1216,56 @@ pub fn Fat32FS(comptime StreamType: type) type {
...
         ///     vfs.Error.NoSuchFileOrDir - Error if the file/folder doesn't exist.
+        ///     vfs.Error.Unexpected      - An error occurred whilst reading the file system, this
+        ///                                 can be caused by a parsing error or errors on reading
+        ///                                 or seeking the underlying stream. If this occurs, then
+        ///                                 the real error is printed using `log.err`.
         ///
-        fn openImpl(fs: *const vfs.FileSystem, dir: *const vfs.DirNode, name: []const u8) (Allocator.Error || ReadError || SeekError || vfs.Error)!*vfs.Node {
+        fn openImpl(fs: *const vfs.FileSystem, dir: *const vfs.DirNode, name: []const u8) (Allocator.Error || vfs.Error)!*vfs.Node {
             const self = @fieldParentPtr(Fat32Self, "instance", fs.instance);
+            // TODO: Cache the files in this dir, so when opening, don't have to iterator the directory every time
```
SamTebbs33, line 1229: :+1: We should definitely add some fs caching at some point as operations will be slow otherwise

src/kernel/filesystem/fat32.zig
```diff
@@ -1195,16 +1216,56 @@ pub fn Fat32FS(comptime StreamType: type) type {
...
+            defer it.deinit();
+            while (it.next() catch |e| switch (e) {
+                error.OutOfMemory => return vfs.Error.IsAFile,
+                else => {
+                    log.err("Error initialising the entry iterator. Error: {}\n", .{e});
+                    return vfs.Error.Unexpected;
+                },
+            }) |entry| {
+                defer entry.deinit();
+                // File name compare is case insensitive
```
SamTebbs33, line 1249: I think file names should be case sensitive, just like on Linux.

src/kernel/filesystem/fat32.zig
```diff
@@ -1247,6 +1308,7 @@ pub fn Fat32FS(comptime StreamType: type) type {
             self.opened_files.deinit();
             self.allocator.destroy(self.root_node.node);
             self.allocator.destroy(self.fs);
+            self.allocator.destroy(self);
```
SamTebbs33, line 1312: It would be good to specify that this function destroys the pointer as well. Alternatively you could rename `init` to `create` and `deinit` to `destroy` to follow the convention from other structs.

I think option two is best.

src/kernel/filesystem/fat32.zig
```diff
@@ -1927,18 +1988,52 @@ test "Fat32FS.write" {
...
+    };
+
+    // Goto root dir and write a long and short entry
+    const sector = test_fs.fat_config.clusterToSector(test_fs.root_node.cluster);
+    try test_fs.stream.seekableStream().seekTo(sector * test_fs.fat_config.bytes_per_sector);
+    try test_fs.stream.writer().writeAll(entry_buff[0..]);
+
+    vfs.setRoot(test_fs.root_node.node);
+
+    const file = try vfs.openFile("/ramdisk_test1.txt", .NO_CREATION);
```
SamTebbs33, line 2031: Does `file` need to be checked at all? Such as it pointing to the correct file entry (like how the test fs checks that it is the same as what is in the file tree)?

SamTebbs33 (changes requested):

src/kernel/filesystem/fat32.zig
DrDeano: FAT32 specifically says that they are case insensitive. but Ext2/3/4 is case sensitive so depends on the filesystem

src/kernel/filesystem/fat32.zig
DrDeano: can add extra tests

src/kernel/filesystem/fat32.zig
SamTebbs33: Oh I see, I didn't realise this sort of thing was part of the spec.

src/kernel/filesystem/fat32.zig
DrDeano: I think it should still be `init` but have added to the doc comment.

SamTebbs33: How come you kept it as `init`?

DrDeano: We haven't really been using `create` for initialising structs (apart from task). So would have to change all the inits to creates. Also, I don't really know what the semantic difference is, aren't they basically the same meaning anyway?

SamTebbs33: I thought the difference between `create` and `init` was that the former allocates on the heap and the latter allocates on the stack. We wouldn't have to change all `init`s to `create`s because most allocate on the stack rather than heap (which is the preferred way because it's more flexible). I personally think that we should keep this convention here to make it very clear if the result is stack allocated or heap allocated, as otherwise we could run into seg-faults and double-frees because we forget if a `deinit` function frees the pointer or not.

DrDeano: That makes sense :)

SamTebbs33 (approved):
Boom boom

Issue #224, opened 2020-08-16
Resource cleanup


SamTebbs33:
Many functions in the kernel allocate resources that are not cleaned up in the case of an error. We should analyse the source and fix where this happens, adding tests as appropriate to make sure this is maintained.

ghost:
Hi, is this being worked on currently ? If yes then I would like to participate, if not then I will try

DrDeano:
Sure, you have ago. So yeh, anything that allocates memory will need leak checking for all error cases. There are examples, e.g. in the initrd.zig tests, that checks for memory leaks for each error condition. So we would like to ensure this testing is consistent through the code base.

DrDeano:
Might also be interesting to add a function that can do this kinds of testing where it will automatically test each failed allocation. Basically how the Zig fmt testing works when testing its parsing and rendering.

PR #273, opened 2020-12-02
Create a task from elf data

### LINKED ISSUE #230: Load and execute ELF files

SamTebbs33:
This PR adds the ability create a task from a loaded ELF file. It also adds the capability to allocate at a specific address to the PMM and VMM. Closes #230﻿

### DIFF
--- a/src/kernel/arch/x86/paging.zig
+++ b/src/kernel/arch/x86/paging.zig
@@ -222,10 +222,8 @@ fn mapDirEntry(dir: *Directory, virt_start: usize, virt_end: usize, phys_start: 
         // Create a table and put the physical address in the dir entry
         table = &(try allocator.alignedAlloc(Table, @truncate(u29, PAGE_SIZE_4KB), 1))[0];
         @memset(@ptrCast([*]u8, table), 0, @sizeOf(Table));
-        const table_phys_addr = vmm.kernel_vmm.virtToPhys(@ptrToInt(table)) catch |e| blk: {
-            // When testing this will fail, but that's ok
-            if (!is_test) panic(@errorReturnTrace(), "Failed getting the physical address for a page table: {}\n", .{e});
-            break :blk 0;
+        const table_phys_addr = if (builtin.is_test) @ptrToInt(table) else vmm.kernel_vmm.virtToPhys(@ptrToInt(table)) catch |e| {
+            panic(@errorReturnTrace(), "Failed getting the physical address for a page table: {}\n", .{e});
         };
         dir_entry.* |= DENTRY_PAGE_ADDR & table_phys_addr;
         dir.tables[entry] = table;
@@ -262,7 +260,7 @@ fn mapDirEntry(dir: *Directory, virt_start: usize, virt_end: usize, phys_start: 
         phys += PAGE_SIZE_4KB;
         tentry += 1;
     }) {
-        try mapTableEntry(&table.entries[tentry], phys, attrs);
+        try mapTableEntry(dir, &table.entries[tentry], virt, phys, attrs);
     }
 }
 
@@ -286,6 +284,13 @@ fn unmapDirEntry(dir: *Directory, virt_start: usize, virt_end: usize, allocator:
         var table_entry = &table.entries[virtToTableEntryIdx(addr)];
         if (table_entry.* & TENTRY_PRESENT != 0) {
             clearAttribute(table_entry, TENTRY_PRESENT);
+            if (dir == &kernel_directory) {
+                asm volatile ("invlpg (%[addr])"
+                    :
+                    : [addr] "r" (addr)
+                    : "memory"
+                );
+            }
         } else {
             return vmm.MapperError.NotMapped;
         }
@@ -297,13 +302,17 @@ fn unmapDirEntry(dir: *Directory, virt_start: usize, virt_end: usize, allocator:
 /// Sets the entry to be present, writable, kernel access, write through, cache enabled, non-global and the page address bits.
 ///
 /// Arguments:
+///     IN dir: *const Directory - The directory that is being mapped within.
+///                           The function checks if this is the kernel directory and if so invalidates the page being mapped so the TLB reloads it.
 ///     OUT entry: *align(1) TableEntry - The entry to map. 1 byte aligned.
+///     IN virt_addr: usize - The virtual address that this table entry is responsible for.
+///                           Used to invalidate the page if mapping within the kernel page directory.
 ///     IN phys_addr: usize - The physical address to map the table entry to.
 ///
 /// Error: PagingError
 ///     PagingError.UnalignedPhysAddresses - If the physical address isn't page size aligned.
 ///
-fn mapTableEntry(entry: *align(1) TableEntry, phys_addr: usize, attrs: vmm.Attributes) vmm.MapperError!void {
+fn mapTableEntry(dir: *const Directory, entry: *align(1) TableEntry, virt_addr: usize, phys_addr: usize, attrs: vmm.Attributes) vmm.MapperError!void {
     if (!std.mem.isAligned(phys_addr, PAGE_SIZE_4KB)) {
         return vmm.MapperError.MisalignedPhysicalAddress;
     }
@@ -318,18 +327,24 @@ fn mapTableEntry(entry: *align(1) TableEntry, phys_addr: usize, attrs: vmm.Attri
     } else {
         setAttribute(entry, TENTRY_USER);
     }
-    if (attrs.writable) {
-        setAttribute(entry, TENTRY_WRITE_THROUGH);
-    } else {
-        clearAttribute(entry, TENTRY_WRITE_THROUGH);
-    }
+
     if (attrs.cachable) {
+        clearAttribute(entry, TENTRY_WRITE_THROUGH);
         clearAttribute(entry, TENTRY_CACHE_DISABLED);
     } else {
+        setAttribute(entry, TENTRY_WRITE_THROUGH);
         setAttribute(entry, TENTRY_CACHE_DISABLED);
     }
+
     clearAttribute(entry, TENTRY_GLOBAL);
     setAttribute(entry, TENTRY_PAGE_ADDR & phys_addr);
+    if (dir == &kernel_directory) {
+        asm volatile ("invlpg (%[addr])"
+            :
+            : [addr] "r" (virt_addr)
+            : "memory"
+        );
+    }
 }
 
 ///
--- a/src/kernel/bitmap.zig
+++ b/src/kernel/bitmap.zig
@@ -89,11 +89,13 @@ pub fn ComptimeBitmap(comptime BitmapType: type) type {
         /// Arguments:
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?IndexType - The entry number to start allocate from or null if it can start anywhere
         ///
         /// Return: ?IndexType
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
         ///
-        pub fn setContiguous(self: *Self, num: usize) ?IndexType {
+        pub fn setContiguous(self: *Self, num: usize, from: ?IndexType) ?IndexType {
             if (num > self.num_free_entries) {
                 return null;
             }
@@ -101,7 +103,7 @@ pub fn ComptimeBitmap(comptime BitmapType: type) type {
             var count: usize = 0;
             var start: ?IndexType = null;
 
-            var bit: IndexType = 0;
+            var bit = from orelse 0;
             while (true) {
                 const entry = bit;
                 if (entry >= NUM_ENTRIES) {
@@ -111,6 +113,11 @@ pub fn ComptimeBitmap(comptime BitmapType: type) type {
                     // This is a one so clear the progress
                     count = 0;
                     start = null;
+                    // If the caller requested the allocation to start from
+                    // a specific entry and it failed then return null
+                    if (from) |_| {
+                        return null;
+                    }
                 } else {
                     // It's a zero so increment the count
                     count += 1;
@@ -334,19 +341,26 @@ pub fn Bitmap(comptime BitmapType: type) type {
         /// Arguments:
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?usize - The entry number to allocate from or null if it can start anywhere
         ///
         /// Return: ?usize
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
         ///
-        pub fn setContiguous(self: *Self, num: usize) ?usize {
+        pub fn setContiguous(self: *Self, num: usize, from: ?usize) ?usize {
             if (num > self.num_free_entries) {
                 return null;
             }
 
             var count: usize = 0;
-            var start: ?usize = null;
-            for (self.bitmaps) |bmp, i| {
-                var bit: IndexType = 0;
+            var start: ?usize = from;
+            var i: usize = if (from) |f| f / ENTRIES_PER_BITMAP else 0;
+            var bit: IndexType = if (from) |f| @truncate(IndexType, f % ENTRIES_PER_BITMAP) else 0;
+            while (i < self.bitmaps.len) : ({
+                i += 1;
+                bit = 0;
+            }) {
+                var bmp = self.bitmaps[i];
                 while (true) {
                     const entry = bit + i * ENTRIES_PER_BITMAP;
                     if (entry >= self.num_entries) {
@@ -356,6 +370,11 @@ pub fn Bitmap(comptime BitmapType: type) type {
                         // This is a one so clear the progress
                         count = 0;
                         start = null;
+                        // If the caller requested the allocation to start from
+                        // a specific entry and it failed then return null
+                        if (from) |_| {
+                            return null;
+                        }
                     } else {
                         // It's a zero so increment the count
                         count += 1;
@@ -383,10 +402,10 @@ pub fn Bitmap(comptime BitmapType: type) type {
 
             if (count == num) {
                 if (start) |start_entry| {
-                    var i: usize = 0;
-                    while (i < num) : (i += 1) {
+                    var j: usize = 0;
+                    while (j < num) : (j += 1) {
                         // Can't fail as the entry was found to be free
-                        self.setEntry(start_entry + i) catch unreachable;
+                        self.setEntry(start_entry + j) catch unreachable;
                     }
                     return start_entry;
                 }
@@ -547,24 +566,40 @@ test "Comptime indexToBit" {
 }
 
 test "Comptime setContiguous" {
-    var bmp = ComptimeBitmap(u15).init();
+    var bmp = ComptimeBitmap(u16).init();
     // Test trying to set more entries than the bitmap has
-    testing.expectEqual(bmp.setContiguous(ComptimeBitmap(u15).NUM_ENTRIES + 1), null);
+    testing.expectEqual(bmp.setContiguous(bmp.num_free_entries + 1, null), null);
+    testing.expectEqual(bmp.setContiguous(bmp.num_free_entries + 1, 1), null);
     // All entries should still be free
-    testing.expectEqual(bmp.num_free_entries, ComptimeBitmap(u15).NUM_ENTRIES);
-    testing.expectEqual(bmp.setContiguous(3) orelse unreachable, 0);
-    testing.expectEqual(bmp.setContiguous(4) orelse unreachable, 3);
-    // 0b0000.0000.0111.1111
-    bmp.bitmap |= 0x200;
-    // 0b0000.0010.0111.1111
-    testing.expectEqual(bmp.setContiguous(3) orelse unreachable, 10);
-    // 0b0001.1110.0111.1111
-    testing.expectEqual(bmp.setContiguous(5), null);
-    testing.expectEqual(bmp.setContiguous(2), 7);
-    // 0b001.1111.1111.1111
-    // Test trying to set beyond the end of the bitmaps
-    testing.expectEqual(bmp.setContiguous(3), null);
-    testing.expectEqual(bmp.setContiguous(2), 13);
+    testing.expectEqual(bmp.num_free_entries, 16);
+    testing.expectEqual(bmp.bitmap, 0b0000000000000000);
+
+    testing.expectEqual(bmp.setContiguous(3, 0) orelse unreachable, 0);
+    testing.expectEqual(bmp.bitmap, 0b0000000000000111);
+
+    // Test setting from top
+    testing.expectEqual(bmp.setContiguous(2, 14) orelse unreachable, 14);
+    testing.expectEqual(bmp.bitmap, 0b1100000000000111);
+
+    testing.expectEqual(bmp.setContiguous(3, 12), null);
+    testing.expectEqual(bmp.bitmap, 0b1100000000000111);
+
+    testing.expectEqual(bmp.setContiguous(3, null) orelse unreachable, 3);
+    testing.expectEqual(bmp.bitmap, 0b1100000000111111);
+
+    // Test setting beyond the what is available
+    testing.expectEqual(bmp.setContiguous(9, null), null);
+    testing.expectEqual(bmp.bitmap, 0b1100000000111111);
+
+    testing.expectEqual(bmp.setContiguous(8, null) orelse unreachable, 6);
+    testing.expectEqual(bmp.bitmap, 0b1111111111111111);
+
+    // No more are possible
+    testing.expectEqual(bmp.setContiguous(1, null), null);
+    testing.expectEqual(bmp.bitmap, 0b1111111111111111);
+
+    testing.expectEqual(bmp.setContiguous(1, 0), null);
+    testing.expectEqual(bmp.bitmap, 0b1111111111111111);
 }
 
 test "setEntry" {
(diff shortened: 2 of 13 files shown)


src/kernel/vmm.zig
```diff
@@ -389,14 +390,15 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
         /// Error: Allocator.Error
         ///     error.OutOfMemory: The required amount of memory couldn't be allocated
         ///
-        pub fn alloc(self: *Self, num: usize, attrs: Attributes) Allocator.Error!?usize {
+        pub fn alloc(self: *Self, num: usize, virtual_addr: ?usize, attrs: Attributes) Allocator.Error!?usize {
```
DrDeano, line 394: Could add some tests for allocating at a specific address

test/user_program.s
```diff
@@ -1,3 +1,4 @@
+.section .text
```
DrDeano, line 1: Could add an extra test file that has a data section and read and write from that.
```asm
.section .test
entry:
    mov item, %eax

.section .data
    item: .long 1234
```

src/kernel/bitmap.zig
```diff
@@ -547,24 +566,40 @@ test "Comptime indexToBit" {
 }
 
 test "Comptime setContiguous" {
-    var bmp = ComptimeBitmap(u15).init();
+    var bmp = ComptimeBitmap(u16).init();
```
DrDeano, line 569: How come this is now a u16?

src/kernel/task.zig
```diff
@@ -99,6 +104,25 @@ pub const Task = struct {
...
+        errdefer task.destroy(allocator);
+
+        // Iterate over sections
+        for (program_elf.section_headers) |header, i| {
+            if ((header.flags & elf.SECTION_ALLOCATABLE) == 0) {
+                continue;
+            }
+            // If it is loadable then allocate it at its virtual address
+            const attrs = vmm.Attributes{ .kernel = kernel, .writable = (header.flags & elf.SECTION_WRITABLE) != 0, .cachable = true };
+            const vaddr = (try task_vmm.alloc(std.mem.alignForward(header.size, vmm.BLOCK_SIZE) / vmm.BLOCK_SIZE, header.virtual_address, attrs)) orelse return if (try task_vmm.isSet(header.virtual_address)) vmm.VmmError.AlreadyAllocated else bitmap.Bitmap(usize).BitmapError.OutOfBounds;
```
DrDeano: This is a very long line, could this be refactored

src/kernel/task.zig
```diff
@@ -99,6 +104,25 @@ pub const Task = struct {
         return task;
     }
 
+    pub fn createFromElf(program_elf: elf.Elf, kernel: bool, task_vmm: *vmm.VirtualMemoryManager(arch.VmmPayload), allocator: *Allocator) (bitmap.Bitmap(usize).BitmapError || vmm.VmmError || elf.Error || Allocator.Error)!*Task {
+        const task = try create(program_elf.header.entry_address, kernel, task_vmm, allocator);
+        errdefer task.destroy(allocator);
+
+        // Iterate over sections
+        for (program_elf.section_headers) |header, i| {
```
DrDeano: If there are multiple headers, lets say 4, and the in the 3rd loop the vaddr allocation fails, then there will still be dangling addresses that will need to be freed. So will need a keep track of all the vaddrs and free then if there is a failure. Could also add a test for this.

src/kernel/scheduler.zig
```diff
@@ -375,32 +376,36 @@ fn rt_user_task(allocator: *Allocator, mem_profile: *const mem.MemProfile) void
...
-    // 4. Allocate space in the vmm for the user_program
-    const code_start = task_vmm.alloc(std.mem.alignForward(code_len, vmm.BLOCK_SIZE) / vmm.BLOCK_SIZE, .{ .kernel = false, .writable = true, .cachable = true }) catch |e| {
-        panic(@errorReturnTrace(), "Failed to allocate VMM memory for user program code: {}\n", .{e});
-    } orelse panic(null, "User task VMM didn't allocate space for the user program\n", .{});
-    if (code_start != 0) panic(null, "User program start address was {} instead of 0\n", .{code_start});
-    // 5. Copy user_program code over
-    vmm.kernel_vmm.copyData(task_vmm, code[0..code_len], code_start, true) catch |e| {
-        panic(@errorReturnTrace(), "Failed to copy user code: {}\n", .{e});
+    const program_elf = elf.Elf.init(code[0..code_len], builtin.arch, allocator) catch |e| panic(@errorReturnTrace(), "Failed to load user program elf: {}\n", .{e});
+    errdefer program_elf.deinit();
```
DrDeano: Wouldn't this need to be a `defer` instead?

test/user_program.s
SamTebbs33: And this is finally finished and working :)

src/kernel/bitmap.zig
SamTebbs33: I believe the reason was that it gave me more bits to manipulate and test with.

src/kernel/task.zig
SamTebbs33: Well spotted

src/kernel/bitmap.zig
DrDeano: Was just wondering, but cool.

src/kernel/scheduler.zig
SamTebbs33: Indeed :)

src/kernel/task.zig
```diff
@@ -113,10 +113,20 @@ pub const Task = struct {
...
             }
+            errdefer {
+                // Free the previously allocated addresses
+                for (program_elf.section_headers) |header2, j| {
+                    if (j >= i)
+                        break;
+                    if ((header2.flags & elf.SECTION_ALLOCATABLE) != 0)
+                        task_vmm.free(header.virtual_address) catch |e| panic(null, "VMM failed to clean up a previously-allocated address after an error: {}\n", .{e});
+                }
+            }
```
DrDeano: Could this be moved outside the for loop. Also could use a while loop to avoid the `if (j >= i)`
Example in initrd.zig `init` function

src/kernel/bitmap.zig
```diff
@@ -89,19 +89,21 @@ pub fn ComptimeBitmap(comptime BitmapType: type) type {
...
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?usize - The entry number to start allocate from or null if it can start anywhere
         ///
         /// Return: ?IndexType
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
         ///
-        pub fn setContiguous(self: *Self, num: usize) ?IndexType {
+        pub fn setContiguous(self: *Self, num: usize, from: ?usize) ?IndexType {
```
DrDeano: Can't the type be `?IndexType` so could avoid the `@intCast`

src/kernel/bitmap.zig
```diff
@@ -334,19 +341,26 @@ pub fn Bitmap(comptime BitmapType: type) type {
...
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?usize - The entry number to allocate from or null if it can start anywhere
         ///
         /// Return: ?usize
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
         ///
-        pub fn setContiguous(self: *Self, num: usize) ?usize {
+        pub fn setContiguous(self: *Self, num: usize, from: ?usize) ?usize {
```
DrDeano, line 350: Same here, `?IndexType`

test/mock/kernel/arch_mock.zig
```diff
@@ -54,11 +55,18 @@ pub const CpuState = struct {
...
-pub const KERNEL_VMM_PAYLOAD: usize = 0;
+pub const VmmPayload = switch (builtin.arch) {
+    .i386 => *@import("../../../src/kernel/arch/x86/paging.zig").Directory,
+    else => unreachable,
+};
+
+pub const KERNEL_VMM_PAYLOAD: VmmPayload = switch (builtin.arch) {
+    .i386 => &@import("../../../src/kernel/arch/x86/paging.zig").kernel_directory,
+    else => unreachable,
+};
```
DrDeano: Could import paging at the top level instead of importing twice

test/user_program_data.s
```diff
@@ -0,0 +1,11 @@
...
+.globl entry
+entry:
+    mov item1, %eax
+    mov item2, %ebx
+loop:
+    jmp loop
+
+.section .data
+item1: .long 0xCAFE
+item2: .long 0xBEEF
```
DrDeano, line 11: This doesn't have to be for this PR, can make an issue for future testing
Could we have another program that writes.
Could try to write to a .rodata section and make sure the protections are there.

src/kernel/task.zig
```diff
@@ -5,11 +5,16 @@ const builtin = @import("builtin");
 const is_test = builtin.is_test;
 const build_options = @import("build_options");
 const mock_path = build_options.mock_path;
-const arch = @import("arch.zig").internals;
+const arch = if (is_test) @import(mock_path ++ "arch_mock.zig") else @import("arch.zig").internals;
```
DrDeano: The mocked arch file is imported already in the main arch.zig file

src/kernel/task.zig
```diff
@@ -5,11 +5,16 @@ const builtin = @import("builtin");
...
 const panic = @import("panic.zig").panic;
 const ComptimeBitmap = @import("bitmap.zig").ComptimeBitmap;
 const vmm = @import("vmm.zig");
+const pmm = @import("pmm.zig");
+const mem = @import("mem.zig");
+const elf = @import("elf.zig");
+const bitmap = @import("bitmap.zig");
```
DrDeano, line 14: Bitmap is imported twice now, can be combined

src/kernel/task.zig
```diff
@@ -254,3 +290,39 @@ test "allocatePid and freePid" {
 
     expectEqual(all_pids.bitmap, 0);
 }
+
```
DrDeano, line 294: Could have a test to go over all the errors to ensure the memory cleanup.
#224 😉

src/kernel/task.zig
```diff
@@ -99,6 +104,37 @@ pub const Task = struct {
...
+                    if ((header2.flags & elf.SECTION_ALLOCATABLE) != 0)
+                        task_vmm.free(header.virtual_address) catch |e| panic(null, "VMM failed to clean up a previously-allocated address after an error: {}\n", .{e});
+                }
+            }
+            // If it is loadable then allocate it at its virtual address
+            const attrs = vmm.Attributes{ .kernel = kernel, .writable = (header.flags & elf.SECTION_WRITABLE) != 0, .cachable = true };
+            const vmm_blocks = std.mem.alignForward(header.size, vmm.BLOCK_SIZE) / vmm.BLOCK_SIZE;
+            const vaddr_opt = try task_vmm.alloc(vmm_blocks, header.virtual_address, attrs);
+            const vaddr = vaddr_opt orelse return if (try task_vmm.isSet(header.virtual_address)) error.AlreadyAllocated else error.OutOfBounds;
+            if (vaddr != header.virtual_address) panic(null, "VMM didn't allocate at correct address. Expected 0x{X} but got 0x{X}\n", .{ header.virtual_address, vaddr });
```
DrDeano: Is this possible as the alloc would fail/return null if it can't alloc at that address

src/kernel/arch/x86/interrupts.zig
```diff
@@ -49,11 +49,7 @@ export fn commonStub() callconv(.Naked) void {
     // If so don't change cr3 to avoid a TLB flush
```
DrDeano: Is this comment still relavant, should the cr3 check need to be re-added.

src/kernel/scheduler.zig
```diff
@@ -352,37 +354,50 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
-    // 6. Schedule it
-    scheduleTask(user_task, allocator) catch |e| {
-        panic(@errorReturnTrace(), "Failed to schedule the user task: {}\n", .{e});
-    };
+    for (&[_][]const u8{ "user_program_data.elf", "user_program.elf" }) |user_program| {
+        // 1. Create user VMM
+        var task_vmm = allocator.create(vmm.VirtualMemoryManager(arch.VmmPayload)) catch |e| {
+            panic(@errorReturnTrace(), "Failed to allocate VMM for {s}: {}\n", .{ user_program, e });
+        };
+        task_vmm.* = vmm.VirtualMemoryManager(arch.VmmPayload).init(0, @ptrToInt(mem_profile.vaddr_start), allocator, arch.VMM_MAPPER, undefined) catch unreachable;
```
DrDeano: Is this a `catch unreachable` 👀

src/kernel/vmm.zig
```diff
@@ -460,35 +462,39 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
...
+                //     const v_end = v_start + blocks.items.len * BLOCK_SIZE;
+                //     var v = v_start;
+                //     while (v < v_end) : (v += BLOCK_SIZE) {
+                //         // Cannot be out of bounds as it has been set above
+                //         self.bmp.clearEntry((v - self.start) / BLOCK_SIZE) catch unreachable;
+                //     }
+                //     self.mapper.unmapFn(v_start, v_end, self.allocator, self.payload) catch |e| {
+                //         panic(@errorReturnTrace(), "Failed to unmap virtual region 0x{X} -> 0x{X}: {}\n", .{ v_start, v_end, e });
+                //     };
+                // }
```
DrDeano: Commented code

src/kernel/scheduler.zig
```diff
@@ -352,37 +354,50 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
+        var user_task = task.Task.createFromElf(program_elf, false, task_vmm, allocator) catch |e| {
+            panic(@errorReturnTrace(), "Failed to create task for {s}: {}\n", .{ user_program, e });
+        };
+        // 6. Schedule it
+        scheduleTask(user_task, allocator) catch |e| {
+            panic(@errorReturnTrace(), "Failed to schedule the task for {s}: {}\n", .{ user_program, e });
+        };
+
+        var num_allocatable_sections: usize = 0;
+        var size_allocatable_sections: usize = 0;
```
DrDeano, line 386: This variable isn't used

DrDeano (commented):
Some more observation

src/kernel/task.zig
SamTebbs33: Why do you think it should be moved outside the loop? For loops are made for this kind of iteration, as they provide the iteration count and slice element without any extra lines of code.

src/kernel/bitmap.zig
SamTebbs33: Good idea. The `@truncate` will still be needed though as `%` produces a usize.

SamTebbs33: Actually, this can't be an index type as it corresponds to the *entry* to start at, rather than the *bit* to start at.

test/user_program_data.s
SamTebbs33: Good idea. I'll make an issue post-merge

src/kernel/task.zig
SamTebbs33: True :+1:

src/kernel/arch/x86/interrupts.zig
SamTebbs33: Yeah it should be re-added.

src/kernel/scheduler.zig
SamTebbs33: No.... xD

src/kernel/task.zig
SamTebbs33: If you mean heap memory then I will add an extra `testing.expectError` call for each error that the function can return. If you mean VMM memory, then I have been thinking about that and would like to implement VMM memory leak detection like the testing allocator has, but I think that is outside of the scope of this PR.

src/kernel/task.zig
SamTebbs33: I have thought about this and realised that it would be better outside of the loop :+1:

src/kernel/arch/x86/paging.zig
```diff
@@ -318,18 +327,23 @@ fn mapTableEntry(entry: *align(1) TableEntry, phys_addr: usize, attrs: vmm.Attri
...
         setAttribute(entry, TENTRY_USER);
     }
-    if (attrs.writable) {
-        setAttribute(entry, TENTRY_WRITE_THROUGH);
-    } else {
-        clearAttribute(entry, TENTRY_WRITE_THROUGH);
-    }
+
+    clearAttribute(entry, TENTRY_WRITE_THROUGH);
+
```
DrDeano: What happened to the writable attribute?

src/kernel/bitmap.zig
```diff
@@ -89,19 +89,21 @@ pub fn ComptimeBitmap(comptime BitmapType: type) type {
         /// Arguments:
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?IndexType - The entry number to start allocate from or null if it can start anywhere
         ///
         /// Return: ?IndexType
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
```
DrDeano, line 96: I would say that this comment would belong in the main description. But up to you.

src/kernel/bitmap.zig
```diff
@@ -334,19 +341,26 @@ pub fn Bitmap(comptime BitmapType: type) type {
         /// Arguments:
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?usize - The entry number to allocate from or null if it can start anywhere
         ///
         /// Return: ?usize
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
```
DrDeano, line 348: Same here.

src/kernel/bitmap.zig
```diff
@@ -89,19 +89,21 @@ pub fn ComptimeBitmap(comptime BitmapType: type) type {
...
         ///     IN/OUT self: *Self - The bitmap to modify.
         ///     IN num: usize - The number of entries to set.
+        ///     IN from: ?IndexType - The entry number to start allocate from or null if it can start anywhere
         ///
         /// Return: ?IndexType
         ///     The first entry set or null if there weren't enough contiguous entries.
+        ///     If `from` was not null and any entry between `from` and `from` + num is set then null is returned.
         ///
-        pub fn setContiguous(self: *Self, num: usize) ?IndexType {
+        pub fn setContiguous(self: *Self, num: usize, from: ?IndexType) ?IndexType {
```
DrDeano, line 98: From the previous comments, you said that `from` is the entry to start from and not the bit index, so this should be `?usize`

src/kernel/elf.zig
```diff
@@ -542,6 +542,9 @@ fn testInitData(section_name: []const u8, string_section_name: []const u8, file_
     testSetSection(data, string_section_header, 1);
     data_offset += s_header_size;
 
+    std.mem.copy(u8, data[data_offset .. data_offset + section_size], &[_]u8{0} ** section_size);
```
DrDeano: Could use `std.mem.set(u8, data[data_offset .. data_offset + section_size], 0);`

src/kernel/elf.zig
```diff
@@ -639,7 +646,7 @@ test "getName" {
     // The entire ELF test data. The header, program header, two section headers and the section name (with the null terminator)
     var section_name = "some_section";
     var string_section_name = "strings";
-    const data = testInitData(section_name, string_section_name, .Executable, 0, undefined, undefined, undefined, undefined, undefined);
+    const data = try testInitData(testing.allocator, section_name, string_section_name, .Executable, 0, undefined, undefined, undefined, undefined, undefined);
```
DrDeano, line 649: Not sure it is a good idea to use `undefined` as parameters, might be better to use `0xAA` directly so it is more defined.

src/kernel/scheduler.zig
```diff
@@ -304,7 +306,7 @@ fn rt_variable_preserved(allocator: *Allocator) void {
     defer allocator.destroy(is_set);
     is_set.* = true;
 
-    var test_task = Task.create(@ptrToInt(task_function), true, undefined, allocator) catch unreachable;
+    var test_task = Task.create(@ptrToInt(task_function), true, &vmm.kernel_vmm, allocator) catch unreachable;
```
DrDeano: Could these `catch undefined` be made into `panic` calls.

src/kernel/scheduler.zig
```diff
@@ -352,37 +354,55 @@ fn rt_variable_preserved(allocator: *Allocator) void {
...
-    if (code_start != 0) panic(null, "User program start address was {} instead of 0\n", .{code_start});
-    // 5. Copy user_program code over
-    vmm.kernel_vmm.copyData(task_vmm, code[0..code_len], code_start, true) catch |e| {
-        panic(@errorReturnTrace(), "Failed to copy user code: {}\n", .{e});
-    };
-    // 6. Schedule it
-    scheduleTask(user_task, allocator) catch |e| {
-        panic(@errorReturnTrace(), "Failed to schedule the user task: {}\n", .{e});
-    };
+    for (&[_][]const u8{ "user_program_data.elf", "user_program.elf" }) |user_program| {
```
DrDeano: Why not include the `/` in the paths so you don't need the `std.mem.join`?

src/kernel/task.zig
```diff
@@ -254,3 +291,45 @@ test "allocatePid and freePid" {
...
+
+    // Test errdefer clean-up by fillng up all but one block in the VMM so allocating the last section fails
+    // The allocation for the first section should be cleaned up in case of an error
+    const available_address = (try the_vmm.alloc(1, null, .{ .writable = false, .kernel = false, .cachable = false })) orelse unreachable;
+    the_elf.section_headers[0].virtual_address = available_address;
+    _ = try the_vmm.alloc(the_vmm.bmp.num_free_entries, null, .{ .kernel = false, .writable = false, .cachable = false });
+    try the_vmm.free(available_address);
+    // Make the strings section allocatable so createFromElf tries to allocate more than one
+    the_elf.section_headers[1].flags |= elf.SECTION_ALLOCATABLE;
+    std.testing.expectError(error.AlreadyAllocated, Task.createFromElf(the_elf, true, &the_vmm, std.testing.allocator));
```
DrDeano, line 351: I think this should be refactored as a separate test

src/kernel/task.zig
```diff
@@ -99,6 +104,38 @@ pub const Task = struct {
...
+            const header = program_elf.section_headers[i];
+            if ((header.flags & elf.SECTION_ALLOCATABLE) == 0) {
+                continue;
+            }
+            // If it is loadable then allocate it at its virtual address
+            const attrs = vmm.Attributes{ .kernel = kernel, .writable = (header.flags & elf.SECTION_WRITABLE) != 0, .cachable = true };
+            const vmm_blocks = std.mem.alignForward(header.size, vmm.BLOCK_SIZE) / vmm.BLOCK_SIZE;
+            const vaddr_opt = try task_vmm.alloc(vmm_blocks, header.virtual_address, attrs);
+            const vaddr = vaddr_opt orelse return if (try task_vmm.isSet(header.virtual_address)) error.AlreadyAllocated else error.OutOfBounds;
+            errdefer task_vmm.free(vaddr) catch |e| panic(@errorReturnTrace(), "Failed to free VMM memory in createFromElf: {}\n", .{e});
```
DrDeano, line 132: This `errdefer` is what the larger `errdefer` does, so this can be removed. Else you will have a double free, but given the tests pass, then this double free is never hit, so either this is needed and I am dumb or there might need to be another test to check the cleanup.

DrDeano (commented):
I promise these are the last comments lol

src/kernel/arch/x86/paging.zig
SamTebbs33: Looks like I misunderstood what write-through meant when I first wrote the paging code. When set, it means that changes are written directly to memory instead of being cached and written at a later time, so this should depend on the `cachable` attribute instead.

src/kernel/bitmap.zig
SamTebbs33: But this is in the comptime bitmap where an entry in the bitmap is denoted with an IndexType.

Comptime bitmap: ++++++++++++++++++++++++++++++++++++++++
_______________|-----------------^--------------------------------------------------|
                                              IndexType

Normal bitmap: [++++++++][++++++++][++++++++][++++++++][++++++++]
IndexType: _____________|----^--------|
usize: _______|---------------------^--------------------------------------------------------|

An IndexType in the normal bitmap refers to an index within one of the array elements but refers to an index within the whole bitmap in the comptime one (underscores are just for padding).

src/kernel/elf.zig
SamTebbs33: Nice

src/kernel/elf.zig
SamTebbs33: The point of `undefined` is that we don't care about the value. Using `0xAA` would imply that we care about the value.

src/kernel/scheduler.zig
SamTebbs33: Yarp

src/kernel/scheduler.zig
SamTebbs33: Sure ting

src/kernel/task.zig
SamTebbs33: The double free is prevented with the `j >= i` check. The loop body can error *before* the virtual address is allocated (it can error on line 130) and so if this was merged with the errdefer block above it would try to free even though the address hadn't been allocated. I guess I could add an `if (task_vmm.isSet(...))` check to the errdefer block.

DrDeano: Nah, I get you :)

DrDeano (approved):
Sorry for the long wait, but bung it in 🚀 
Also, there was no comment on the changes for the doc comment, but as it mentions 'return' then is probably fine.

SamTebbs33:
> Sorry for the long wait, but bung it in 🚀
> Also, there was no comment on the changes for the doc comment, but as it mentions 'return' then is probably fine.

Sweeeeeet :) Nice one

PR #274, opened 2020-12-07
Remove unneeded mocking files

SamTebbs33:
I have experienced situations where some functionality I've added depends on the proper functionality of another file, such as vmm.zig, but every other file (such as task.zig) expects to get a mocked vmm, which doesn't implement the actual functionality. This isn't a great situation because it either means the mocking files have to implement the same functionality as the normal files (i.e. they become pointless) or we cannot fully test new features.

This PR removes all mocking files that I see as being unnecessary so that we can test functionality properly. The only mocking files I have left are arch-specific ones because they emulate operations with the CPU.﻿


DrDeano (approved):
🚀

PR #275, opened 2020-12-10
Added read functionality for FAT32

DrDeano:
Create test FAT32 image on build
Moved test files to test_files directory
Removed global array for test fat32
Part of #216


src/kernel/filesystem/fat32.zig
```diff
@@ -3837,87 +3753,195 @@ test "Fat32FS.openImpl - entry iterator failed 2nd next" {
...
+        while (j < 26) : (j += 1) {
+            // Each character repeated 64 times
+            var k: usize = 0;
+            while (k < 64) : (k += 1) {
+                expectEqual(buff[(i * 65 * 26) + (j * 65) + k], 'A' + @intCast(u8, j));
+            }
+            // New line
+            expectEqual(buff[(i * 65 * 26) + (j * 65) + 64], '\n');
+        }
+    }
```
SamTebbs33: Instead of this, you could use `@embedFile("test/fat32/test_files/large_file2.txt")` to get a slice of the file's contents and do a slice equality check. It's up to you depending on if you think it's cleaner or not!

SamTebbs33 (approved):
Looking very good. I like the breadth of testing. I just have one small suggestion but it's up to you.

src/kernel/filesystem/fat32.zig
DrDeano: That is a good idea

PR #278, opened 2021-01-01
SymlinkNode + umount + closeing dir path

### LINKED ISSUE #277: VFS close dangling directories
When opening a path with multiple folders, the parent folders are not closed when the opened result is closed.

DrDeano:
Added a symlink node so can close a open symlink.
Added umount to can unmount a directory.
Closes open directories while traversing the path.
Closes #277

### DIFF
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -29,7 +29,7 @@ pub const Node = union(enum) {
     /// The dir node if this represents a directory
     Dir: DirNode,
     /// The absolute path that this symlink is linked to
-    Symlink: []const u8,
+    Symlink: SymlinkNode,
 
     const Self = @This();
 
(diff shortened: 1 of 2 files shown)


src/kernel/filesystem/vfs.zig
```diff
@@ -224,12 +224,28 @@ pub const DirNode = struct {
     pub fn close(self: *const DirNode) void {
         var fs = self.fs;
         var node = self;
-        if (self.mount) |mnt| {
-            fs = mnt.fs;
-            node = mnt;
+
+        // TODO: Use @fieldParentPtr() once implemented for unions
+        const cast_node = @ptrCast(*const Node, node);
+        // Can't close the root node
```
SamTebbs33, line 230: Why not?

src/kernel/filesystem/vfs.zig
```diff
@@ -372,6 +394,17 @@ pub fn mount(dir: *DirNode, fs: *const FileSystem) MountError!void {
     dir.mount = fs.getRootNode(fs);
 }
 
+///
+/// Unmount the the filesystem attached to a directory. Any opened files or directories will be
```
SamTebbs33: One too many "the" here. Why does this cause opened files within the mounted dir to be left dangling?

SamTebbs33 (commented):
Nice job and well spotted.

src/kernel/filesystem/vfs.zig
DrDeano: Because the root node is never 'opened' thought the `open` function, it just gets returned by `getRootNode` so can't be closed by `close`.

src/kernel/filesystem/vfs.zig
DrDeano: All the the's. Yes this is now lie, I has this before I changed how close works, can remove the last part of the doc comment.

src/kernel/filesystem/vfs.zig
SamTebbs33: True, we want to keep the root around at all times. I think this has exposed a possible design or mental representation issue with the VFS: we want to let a process open any node (including the root) and close it when they are done since we'd want to release the file handle etc., but if this then called the close function on the node it would close it for all other processes too, which we wouldn't want. Perhaps we need to have a think about what closing *actually* means.

DrDeano: ~~We could have it so that the underlying filesystem handle the closing of the root node so if it isn't 'opened', the the filesystem will ignore the request or do what it need to do with closing the root node if it is open.~~
Opened an issue for this #281

SamTebbs33 (approved):

PR #285, opened 2021-01-02
Feature/fat32 write

### LINKED ISSUE #216: Add the FAT32 filesystem

DrDeano:
Added FAT32 write capabilities. See commits for the journey.
This is a long and painful journey of reverse engineering and the FAT32 manual.
After many months of pain and hard work, this will closes #216

### DIFF
--- a/mkfat32.zig
+++ b/mkfat32.zig
@@ -523,8 +523,8 @@ pub const Fat32 = struct {
         }
 
         // Valid clusters are 1, 2, 4, 8, 16, 32, 64 and 128
-        if (options.cluster_size < 0 or options.cluster_size > 128 or !std.math.isPowerOfTwo(options.cluster_size)) {
-            log.err("Sectors per cluster is invalid. Must be less then 32 and a power of 2. Found: {}", .{options.cluster_size});
+        if (options.cluster_size < 1 or options.cluster_size > 128 or !std.math.isPowerOfTwo(options.cluster_size)) {
+            log.err("Sectors per cluster is invalid. Must be less then or equal to 128 and a power of 2. Found: {}", .{options.cluster_size});
             return Error.InvalidOptionValue;
         }
 
--- a/src/kernel/filesystem/fat32.zig
+++ b/src/kernel/filesystem/fat32.zig
@@ -8,6 +8,7 @@ const log = std.log.scoped(.fat32);
 const AutoHashMap = std.AutoHashMap;
 const Allocator = std.mem.Allocator;
 const ArrayList = std.ArrayList;
+const arch = @import("../arch.zig").internals;
 const vfs = @import("vfs.zig");
 const mem = @import("../mem.zig");
 const CodePage = @import("../code_page/code_page.zig").CodePage;
@@ -376,7 +377,7 @@ const ShortName = struct {
     }
 
     ///
-    /// Calculate the check sum for the long name entry.
+    /// Calculate the check sum for the short name entry.
     ///
     /// Arguments:
     ///     IN self: *const ShortName - The short entry to calculate the check sum for.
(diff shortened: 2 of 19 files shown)


mkfat32.zig
```diff
@@ -523,7 +523,7 @@ pub const Fat32 = struct {
         }
 
         // Valid clusters are 1, 2, 4, 8, 16, 32, 64 and 128
-        if (options.cluster_size < 0 or options.cluster_size > 128 or !std.math.isPowerOfTwo(options.cluster_size)) {
+        if (options.cluster_size < 1 or options.cluster_size > 128 or !std.math.isPowerOfTwo(options.cluster_size)) {
             log.err("Sectors per cluster is invalid. Must be less then 32 and a power of 2. Found: {}", .{options.cluster_size});
```
SamTebbs33: One thing I didn't notice before is that the error message says that the cluster size must be less than 32 but the condition uses 128.

src/kernel/filesystem/vfs.zig
```diff
@@ -552,7 +552,13 @@ pub fn isAbsolute(path: []const u8) bool {
 /// Arguments:
 ///     IN node: *Node - The node to initialise the root node.
 ///
-pub fn setRoot(node: *Node) void {
+/// Error: Error
+///     Error.NotADirectory - The node isn't a directory node.
+///
+pub fn setRoot(node: *Node) Error!void {
+    if (node.isFile() or node.isSymlink()) {
```
SamTebbs33: It may be best to check `!isDir()` instead.

src/kernel/filesystem/fat32.zig
```diff
@@ -441,11 +442,11 @@ const FATConfig = struct {
     /// If the filesystem has a FSInfo block.
     has_fs_info: bool,
 
-    /// The number of free clusters on the disk. Will be 0xFFFFFFFF is unknown.
+    /// The number of free clusters on the disk. Will be 0x0FFFFFFF is unknown.
     number_free_clusters: u32,
 
     /// The next free cluster to start looking for when allocating a new cluster for a
-    /// file. Will be 0xFFFFFFFF is unknown.
+    /// file. Will be 0x0FFFFFFF is unknown.
```
SamTebbs33: It's a nitpick, but the first zeroes in the hex numbers here are unnecessary

src/kernel/filesystem/fat32.zig
```diff
@@ -514,6 +515,48 @@ fn initStruct(comptime Type: type, bytes: []const u8) Type {
...
 }
 
+///
+/// Initialise a slice with the values from a struct.
+/// TODO: Once packed structs are good to go, then use std.mem.bytesAsValue.
+///
+/// Arguments:
+///     IN comptime Type: type - The type of the struct.
+///     IN copy_struct: Type   - The struct to copt from.
+///     IN bytes: []u8         - The bytes to copt to.
```
SamTebbs33: copt -> copy

src/kernel/filesystem/fat32.zig
```diff
@@ -950,7 +1007,7 @@ pub fn Fat32FS(comptime StreamType: type) type {
                     defer if (long_entries) |entries| self.allocator.free(entries);
 
                     // If attribute is 0x0F, then is a long file name and
```
SamTebbs33: extra "and" here

src/kernel/filesystem/fat32.zig
```diff
@@ -1147,43 +1204,79 @@ pub fn Fat32FS(comptime StreamType: type) type {
...
         }
 
         /// See vfs.FileSystem.open
         fn open(fs: *const vfs.FileSystem, dir: *const vfs.DirNode, name: []const u8, flags: vfs.OpenFlags, open_args: vfs.OpenArgs) (Allocator.Error || vfs.Error)!*vfs.Node {
             return switch (flags) {
                 .NO_CREATION => openImpl(fs, dir, name),
-                .CREATE_DIR, .CREATE_FILE, .CREATE_SYMLINK => createFileOrDirOrSymlink(fs, dir, name, flags, open_args),
+                .CREATE_FILE => createFileOrDir(fs, dir, name, false),
+                .CREATE_DIR => createFileOrDir(fs, dir, name, true),
+                // FAT32 doesn't support symlinks
```
SamTebbs33, line 1248: I wonder how we could keep symlinks persistent in a disk then 🤔

src/kernel/filesystem/fat32.zig
```diff
@@ -1277,34 +1377,672 @@ pub fn Fat32FS(comptime StreamType: type) type {
...
+                    sector_offset = check_offset;
+                    try seeker.seekTo((self.fat_config.reserved_sectors + sector_offset) * self.fat_config.bytes_per_sector);
+                    fat_read = try reader.readAll(std.mem.sliceAsBytes(fat_buff));
+                    if (fat_read != fat_buff.len * @sizeOf(u32)) {
+                        return Fat32Self.Error.BadRead;
+                    }
+                }
+            }
+
+            // Update the FAT for the allocated cluster
```
SamTebbs33, line 1449: From this point onwards the cluster state will be borked if any of the writes fail but I can't think of a nice way to use an errdefer to fix the state in case that happens. Could a copy of the cluster at its state before the writes be kept around and restored in case an error occurs?

SamTebbs33 (commented):
Nice work :) There are a lot of new tests and I don't have the mental rigidity to read through them all but I trust they are doing the right sort of thing ;)

mkfat32.zig
DrDeano: Ah yes 👍

src/kernel/filesystem/vfs.zig
DrDeano: Sure

src/kernel/filesystem/fat32.zig
DrDeano: Yes, was from a copy and replace.

src/kernel/filesystem/fat32.zig
DrDeano: tbh, FAT32 can get in the bin XD
This is just a starting file system. We will can use Ext2 for our main FS as that has symlinks

src/kernel/filesystem/fat32.zig
DrDeano: But what if the restore fails as will still need to write to the disk? I think caching might help, but I don't think I can have all the writes happen atomically. Plus, the write will fail if there is an issue with the disk and that would be a bigger problem.

SamTebbs33: Would one of these writes failing cause any problems for other clusters or the rest of the filesystem? If it would only affect this one cluster/file then I don't see it being corrupted as too big an issue as long as the error gets reported properly so the user knows.

DrDeano: This write would only effect either a newly created file, writing to a file that is bigger than a cluster. so could corrupt the file being worked on and nothing else (I think). The error will bubble up to who called write or open create. It would result in lost clusters, but won't effect the file system as a whole. But depending on which write failed would corrupt the file system in different ways.

DrDeano:
> 
> 
> Nice work :) There are a lot of new tests and I don't have the mental rigidity to read through them all but I trust they are doing the right sort of thing ;)

I have tried to cover all the normal use cases to ensure everything works as normal, but have probably missed a lot of error cases, but would have too many unit tests. So if there is a error that occurs, can save the corrupt FAT image and use the image as a test case

SamTebbs33 (approved):

PR #289, opened 2021-02-22
Remove constants.zig

SamTebbs33:
We only use constants.zig in x86/boot.zig as everywhere else we use the value exported by the linker﻿ (which I think is the correct approach as that keeps things in sync) so I have removed constants.zig in this PR to make things a little cleaner.


DrDeano (approved):
🚀

PR #290, opened 2021-02-22
Add allocator to paging unmap to free allocated entries

DrDeano:


SamTebbs33:
~~I don't have any problems with the change, but why not just use the allocated stored in the `Directory`?~~

Hmm I just realised that the paging stuff uses the fixed buffer allocator which can't free, and that the map functions take an allocator as well, so this looks good :)

SamTebbs33 (approved):

Issue #169, opened 2020-06-20
x86_64 port


DrDeano:
As 32 bit is a good start, most PC today use 64biut CPUs. We would like this OS to be modern, so will need to support 64 bit.

In the future, the 32 bit port may be rolled out in favour of 64bit as the default build. 32 bit will still be supported, but second or third to x86_64 and Aarch64 ports.

SamTebbs33:
Agreed that it should be the default.

PR #291, opened 2021-02-25
Boot 64 bit using Limine

DrDeano:
Boot the kernel into 64 bit mode using the Limine bootloader.
Basic memory initialisation
Fixes arch agnostic tests
Part of #169

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -21,10 +21,34 @@ const x86_i686 = CrossTarget{
     .cpu_model = .{ .explicit = &Target.x86.cpu._i686 },
 };
 
+const x86_64 = brk: {
+    var tmp = CrossTarget{
+        .cpu_arch = .x86_64,
+        .os_tag = .freestanding,
+    };
+    var disabled_features = std.Target.Cpu.Feature.Set.empty;
+    var enabled_features = std.Target.Cpu.Feature.Set.empty;
+
+    const features = std.Target.x86.Feature;
+    // Disable SIMD registers
+    disabled_features.addFeature(@enumToInt(features.mmx));
+    disabled_features.addFeature(@enumToInt(features.sse));
+    disabled_features.addFeature(@enumToInt(features.sse2));
+    disabled_features.addFeature(@enumToInt(features.avx));
+    disabled_features.addFeature(@enumToInt(features.avx2));
+
+    enabled_features.addFeature(@enumToInt(features.soft_float));
+
+    tmp.cpu_features_sub = disabled_features;
+    tmp.cpu_features_add = enabled_features;
+    break :brk tmp;
+};
+
 pub fn build(b: *Builder) !void {
-    const target = b.standardTargetOptions(.{ .whitelist = &[_]CrossTarget{x86_i686}, .default_target = x86_i686 });
+    const target = b.standardTargetOptions(.{ .whitelist = &[_]CrossTarget{ x86_i686, x86_64 }, .default_target = x86_64 });
     const arch = switch (target.getCpuArch()) {
         .i386 => "x86",
+        .x86_64 => "x86_64",
         else => unreachable,
     };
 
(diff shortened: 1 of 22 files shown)


build.zig
```diff
@@ -64,9 +92,11 @@ pub fn build(b: *Builder) !void {
     exec.setBuildMode(build_mode);
     exec.setLinkerScriptPath(linker_script_path);
     exec.setTarget(target);
+    exec.code_model = .kernel;
 
     const make_iso = switch (target.getCpuArch()) {
         .i386 => b.addSystemCommand(&[_][]const u8{ "./makeiso.sh", boot_path, modules_path, iso_dir_path, exec.getOutputPath(), ramdisk_path, output_iso }),
+        .x86_64 => b.addSystemCommand(&[_][]const u8{ "./makeiso_64.sh", b.exe_dir, exec.getOutputPath(), output_iso, ramdisk_path }),
```
SamTebbs33, line 99: Would it be feasible to have just one makiso.sh file and change behaviour based on the architecture being built?

src/kernel/arch/x86_64/arch.zig
```diff
@@ -0,0 +1,458 @@
...
+    // Interrupt number and error code
+    int_num: u32,
+    error_code: u32,
+
+    // Instruction pointer, code segment and flags
+    eip: u32,
+    cs: u32,
+    eflags: u32,
+    user_esp: u32,
+    user_ss: u32,
```
SamTebbs33: Shouldn't most of these be u64 (usize)?

src/kernel/arch/x86_64/arch.zig
```diff
@@ -0,0 +1,458 @@
...
+        if (entry.memtype != MemType.Usable and entry.memtype != MemType.BootloaderReclaimable and entry.memtype != MemType.KernelAndModules) {
+            // If addr + len is greater than maxInt(usize) just ignore whatever comes after maxInt(usize) since it can't be addressed anyway
+            // TODO: I thing this needs to be u40 for 64 bit or u48?
+            const end: usize = if (entry.base_addr > std.math.maxInt(usize) - entry_len) std.math.maxInt(usize) else @intCast(usize, entry.base_addr + entry_len);
+            try reserved_physical_mem.append(.{
+                .start = @intCast(usize, entry.base_addr),
+                .end = end,
+            });
+        }
+    }
```
SamTebbs33, line 361: If I understand correctly then `memory_size_bytes` will only be the amount of memory that isn't reserved. The memory profile is however expected to specify the total amount of memory, including what is reserved.

build.zig
```diff
@@ -36,7 +60,11 @@ pub fn build(b: *Builder) !void {
     });
     b.default_step.dependOn(&fmt_step.step);
 
-    const main_src = "src/kernel/kmain.zig";
+    const main_src = switch (target.getCpuArch()) {
+        .i386 => "src/kernel/kmain.zig",
+        .x86_64 => "src/kernel/kmain_64.zig",
```
SamTebbs33, line 65: What I do in the aarch64 port is check for the architecture in kmain and skip certain parts of the initialisation if it's aarch64. You could consider doing that for x86_64 as well but either is fine as it's WIP.

src/kernel/arch/x86_64/arch.zig
DrDeano: Have removed this as will be different for 64 bit

src/kernel/arch/x86_64/arch.zig
DrDeano: I have reworked the memory size.

build.zig
DrDeano: From offline discussion, until the 64 bit port is at the same level as the 32 bit port, these files will be merged as a lot of the kernel.zig code will need to be commented out for the 64 bit port to work. So creating a new kmain and using that will reduce the amount of commented code and make thing clearer.

build.zig
DrDeano: The two files don't have much in common, so is easier to have 2 separate bash files than switching inside the bash file for the build target. This might change later on though.

SamTebbs33 (approved):
LGTM

PR #292, opened 2021-03-05
Improve build speed

DrDeano:
This improves the speed of clearing of the FAT image with using a bigger buffer


SamTebbs33 (approved):
Nice improvement!

PR #294, opened 2021-03-17
Feature/arch init gdt idt

DrDeano:
This adds the GDT, TSS and IDT implementations for the 64 bit port.
Part of #169

### DIFF
--- a/src/kernel/arch/x86_64/arch.zig
+++ b/src/kernel/arch/x86_64/arch.zig
@@ -1,6 +1,8 @@
 const std = @import("std");
 const Allocator = std.mem.Allocator;
 const log = std.log.scoped(.x86_64_arch);
+const gdt = @import("gdt.zig");
+const idt = @import("idt.zig");
 const serial = @import("serial.zig");
 const stivale2 = @import("stivale2.zig");
 const paging = @import("paging.zig");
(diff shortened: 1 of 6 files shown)


SamTebbs33 (commented):
Lots of the functions and constants are the same as in 32 bit so can we share them instead of copying them?

DrDeano:
I was thinking about this, but will probably have to reorder the arch folder structure. Would:
|- arch
     |--- x86
     |        |--- 32bit
     |        |--- 64bit
     |        |--- common
     |--- arm

So if we wanted to support 32 bit arm, then can follow the came structure. What do you think?

SamTebbs33:
> I was thinking about this, but will probably have to reorder the arch folder structure. Would:
> |- arch
> |--- x86
> | |--- 32bit
> | |--- 64bit
> | |--- common
> |--- arm
> 
> So if we wanted to support 32 bit arm, then can follow the came structure. What do you think?

Yeah that looks good to me. Having a dir for common code would be cleaner than the 64 bit port referencing 32 bit stuff.

SamTebbs33 (approved):
Feel free to make the dir structure changes in this PR or a future one.

DrDeano:
> 
> 
> Feel free to make the dir structure changes in this PR or a future one.

I shall do in another PR

PR #296, opened 2021-04-04
Fix typo

ghost:
Fixes a typo in the shell script to generate the iso file

### DIFF
--- a/makeiso_64.sh
+++ b/makeiso_64.sh
@@ -9,7 +9,7 @@ RAMDISK=$4
 
 if [[ ! -f ./echfs/echfs-utils ]]
 then
-    echo "Biulding echfs-utils"
+    echo "Building echfs-utils"
 	cd echfs
 	make echfs-utils
 	cd ..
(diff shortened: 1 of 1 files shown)


DrDeano:
The Github actions didn't run. I think you need to enable actions in your fork.

ghost:
This is weird, I have manually enabled them and have tried to trigger it manually but I don't see the button for it;
I think an event needs to be explicitly configured for manual triggers to work

SamTebbs33:
Hmm we've had issues with checks not running for PRs from forks but I thought they were fixed 🤔 I'm happy to approve this as it has no change to functionality. Perhaps we should move to a different service for running checks.

ghost:
in that case I'd recommend Travis CI, the integration to GitHub is pretty seamless / painless (This might be totally unrelated but I think it'd be a good idea to ship a Dockerfile as an environment to build the ISO in, I can also make it :thinking:)

DrDeano (approved):
Yeh, as this has no build changes, this can be merged 🚀

SamTebbs33:
Thanks @InfRandomness  :)

PR #298, opened 2021-04-08
Adding MBR partitioning and using it to create a boot image

DrDeano:
Sorry for this being a bit long. I was hoping to split this up into two PR, one for the MBR stuff and one for building a boot image.
But what this adds:
 - Simple MBR partition scheme implementation
 - Refactored makefs to include making a MBR partition table.
 - Refactored some common filesystem operations
 - Removed echfs and parted dependencies (used MBR and FAT32 in house code)
Pls don't look at the build file. It really needs a big refactor, it is a mess, but is working ;) This should be fixed in another PR, contributor friendly :)


src/kernel/filesystem/filesystem_common.zig
```diff
@@ -0,0 +1,173 @@
+const std = @import("std");
+
+// This is the assembly for the FAT bootleader.
```
SamTebbs33: bootleader -> bootloader

src/kernel/filesystem/filesystem_common.zig
```diff
@@ -0,0 +1,173 @@
...
+//     hlt
+//     jmp .loop_forever
+// message db "This is not a bootable disk. Please insert a bootable floppy and press any key to try again", 0
+// times 510 - ($ - $$) db 0
+// dw 0xAA55
+
+/// Basic boot code that will just print to the scream to insert a bootable image. This is intended
+/// as a place holder to be over written with real boot code if needed.
+/// This assumes the 512 sector size and includes the 0xAA55 boot signature.
+pub const bootsector_boot_code = [512]u8{
```
SamTebbs33: Wouldn't it be more suitable for this to be in the mbr file as that is what uses it?

src/kernel/time.zig
```diff
@@ -0,0 +1,31 @@
...
+    hour: u32,
+    day: u32,
+    month: u32,
+    year: u32,
+    century: u32,
+    day_of_week: u32,
+};
+
+pub fn getDateTime() DateTime {
+    return switch (builtin.os.tag) {
```
SamTebbs33, line 15: I'm a little confused as to why this is necessary, when would we not be in freestanding and why do we need to accommodate for other cases?

src/kernel/filesystem/filesystem_common.zig
DrDeano: It is also used in the FAT32 header, and maybe others for the boot sector.

src/kernel/time.zig
DrDeano: I use this driver in the build script so runs on the hosts OS, therefore need a non-freestanding version but there isn't one in the standard library yet, see: https://github.com/ziglang/zig/issues/8396

src/kernel/filesystem/filesystem_common.zig
SamTebbs33: Hmm, since this is so specialised (we don't yet have other filesystems that use it) and is x86-specific, perhaps it should be moved to somewhere in arch/x86{_64}.

src/kernel/time.zig
SamTebbs33: I see, but why use this when building? Is it just because it has a nicer interface than what is currently in std? Using kernel code on the host is a little weird to me.

src/kernel/filesystem/filesystem_common.zig
DrDeano: I can do as it is x86 code. I don't know how this boot code is done for ARM, might do some testing for creating a FAT32/MBR partition on a ARM system.

src/kernel/time.zig
DrDeano: We could use host stuff like parted and mkfs, but I like using the code we already have that is used in the kernel as we have it so why not use it, plus less dependencies as this can be used on any platform when building as Windows doesn't have parted or mkfs. Plus ensures that what we make using the drivers can be read using the driver (self testing).

src/kernel/filesystem/filesystem_common.zig
SamTebbs33: Nice :)

DrDeano: > 
> 
> I can do as it is x86 code. I don't know how this boot code is done for ARM, might do some testing for creating a FAT32/MBR partition on a ARM system.

I'll make this a contributor friendly issue for the ARM boot sector

SamTebbs33 (approved):

PR #303, opened 2021-05-01
Add syscalls for VFS

### LINKED ISSUE #265: Filesystem syscalls
Now that we have user mode implemented, we need some syscalls that will allow programs to interact with the filesystem.
```
openFile(path: [*]const u8, path_len: usize, create: bool) usize
openDir(path: [*]const u8, path_len: usize, create: bool) usize
openSymlink(path: [*]const u8, path_len: usize, target: [*]const u8, target_len: usize, create: bool) usize
read(handle: usize, buffer: [*]u8, buffer_len: usize) usize
write(handle: usize, buffer: [*]const u8, buffer_len: usize) usize
close(handle: usize)
```
The open functions return an integer representing a handle to that file. That integer will be used to refer to the file once it is opened. File handles are unique to each process.

Slices must be passed as a pointer and size separately since syscalls use the CPU registers which cannot store complex types. such as slices.

SamTebbs33:
This patch adds the open, read, write and close VFS operations to the syscall interface.

* open(path_ptr, path_len, flags, args_ptr) handle
* close(handle) void
* read(handle, buff_ptr, buf_len) bytes_read
* write(handle, buff_ptr, buff_len) bytes_written

It also adds a handle -> vfs node mapping for each task to keep track of what a handle corresponds to for that task.

Closes #265

### DIFF
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -4,6 +4,8 @@ const TailQueue = std.TailQueue;
 const ArrayList = std.ArrayList;
 const Allocator = std.mem.Allocator;
 
+pub const Handle = usize;
+
 /// Flags specifying what to do when opening a file or directory
 pub const OpenFlags = enum {
     /// Create a directory if it doesn't exist
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -284,7 +284,7 @@ pub const MountError = error{
 pub const SEPARATOR: u8 = '/';
 
 /// The root of the system's top-level filesystem
-pub var root: *Node = undefined;
+var root: *Node = undefined;
 
 ///
 /// Traverse the specified path from the root and open the file/dir corresponding to that path. If the file/dir doesn't exist it can be created by specifying the open flags
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -422,6 +422,20 @@ pub fn open(path: []const u8, follow_symlinks: bool, flags: OpenFlags, args: Ope
 }
 
 ///
+/// Close a node.
+///
+/// Arguments:
+///     IN node: Node - The node to close
+///
+pub fn close(node: Node) void {
+    switch (node) {
+        .Dir => |d| d.close(),
+        .File => |f| f.close(),
+        .Symlink => |s| s.close(),
+    }
+}
+
+///
 /// Open a file at a path.
 ///
 /// Arguments:
--- a/src/kernel/filesystem/vfs.zig
+++ b/src/kernel/filesystem/vfs.zig
@@ -4,8 +4,6 @@ const TailQueue = std.TailQueue;
 const ArrayList = std.ArrayList;
 const Allocator = std.mem.Allocator;
 
-pub const Handle = usize;
-
 /// Flags specifying what to do when opening a file or directory
 pub const OpenFlags = enum {
     /// Create a directory if it doesn't exist
@@ -213,13 +211,8 @@ pub const DirNode = struct {
 
     /// See the documentation for FileSystem.Open
     pub fn open(self: *const DirNode, name: []const u8, flags: OpenFlags, args: OpenArgs) (Allocator.Error || Error)!*Node {
-        var fs = self.fs;
-        var node = self;
-        if (self.mount) |mnt| {
-            fs = mnt.fs;
-            node = mnt;
-        }
-        return fs.open(fs, node, name, flags, args);
+        var node = self.mount orelse self;
+        return node.fs.open(node.fs, node, name, flags, args);
     }
 
     /// See the documentation for FileSystem.Close
@@ -291,7 +284,7 @@ pub const MountError = error{
 pub const SEPARATOR: u8 = '/';
 
 /// The root of the system's top-level filesystem
-var root: *Node = undefined;
+pub var root: *Node = undefined;
 
 ///
 /// Traverse the specified path from the root and open the file/dir corresponding to that path. If the file/dir doesn't exist it can be created by specifying the open flags
@@ -594,7 +587,7 @@ const TestFS = struct {
 
     const Self = @This();
 
-    fn deinit(self: *@This()) void {
+    pub fn deinit(self: *@This()) void {
         self.tree.deinit(self.allocator);
         self.allocator.destroy(self.fs);
     }
@@ -720,7 +713,7 @@ const TestFS = struct {
     }
 };
 
-fn testInitFs(allocator: Allocator) !*TestFS {
+pub fn testInitFs(allocator: Allocator) !*TestFS {
     const fs = try allocator.create(FileSystem);
     var testfs = try allocator.create(TestFS);
     var root_node = try allocator.create(Node);
--- a/src/kernel/scheduler.zig
+++ b/src/kernel/scheduler.zig
@@ -27,7 +27,7 @@ extern var KERNEL_STACK_START: []u32;
 extern var KERNEL_STACK_END: []u32;
 
 /// The current task running
-var current_task: *Task = undefined;
+pub var current_task: *Task = undefined;
 
 /// Array list of all runnable tasks
 var tasks: TailQueue(*Task) = undefined;
--- a/src/kernel/scheduler.zig
+++ b/src/kernel/scheduler.zig
@@ -139,17 +139,12 @@ pub fn init(allocator: Allocator, mem_profile: *const mem.MemProfile) Allocator.
     tasks = TailQueue(*Task){};
 
     // Set up the init task to continue execution
-    current_task = try allocator.create(Task);
+    current_task = try Task.create(0, true, &vmm.kernel_vmm, allocator);
     errdefer allocator.destroy(current_task);
-    // PID 0
-    current_task.pid = 0;
+
+    allocator.free(current_task.kernel_stack);
     const kernel_stack_size = @ptrToInt(&KERNEL_STACK_END) - @ptrToInt(&KERNEL_STACK_START);
     current_task.kernel_stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..kernel_stack_size];
-    current_task.user_stack = &[_]usize{};
-    current_task.kernel = true;
-    // TODO: Use Task.create instead of setting it up manually
-    current_task.file_handles = try bitmap.Bitmap(usize).init(task.VFS_HANDLES_PER_PROCESS, allocator);
-    current_task.file_handle_mapping = std.hash_map.AutoHashMap(task.Handle, *fs.Node).init(allocator);
     // ESP will be saved on next schedule
 
     // Run the runtime tests here
@@ -199,9 +194,9 @@ test "pickNextTask" {
     tasks = TailQueue(*Task){};
 
     // Set up a current task
-    current_task = try allocator.create(Task);
-    defer allocator.destroy(current_task);
-    current_task.pid = 0;
+    current_task = try Task.create(0, true, &vmm.kernel_vmm, allocator);
+    defer current_task.destroy(allocator);
+    allocator.free(current_task.kernel_stack);
     current_task.kernel_stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..4096];
     current_task.stack_pointer = @ptrToInt(&KERNEL_STACK_START);
 
--- a/src/kernel/scheduler.zig
+++ b/src/kernel/scheduler.zig
@@ -14,6 +14,7 @@ const mem = @import("mem.zig");
 const fs = @import("filesystem/vfs.zig");
 const elf = @import("elf.zig");
 const pmm = @import("pmm.zig");
+const bitmap = @import("bitmap.zig");
 const Task = task.Task;
 const EntryPoint = task.EntryPoint;
 const Allocator = std.mem.Allocator;
@@ -146,6 +147,9 @@ pub fn init(allocator: Allocator, mem_profile: *const mem.MemProfile) Allocator.
     current_task.kernel_stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..kernel_stack_size];
     current_task.user_stack = &[_]usize{};
     current_task.kernel = true;
+    // TODO: Use Task.create instead of setting it up manually
+    current_task.file_handles = try bitmap.Bitmap(usize).init(task.VFS_HANDLES_PER_PROCESS, allocator);
+    current_task.file_handle_mapping = std.hash_map.AutoHashMap(task.Handle, *fs.Node).init(allocator);
     // ESP will be saved on next schedule
 
     // Run the runtime tests here
--- a/src/kernel/syscalls.zig
+++ b/src/kernel/syscalls.zig
@@ -211,17 +211,29 @@ fn getData(ptr: usize, len: usize) Error![]u8 {
     }
 }
 
-fn handleOpen(path_ptr: usize, path_len: usize, flags: usize, ignored1: usize, ignored2: usize) Error!usize {
+fn handleOpen(path_ptr: usize, path_len: usize, flags: usize, args: usize, ignored: usize) Error!usize {
     const current_task = scheduler.current_task;
     if (!current_task.hasFreeVFSHandle()) {
         return Error.NoMoreFSHandles;
     }
 
+    // Fetch the open arguments from user/kernel memory
+    var open_args: vfs.OpenArgs = if (args == 0) .{} else blk: {
+        const data = try getData(args, @sizeOf(vfs.OpenArgs));
+        defer if (!current_task.kernel) allocator.free(data);
+        break :blk std.mem.bytesAsValue(vfs.OpenArgs, data[0..@sizeOf(vfs.OpenArgs)]).*;
+    };
+    // The symlink target could refer to a location in user memory so convert that too
+    if (open_args.symlink_target) |target| {
+        open_args.symlink_target = try getData(@ptrToInt(target.ptr), target.len);
+    }
+    defer if (!current_task.kernel) if (open_args.symlink_target) |target| allocator.free(target);
+
     const open_flags = std.meta.intToEnum(vfs.OpenFlags, flags) catch return Error.InvalidFlags;
     const path = try getData(path_ptr, path_len);
     defer if (!current_task.kernel) allocator.free(path);
 
-    var node = try vfs.open(path, true, open_flags, .{});
+    var node = try vfs.open(path, true, open_flags, open_args);
     return current_task.addVFSHandle(node) orelse panic(null, "Failed to add a VFS handle to current_task\n", .{});
 }
 
@@ -426,5 +438,12 @@ test "handleRead" {
     {
         const length = try handleRead(test_file, @ptrToInt(&buffer[0]), 0, undefined, undefined);
         testing.expect(std.mem.eql(u8, str[0..0], buffer[0..length]));
+    }
+    // Try reading from a symlink
+    const args = vfs.OpenArgs{ .symlink_target = "/foo.txt" };
+    var test_link = try handleOpen(@ptrToInt("/link"), "/link".len, @enumToInt(vfs.OpenFlags.CREATE_SYMLINK), @ptrToInt(&args), undefined);
+    {
+        const length = try handleRead(test_link, @ptrToInt(&buffer[0]), buffer.len, undefined, undefined);
+        testing.expect(std.mem.eql(u8, str[0..str.len], buffer[0..length]));
     }
 }
--- a/src/kernel/syscalls.zig
+++ b/src/kernel/syscalls.zig
@@ -78,12 +78,12 @@ pub fn init(alloc: *std.mem.Allocator) void {
 /// Passing an error code that does not correspond to an error results in safety-protected undefined behaviour
 ///
 /// Arguments:
-///     IN code: usize - The erorr code to convert
+///     IN code: u16 - The erorr code to convert
 ///
 /// Return: Error
 ///     The error corresponding to the error code
 ///
-pub fn fromErrorCode(code: usize) Error {
+pub fn fromErrorCode(code: u16) anyerror {
     return @intToError(code);
 }
 
@@ -93,10 +93,10 @@ pub fn fromErrorCode(code: usize) Error {
 /// Arguments:
 ///     IN err: Error - The erorr to convert
 ///
-/// Return: usize
+/// Return: u16
 ///     The error code corresponding to the error
 ///
-pub fn toErrorCode(err: Error) usize {
+pub fn toErrorCode(err: anyerror) u16 {
     return @errorToInt(err);
 }
 
--- a/src/kernel/syscalls.zig
+++ b/src/kernel/syscalls.zig
@@ -220,8 +220,7 @@ fn handleWrite(node_handle: usize, buff_ptr: usize, buff_len: usize, ignored1: u
 
         // TODO: A more performant method would be mapping in the user memory and using that directly. Then we wouldn't need to allocate or copy the buffer
         if (!current_task.kernel) try vmm.kernel_vmm.copyData(current_task.vmm, buff, buff_ptr, false);
-        const bytes_read = try file.read(buff);
-        return bytes_read;
+        return try file.write(buff);
     }
 
     return Error.NotOpened;
@@ -234,7 +233,6 @@ fn handleClose(node_handle: usize, ignored1: usize, ignored2: usize, ignored3: u
     const current_task = scheduler.current_task;
     const node_opt = current_task.getVFSHandle(real_handle) catch panic(@errorReturnTrace(), "Failed to get VFS node for handle {}\n", .{real_handle});
     if (node_opt) |node| {
-        //try node.close();
         current_task.clearVFSHandle(real_handle) catch |e| return switch (e) {
             error.VFSHandleNotSet, error.OutOfBounds => Error.NotOpened,
         };
--- a/src/kernel/syscalls.zig
+++ b/src/kernel/syscalls.zig
@@ -622,6 +622,48 @@ test "handleWrite" {
     try testing.expectEqualSlices(u8, data2, testfs.tree.children.items[1].children.items[0].data.?);
 }
 
+test "handleWrite errors" {
+    allocator = std.testing.allocator;
+    var testfs = try vfs.testInitFs(allocator);
+    {
+        defer allocator.destroy(testfs);
+        defer testfs.deinit();
+
+        testfs.instance = 1;
+        try vfs.setRoot(testfs.tree.val);
+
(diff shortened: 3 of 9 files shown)


src/kernel/filesystem/vfs.zig
```diff
@@ -211,13 +211,8 @@ pub const DirNode = struct {
...
     pub fn open(self: *const DirNode, name: []const u8, flags: OpenFlags, args: OpenArgs) (Allocator.Error || Error)!*Node {
-        var fs = self.fs;
-        var node = self;
-        if (self.mount) |mnt| {
-            fs = mnt.fs;
-            node = mnt;
-        }
-        return fs.open(fs, node, name, flags, args);
+        var node = self.mount orelse self;
+        return node.fs.open(node.fs, node, name, flags, args);
```
DrDeano, line 215: This is so much nicer :)

src/kernel/syscalls.zig
```diff
@@ -80,6 +91,79 @@ pub const Syscall = enum {
...
+        11 => Error.NoSymlinkTarget,
+        12 => Error.OutOfBounds,
+        13 => Error.NotAllocated,
+        14 => Error.NotAFile,
+        15 => Error.AlreadyAllocated,
+        16 => Error.PhysicalAlreadyAllocated,
+        17 => Error.PhysicalVirtualMismatch,
+        18 => Error.InvalidVirtAddresses,
+        19 => Error.InvalidPhysAddresses,
+        20 => Error.Unexpected,
```
DrDeano: Can use `@errorToInt`, so can remove these 'to' and 'from' functions

src/kernel/task.zig
```diff
@@ -1,20 +1,32 @@
...
 const vmm = @import("vmm.zig");
+const vfs = @import("filesystem/vfs.zig");
 const Allocator = std.mem.Allocator;
 
 /// The kernels main stack start as this is used to check for if the task being destroyed is this stack
 /// as we cannot deallocate this.
 extern var KERNEL_STACK_START: *u32;
 
+/// The number of vfs handles that a process can have. This is arbitrarily set to 65535
+pub const VFS_HANDLES_PER_PROCESS = 65535;
```
DrDeano: could use `std.math.maxInt(u16)` then can remove the comptime check below.

src/kernel/filesystem/vfs.zig
```diff
@@ -289,7 +284,7 @@ pub const MountError = error{
 pub const SEPARATOR: u8 = '/';
 
 /// The root of the system's top-level filesystem
-var root: *Node = undefined;
+pub var root: *Node = undefined;
```
DrDeano: How come this is `pub`? There is a `setRoot` function.

src/kernel/scheduler.zig
```diff
@@ -144,6 +145,9 @@ pub fn init(allocator: *Allocator, mem_profile: *const mem.MemProfile) Allocator
     current_task.kernel_stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..kernel_stack_size];
     current_task.user_stack = &[_]usize{};
     current_task.kernel = true;
+    // TODO: Use Task.create instead of setting it up manually
+    current_task.file_handles = try bitmap.Bitmap(usize).init(task.VFS_HANDLES_PER_PROCESS, allocator);
```
DrDeano: This could use the `ComptimeBitmap` as the number of handles is know at comptime. Can update the `ComptimeBitmap` `init` functions to take a size so will create an array with a known size.

src/kernel/syscalls.zig
```diff
@@ -1,44 +1,47 @@
...
+    NoSymlinkTarget,
+    OutOfBounds,
+    NotAllocated,
+    AlreadyAllocated,
+    PhysicalAlreadyAllocated,
+    PhysicalVirtualMismatch,
+    InvalidVirtAddresses,
+    InvalidPhysAddresses,
+    Unexpected,
+};
```
DrDeano: Would be nicer to concatenate the public errors that the syscall will use. (From offline discussion)

src/kernel/syscalls.zig
```diff
@@ -55,6 +58,10 @@ comptime {
 
 /// All implemented syscalls
 pub const Syscall = enum {
+    Open,
+    Read,
+    Write,
+    Close,
```
DrDeano: From offline discusion, should not include the test syscalls in the non-test builds of the kernel. So can construct the enum using `@Type`

src/kernel/task.zig
```diff
@@ -34,6 +46,7 @@ pub const STACK_SIZE: u32 = arch.MEMORY_BLOCK_SIZE / @sizeOf(u32);
 
 /// The task control block for storing all the information needed to save and restore a task.
 pub const Task = struct {
+    pub const Error = error{VFSHandleNotSet};
```
DrDeano: Doc comment

src/kernel/task.zig
```diff
@@ -115,8 +136,41 @@ pub const Task = struct {
...
+
+    pub fn hasFreeVFSHandle(self: @This()) bool {
+        return self.file_handles.num_free_entries > 0;
+    }
+
+    pub fn addVFSHandle(self: *@This(), node: *vfs.Node) ?usize {
+        if (self.hasFreeVFSHandle()) {
+            // Cannot error as we've already checked that there is a free entry
+            const handle = self.file_handles.setFirstFree() orelse unreachable;
+            self.file_handle_mapping.put(handle, node) catch unreachable;
```
DrDeano: Adding to the hash map should handle the error as is unrelated to the `hasFreeVFSHandle` check. So could return a OOM.

src/kernel/task.zig
```diff
@@ -115,8 +136,41 @@ pub const Task = struct {
...
+    }
+
+    pub fn hasFreeVFSHandle(self: @This()) bool {
+        return self.file_handles.num_free_entries > 0;
+    }
+
+    pub fn addVFSHandle(self: *@This(), node: *vfs.Node) ?usize {
+        if (self.hasFreeVFSHandle()) {
+            // Cannot error as we've already checked that there is a free entry
+            const handle = self.file_handles.setFirstFree() orelse unreachable;
```
DrDeano: As `setFirstFree` returns a options, you could have:

```zig
if (self.file_handles.setFirstFree()) |handle| {
    ...
}
return null;
```

src/kernel/scheduler.zig
```diff
@@ -144,6 +145,9 @@ pub fn init(allocator: *Allocator, mem_profile: *const mem.MemProfile) Allocator
     current_task.kernel_stack = @intToPtr([*]u32, @ptrToInt(&KERNEL_STACK_START))[0..kernel_stack_size];
     current_task.user_stack = &[_]usize{};
     current_task.kernel = true;
+    // TODO: Use Task.create instead of setting it up manually
```
DrDeano: Is this TODO referring to the `current_task`? If so, why not do it :P

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+    return current_task.addVFSHandle(node) orelse panic(null, "Failed to add a VFS handle to current_task\n", .{});
+}
+
+fn handleRead(node_handle: usize, buff_ptr: usize, buff_len: usize, ignored1: usize, ignored2: usize) Error!usize {
+    if (buff_len > USER_MAX_DATA_LEN) {
+        return Error.TooBig;
+    }
+
+    const current_task = scheduler.current_task;
+    const node_opt = current_task.getVFSHandle(node_handle) catch panic(@errorReturnTrace(), "Failed to get VFS node for handle {}\n", .{node_handle});
```
DrDeano: Maybe we should return an error instead of panic?

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+    if (node_opt) |node| {
+        const file = switch (node.*) {
+            .File => |f| f,
+            else => return Error.NotAFile,
+        };
+        var buff = if (current_task.kernel) @intToPtr([*]u8, buff_ptr)[0..buff_len] else try allocator.alloc(u8, buff_len);
+        defer if (!current_task.kernel) allocator.free(buff);
+
+        // TODO: A more performant method would be mapping in the user memory and using that directly. Then we wouldn't need to allocate or copy the buffer
```
DrDeano, line 354: Cool, I was thinking how to remove the global allocator, but see it will come later.

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+        if (!current_task.kernel) try vmm.kernel_vmm.copyData(current_task.vmm, buff, buff_ptr, false);
+        const bytes_read = try file.read(buff);
+        return bytes_read;
```
DrDeano: Shouldn't this be write? `return file.write(buff);`?

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+    }
+
+    return Error.NotOpened;
+}
+
+fn handleClose(node_handle: usize, ignored1: usize, ignored2: usize, ignored3: usize, ignored4: usize) Error!usize {
+    const current_task = scheduler.current_task;
+    const node_opt = current_task.getVFSHandle(node_handle) catch panic(@errorReturnTrace(), "Failed to get VFS node for handle {}\n", .{node_handle});
+    if (node_opt) |node| {
+        //try node.close();
```
DrDeano: Commented code

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+        const bytes_read = try file.read(buff);
+        return bytes_read;
+    }
+
+    return Error.NotOpened;
+}
+
+fn handleClose(node_handle: usize, ignored1: usize, ignored2: usize, ignored3: usize, ignored4: usize) Error!usize {
+    const current_task = scheduler.current_task;
+    const node_opt = current_task.getVFSHandle(node_handle) catch panic(@errorReturnTrace(), "Failed to get VFS node for handle {}\n", .{node_handle});
```
DrDeano: Same here, can return an error instead

src/kernel/syscalls.zig
```diff
@@ -1,44 +1,47 @@
...
+    PhysicalAlreadyAllocated,
+    PhysicalVirtualMismatch,
+    InvalidVirtAddresses,
+    InvalidPhysAddresses,
+    Unexpected,
+};
 
 comptime {
     // Make sure toErrorCode and fromErrorCode are synchronised, and that no errors share the same error code
+    @setEvalBranchQuota(1000000);
```
DrDeano: Seems excessive, but using the `@errorToInt` will remove this comptime check

src/kernel/syscalls.zig
DrDeano: From offline discussion, should doc comment the syscall interface with what the parameters are.

src/kernel/task.zig
```diff
@@ -115,8 +136,41 @@ pub const Task = struct {
...
+        return null;
+    }
+
+    pub fn hasVFSHandle(self: @This(), handle: Handle) bitmap.Bitmap(usize).BitmapError!bool {
+        return self.file_handles.isSet(handle);
+    }
+
+    pub fn clearVFSHandle(self: *@This(), handle: Handle) (bitmap.Bitmap(usize).BitmapError || Error)!void {
+        if (try self.hasVFSHandle(handle)) {
+            try self.file_handles.clearEntry(handle);
```
DrDeano, line 199: The entry exists, so don't need the try.

src/kernel/task.zig
```diff
@@ -254,3 +308,75 @@ test "allocatePid and freePid" {
...
     expectEqual(all_pids.bitmap, 0);
 }
+
+test "addVFSHandle" {
+    var task = try Task.create(0, true, undefined, std.testing.allocator);
+    defer task.destroy(std.testing.allocator);
+    var node1 = vfs.Node{ .Dir = .{ .fs = undefined, .mount = null } };
+    var node2 = vfs.Node{ .File = .{ .fs = undefined } };
+
+    const handle1 = task.addVFSHandle(&node1) orelse unreachable;
```
DrDeano: Maybe return an error instead of unreachable, as if the test are compiled as release-fast/small, then will have undefined behavior.

src/kernel/task.zig
```diff
@@ -115,8 +136,41 @@ pub const Task = struct {
         if (!self.kernel) {
             allocator.free(self.user_stack);
         }
+        self.file_handles.deinit();
+        self.file_handle_mapping.deinit();
         allocator.destroy(self);
     }
+
+    pub fn getVFSHandle(self: @This(), handle: Handle) bitmap.Bitmap(usize).BitmapError!?*vfs.Node {
```
DrDeano: Can use `Self` instead of `@This()`

src/kernel/syscalls.zig
```diff
@@ -113,10 +313,125 @@ test "getHandler" {
     std.testing.expectEqual(Syscall.Test1.getHandler(), handleTest1);
     std.testing.expectEqual(Syscall.Test2.getHandler(), handleTest2);
     std.testing.expectEqual(Syscall.Test3.getHandler(), handleTest3);
+    std.testing.expectEqual(Syscall.Open.getHandler(), handleOpen);
+    std.testing.expectEqual(Syscall.Close.getHandler(), handleClose);
+    std.testing.expectEqual(Syscall.Read.getHandler(), handleRead);
+    std.testing.expectEqual(Syscall.Write.getHandler(), handleWrite);
 }
 
 test "handle" {
```
DrDeano: I think this test should have a better name now, "handle basic functionality" maybe?

src/kernel/syscalls.zig
```diff
@@ -113,10 +313,125 @@ test "getHandler" {
...
+    testfs.instance = 1;
+    vfs.root = testfs.tree.val;
+
+    vmm.kernel_vmm = try vmm.VirtualMemoryManager(arch.VmmPayload).init(0, 1024, allocator, arch.VMM_MAPPER, arch.KERNEL_VMM_PAYLOAD);
+    defer vmm.kernel_vmm.deinit();
+    scheduler.current_task = try task.Task.create(0, true, &vmm.kernel_vmm, allocator);
+    defer scheduler.current_task.destroy(allocator);
+    var current_task = scheduler.current_task;
+
+    var test_file = try handleOpen(@ptrToInt("/foo.txt"), "/foo.txt".len, @enumToInt(vfs.OpenFlags.CREATE_FILE), undefined, undefined);
```
DrDeano: Maybe have `"/foo.txt"` in a const variable

src/kernel/syscalls.zig
```diff
@@ -113,10 +313,125 @@ test "getHandler" {
...
+        testing.expect(std.mem.eql(u8, str[0 .. str.len - 1], buffer[0..length]));
+    }
+
+    {
+        const length = try handleRead(test_file, @ptrToInt(&buffer[0]), 0, undefined, undefined);
+        testing.expect(std.mem.eql(u8, str[0..0], buffer[0..length]));
+    }
+    // Try reading from a symlink
+    const args = vfs.OpenArgs{ .symlink_target = "/foo.txt" };
+    var test_link = try handleOpen(@ptrToInt("/link"), "/link".len, @enumToInt(vfs.OpenFlags.CREATE_SYMLINK), @ptrToInt(&args), undefined);
```
DrDeano: `"/link"` as a variable

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+            return Error.TooBig;
+        }
+        var buff = try allocator.alloc(u8, len);
+        errdefer allocator.free(buff);
+        try vmm.kernel_vmm.copyData(scheduler.current_task.vmm, buff, ptr, false);
+        return buff;
+    }
+}
+
+fn handleOpen(path_ptr: usize, path_len: usize, flags: usize, args: usize, ignored: usize) Error!usize {
```
DrDeano, line 246: As these handler functions are C compatible, is it possible to a zig version, that will abstract the parameters to the actual types, like the flags as the `OpenFlags` enum which will then call the C version. So the user doesn't have to covert everything to `usize`.

DrDeano, line 246: Also, we will need to check the pointers as they could be kernel pointers and could lead to bad stuff.
This can be made into an issue, but just to make you aware. As for the write syscall as the user can write to any kernel address.

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+        open_args.symlink_target = try getData(@ptrToInt(target.ptr), target.len);
+    }
+    defer if (!current_task.kernel) if (open_args.symlink_target) |target| allocator.free(target);
+
+    const open_flags = std.meta.intToEnum(vfs.OpenFlags, flags) catch return Error.InvalidFlags;
+    const path = try getData(path_ptr, path_len);
+    defer if (!current_task.kernel) allocator.free(path);
+
+    var node = try vfs.open(path, true, open_flags, open_args);
+    return current_task.addVFSHandle(node) orelse panic(null, "Failed to add a VFS handle to current_task\n", .{});
```
DrDeano: Same here, can return an error

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+    if (open_args.symlink_target) |target| {
+        open_args.symlink_target = try getData(@ptrToInt(target.ptr), target.len);
+    }
+    defer if (!current_task.kernel) if (open_args.symlink_target) |target| allocator.free(target);
+
+    const open_flags = std.meta.intToEnum(vfs.OpenFlags, flags) catch return Error.InvalidFlags;
+    const path = try getData(path_ptr, path_len);
+    defer if (!current_task.kernel) allocator.free(path);
+
+    var node = try vfs.open(path, true, open_flags, open_args);
```
DrDeano: Will need a errdefer for this if `addVFSHandle` fails. Also const?

src/kernel/syscalls.zig
```diff
@@ -113,10 +313,125 @@ test "getHandler" {
...
+    std.testing.expectEqual(Syscall.Read.getHandler(), handleRead);
+    std.testing.expectEqual(Syscall.Write.getHandler(), handleWrite);
 }
 
 test "handle" {
     std.testing.expectEqual(@as(usize, 0), try handle(.Test1, 0, 0, 0, 0, 0));
     std.testing.expectEqual(@as(usize, 1 + 2 + 3 + 4 + 5), try handle(.Test2, 1, 2, 3, 4, 5));
     std.testing.expectError(Error.OutOfMemory, handle(.Test3, 0, 0, 0, 0, 0));
 }
+
```
DrDeano, line 468: Could add a test for memory leaks on errors.

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+}
+
+fn handleOpen(path_ptr: usize, path_len: usize, flags: usize, args: usize, ignored: usize) Error!usize {
+    const current_task = scheduler.current_task;
+    if (!current_task.hasFreeVFSHandle()) {
+        return Error.NoMoreFSHandles;
+    }
+
+    // Fetch the open arguments from user/kernel memory
+    var open_args: vfs.OpenArgs = if (args == 0) .{} else blk: {
```
DrDeano, line 254: const

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+        return Error.TooBig;
+    }
+
+    const current_task = scheduler.current_task;
+    const node_opt = current_task.getVFSHandle(node_handle) catch panic(@errorReturnTrace(), "Failed to get VFS node for handle {}\n", .{node_handle});
+    if (node_opt) |node| {
+        const file = switch (node.*) {
+            .File => |f| f,
+            else => return Error.NotAFile,
+        };
```
DrDeano: How are symlinks handled?

src/kernel/syscalls.zig
```diff
@@ -97,6 +181,122 @@ pub fn handle(syscall: Syscall, arg1: usize, arg2: usize, arg3: usize, arg4: usi
...
+        return bytes_read;
+    }
+
+    return Error.NotOpened;
+}
+
+fn handleWrite(node_handle: usize, buff_ptr: usize, buff_len: usize, ignored1: usize, ignored2: usize) Error!usize {
+    if (buff_len > USER_MAX_DATA_LEN) {
+        return Error.TooBig;
+    }
```
DrDeano: What is the purpose of this check

src/kernel/syscalls.zig
DrDeano, line 468: Missing write and close test

src/kernel/task.zig
```diff
@@ -92,6 +111,8 @@ pub const Task = struct {
             .stack_pointer = @ptrToInt(&k_stack[STACK_SIZE - 1]),
             .kernel = kernel,
             .vmm = task_vmm,
+            .file_handles = try bitmap.Bitmap(usize).init(VFS_HANDLES_PER_PROCESS, allocator),
```
DrDeano: This can be a `ComptimeBitmap` as discussed above.

DrDeano (changes requested):
Over all looks noice. But as always, missing doc comments.

Additionally, from offline discussion, there may not be a the open, read, write, close syscalls if we move the filesystem drivers to userspace. But will need a deeper think for this have how that would work, so for now, these interfaces will be a good start.

Also, could there be a follow up PR once the ELF has been merged with a test to test user -> kernel syscall.

src/kernel/filesystem/vfs.zig
SamTebbs33: There is indeed...

src/kernel/scheduler.zig
SamTebbs33: I tried this but it took a very long time to compile

src/kernel/task.zig
SamTebbs33: True

src/kernel/task.zig
SamTebbs33: Much better :)

src/kernel/scheduler.zig
DrDeano: But why? There shouldn't be any funky comptime stuff going on, just making a struct with an array?

src/kernel/scheduler.zig
SamTebbs33: Why not indeed :p

src/kernel/scheduler.zig
SamTebbs33: I don't know! I guess it struggles creating an integer with 65k bits.

src/kernel/syscalls.zig
SamTebbs33: This erroring wouldn't be the fault of the caller though.

src/kernel/syscalls.zig
SamTebbs33: It turns out that copying and pasting isn't always a good idea.

src/kernel/syscalls.zig
SamTebbs33: See answer to the other comment

src/kernel/scheduler.zig
DrDeano: But if you changed the `ComptimeBitmap` a comptime int that will create an array of the type, like how the normal `Bitmap` does, but just no allocation as is comptime knwon. So will keep the `usize` as the type, but the `init` array size will be comptime. Or if that doesn;t work, include the comptime array size with the type: `ComptimeBitmap(usize, 128).init()`

src/kernel/syscalls.zig
DrDeano: But the user will pass the node_handle?

SamTebbs33: Yes but `getVFSHandle` won't fail if the handle doesn't exist as it will then return null. It will only fail if something goes wrong with the map itself.

src/kernel/syscalls.zig
SamTebbs33: Why? That seems unnecessary as its only used in this one line.

src/kernel/syscalls.zig
SamTebbs33: The user doesn't have to convert everything to usize, the OS abstraction layer will do that part.

SamTebbs33: Yeah if the user provides a kernel pointer then we will end up overwriting it as we map the temporary buffer to the same physical blocks as the buffer pointer is mapped to. Good spot.

SamTebbs33: We should modify `getData` and `handleRead` to account for this.

src/kernel/syscalls.zig
SamTebbs33: The open syscall (that has to be called before reading) returns the node that the symlink points to rather than the symlink itself, so `node` won't refer to a symlink here.

src/kernel/syscalls.zig
SamTebbs33: We don't want the user maliciously asking the kernel to allocate gigabytes of memory so there's a hard limit to how big the buffer can be.

src/kernel/scheduler.zig
SamTebbs33: Success! I managed to do it. I originally modified the ComptimeBitmap struct to do as you suggested, but it became almost identical to the other Bitmap, so I merged them to reduce the maintenance overhead and code duplication.

src/kernel/syscalls.zig
SamTebbs33: I vote we make this an issue as this PR is pretty big already.

src/kernel/syscalls.zig
SamTebbs33: Indeed to both of those :)

src/kernel/task.zig
SamTebbs33: Donezo :)

src/kernel/syscalls.zig
SamTebbs33: This can't be const unfortunately since the `symlink_target` fields needs to be modified after setting this.

SamTebbs33:
This PR has gotten too big for its own good. I'll split it up into more sensibly-sized PRs and close this one. The branch will stay around for archival purposes, at least until everything is merged.

PR #306, opened 2021-05-05
Feature/reorganise arch structure

DrDeano:
Restructured the x86 ports to refactor common code

This adds a new folder structure structure:
- x86
  - 32bit
  - 64bit
  - common

Refactored arch, GDT, IDT and serial


makeiso_64.sh
```diff
@@ -23,7 +23,7 @@ then
 	cd ..
 fi
 
-MAP_FILE="zig-cache/kernel.map"
+MAP_FILE="zig-out/kernel.map"
```
SamTebbs33: I get a zig-cache directory when building unlike the CI, which gets zig-out. It might be a good idea to pass `b.install_path` as an argument to the makeiso files so they can align to whatever path is being used.

SamTebbs33 (commented):
Very nice reorganisation :+1:

makeiso_64.sh
DrDeano: Sure I shall copy you bug fix over as well

SamTebbs33 (approved):
Very nice

PR #307, opened 2021-05-12
Lock zig master version to ad33e3483

SamTebbs33:
`std.testing.expectEqual` has stopped working for the IdtEntry struct as of https://github.com/ziglang/zig/commit/59f9253d94331cedd4d0518250c8094a064f6cd2 . This PR locks the version of zig we use in CI so we don't have to wait for the issue to be fixed before continuing to work on pluto.﻿

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,8 @@ jobs:
     - name: Download zig
       run: |
         export PYTHONIOENCODING=utf8
-        wget $(curl -s 'https://ziglang.org/download/index.json' | python3 -c "import sys, json; print(json.load(sys.stdin)['master']['x86_64-linux']['tarball'])")
+        # Lock the master commit that we download, because of blocking issues
+        wget https://ziglang.org/builds/zig-linux-x86_64-0.8.0-dev.2133+ad33e3483.tar.xz
         sudo apt-get install mtools
         tar -xvf zig*
     - name: Install qemu
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ All of these goals will benefit from the features of Zig.
 
 ## Build
 
-Requires a master build of Zig ([downloaded](https://ziglang.org/download) or [built from source](https://github.com/ziglang/zig#building-from-source)) *xorriso* and the grub tools (such as *grub-mkrescue*). A *qemu-system* binary compatible with your chosen target is required to run the kernel (e.g. *qemu-system-i386*).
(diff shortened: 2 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -19,7 +19,8 @@ jobs:
     - name: Download zig
       run: |
         export PYTHONIOENCODING=utf8
-        wget $(curl -s 'https://ziglang.org/download/index.json' | python3 -c "import sys, json; print(json.load(sys.stdin)['master']['x86_64-linux']['tarball'])")
+        # Lock the master commit that we download, because of blocking issues
+        wget https://ziglang.org/builds/zig-linux-x86_64-0.8.0-dev.2133+ad33e3483.tar.xz #$(curl -s 'https://ziglang.org/download/index.json' | python3 -c "import sys, json; print(json.load(sys.stdin)['master']['x86_64-linux']['tarball'])")
```
DrDeano: Could remove the commented previous version as it is in the commit history if we need it.

DrDeano (commented):
I guess you can update the README to include the version of Zig that we use

.github/workflows/main.yml
SamTebbs33: Roger roger

ghost:
Hi, wanted to fix that in https://github.com/ZystemOS/pluto/issues/308;

I think it'd be a better idea to fix everything, locking the version can cause problems when trying to compile the codebase with a more recent version of zig ([like here](https://github.com/ZystemOS/pluto/pull/300))

.github/workflows/main.yml
SamTebbs33: Donezo

SamTebbs33:
> I think it'd be a better idea to fix everything, locking the version can cause problems when trying to compile the codebase with a more recent version of zig ([like here](https://github.com/ZystemOS/pluto/pull/300))

I think we should still lock the version until the issues with `expectEqual` are fixed (there is a bug where it complains that equal integers aren't actually equal) as waiting for that to be fixed could take a long time, blocking us from merging PRs. We can go back to master once that is fixed and we have updated our codebase with `try` before `expect*` function calls.

> Hi, wanted to fix that in #308;

A few days ago I spent a couple of painful hours adding `try` before each expect function call so I can send you the code for that so we don't have to duplicate work :)

ghost:
> A few days ago I spent a couple of painful hours adding try before each expect function call so I can send you the code for that so we don't have to duplicate work :)

This would be great

DrDeano (commented):
What about updating the README to include the version of Zig

SamTebbs33:
> What about updating the README to include the version of Zig

Nice idea

DrDeano (approved):

PR #310, opened 2021-06-07
Update source tree for zig 0.8.0

kjaleshire:
Updates certain backwards-incompatible std calls, including:

* [Rename](https://ziglang.org/download/0.8.0/release-notes.html#importbuiltin-no-longer-re-exports-stdbuiltin) `@import("builtin");` to `std.builtin`
* [Rename](https://ziglang.org/download/0.8.0/release-notes.html#importbuiltin-no-longer-re-exports-stdbuiltin) (scroll down) `builtin.arch` to `builtin.cpu.arch` 
* One instance of changing the `assert` surrounding `AutoHashMap.remove` calls.

And one language-specific change:
* Replace `callconv(.Inline)` with keyword `inline`


DrDeano (changes requested):
The format check failed, please check by running `zig build` to run the format.

kjaleshire:
The `zig fmt` call is failing on `callconv(.Inline)` => `inline`. The error is erroneous and will disappear once the build checks use the release version of 0.8.0.

SamTebbs33:
Thanks for doing this @kjaleshire ! Sorry for not looking at it earlier, I've had an issue with GitHub not sending me notifications for everything I want it to. I've added a change that will make the CI use 0.8.0 so that the checks pass.

DrDeano:
Ah, yes, I forgot we locked the zig version for the CI, sorry about that.

SamTebbs33:
I've added `try` before each `expect` function call but 0.8.0 has the issue that made us lock the version in the first place so we won't be able to merge this yet, unfortunately. This could become our zig update PR once the issue is fixed in master.

SamTebbs33:
After a long time away from kernel dev, I've managed to reduce the blockers for this PR to a seg fault when running the kernel. At least the tests are passing now!

SamTebbs33:
I submitted PR 9939 to fix the zig segfault. Once that is merged we can go back to master.

README.md
```diff
@@ -17,7 +17,7 @@ All of these goals will benefit from the features of Zig.
 
 ## Build
 
-Requires a master build of Zig at commit ad33e3483 ([downloaded](https://ziglang.org/builds/zig-linux-x86_64-0.8.0-dev.2133+ad33e3483.tar.xz) or [built from source](https://github.com/ziglang/zig#building-from-source)), *xorriso* and the grub tools (such as *grub-mkrescue*). A *qemu-system* binary compatible with your chosen target is required to run the kernel (e.g. *qemu-system-i386*).
+Requires a master build of Zig 0.8.0([downloaded](https://ziglang.org/download/0.8.0/zig-linux-x86_64-0.8.0.tar.xz) or [built from source](https://github.com/ziglang/zig#building-from-source)), *xorriso* and the grub tools (such as *grub-mkrescue*). A *qemu-system* binary compatible with your chosen target is required to run the kernel (e.g. *qemu-system-i386*).
```
DrDeano: This is now 0.9.1

build.zig
```diff
@@ -108,13 +111,15 @@ pub fn build(b: *Builder) !void {
...
+    const unit_test_options = b.addOptions();
+    unit_tests.addOptions("build_options", unit_test_options);
+    unit_test_options.addOption(TestMode, "test_mode", test_mode);
+    unit_test_options.addOption([]const u8, "mock_path", mock_path);
+    unit_test_options.addOption([]const u8, "arch_mock_path", arch_mock_path);
     unit_tests.setTarget(.{ .cpu_arch = target.cpu_arch });
 
     if (builtin.os.tag != .windows) {
-        unit_tests.enable_qemu = true;
+        b.enable_qemu = true;
```
DrDeano, line 118: How come this has changed to enable qemu for the entire build? As this was used when running the unit tests only.

src/kernel/arch/x86/gdt.zig
```diff
@@ -6,8 +6,7 @@ const builtin = @import("builtin");
 const is_test = builtin.is_test;
 const panic = @import("../../panic.zig").panic;
 const build_options = @import("build_options");
-const mock_path = build_options.arch_mock_path;
-const arch = if (is_test) @import(mock_path ++ "arch_mock.zig") else @import("arch.zig");
+const arch = if (is_test) @import("../../../../test/mock/kernel/arch_mock.zig") else @import("arch.zig");
```
DrDeano, line 9: Given that the `mock_path` is not being used, could remove it from `build.zig` as well.

src/kernel/arch/x86/gdt.zig
```diff
@@ -427,73 +426,73 @@ pub fn init() void {
 }
 
 fn mock_lgdt(ptr: *const GdtPtr) void {
-    expectEqual(TABLE_SIZE, ptr.limit);
-    expectEqual(@ptrToInt(&gdt_entries[0]), ptr.base);
+    expectEqual(TABLE_SIZE, ptr.limit) catch unreachable;
```
DrDeano: Given tests are run in all test modes, should catch panic rather than catch unreachable.

src/kernel/arch/x86/idt.zig
```diff
@@ -196,23 +195,23 @@ fn testHandler0() callconv(.Naked) void {}
 fn testHandler1() callconv(.Naked) void {}
 
 fn mock_lidt(ptr: *const IdtPtr) void {
-    expectEqual(TABLE_SIZE, ptr.limit);
-    expectEqual(@ptrToInt(&idt_entries[0]), ptr.base);
+    expectEqual(TABLE_SIZE, ptr.limit) catch unreachable;
```
DrDeano: Same here.

src/kernel/arch/x86/idt.zig
```diff
@@ -261,7 +259,8 @@ test "openInterruptGate" {
         .base_high = @truncate(u16, test_fn_0_addr >> 16),
     };
 
-    expectEqual(expected_entry0, idt_entries[index]);
+    std.debug.print("Expected {}, actual {}\n", .{ expected_entry0, idt_entries[index] });
```
DrDeano: I assume this debug print can be removed

src/kernel/arch/x86/keyboard.zig
```diff
@@ -164,7 +164,7 @@ fn onKeyEvent(ctx: *arch.CpuState) usize {
     const scan_code = readKeyboardBuffer();
     if (parseScanCode(scan_code)) |action| {
         if (!keyboard.writeKey(action)) {
-            log.notice("No room for keyboard action {}\n", .{action});
+            log.info("No room for keyboard action {}\n", .{action});
```
DrDeano: What is your opinion that this could be a `warn`? As we would be losing key presses, so isn't really info, and is more of a warning.

src/kernel/arch/x86/pit.zig
```diff
@@ -431,30 +431,22 @@ test "setupCounter lowest frequency" {
...
-        // expectEqual(expected_freq, getFrequency());
+        // try expectEqual(u32(0), ticks);
+        // try expectEqual(expected_freq, current_freq_0);
+        // try expectEqual(expected_freq, getFrequency());
 
         // // These are the hard coded expected values. Calculated externally to check the internal calculation
-        // expectEqual(u32(52631578), time_ns);
-        // expectEqual(u32(947), time_under_1_ns);
+        // try expectEqual(u32(52631578), time_ns);
+        // try expectEqual(u32(947), time_under_1_ns);
```
DrDeano: How come these are commented out?

src/kernel/arch/x86/pit.zig
```diff
@@ -468,31 +460,23 @@ test "setupCounter highest frequency" {
...
-    // expectEqual(expected_freq, getFrequency());
+    // try expectEqual(u32(0), ticks);
+    // try expectEqual(expected_freq, current_freq_0);
+    // try expectEqual(expected_freq, getFrequency());
 
     // // These are the hard coded expected values. Calculated externally to check the internal calculation
-    // expectEqual(u32(838), time_ns);
-    // expectEqual(u32(95), time_under_1_ns);
+    // try expectEqual(u32(838), time_ns);
+    // try expectEqual(u32(95), time_under_1_ns);
```
DrDeano: Same here, commented code?

test/mock/kernel/mock_framework_template.zig
```diff
@@ -194,7 +194,7 @@ fn Mock() type {
 
             // Test that the types match
             const expect_type = comptime getDataElementType(expected_function);
-            expectEqual(expect_type, @as(DataElementType, test_element));
+            expectEqual(expect_type, @as(DataElementType, test_element)) catch unreachable;
```
DrDeano: No catch unreachable, should panic because of release-fast/small.
Same for the unreachable's below

test/mock/kernel/vga_mock.zig
```diff
@@ -79,8 +79,8 @@ pub fn orig_entry(uc: u8, c: u8) u16 {
 
 pub fn mock_updateCursor(x: u16, y: u16) void {
     // Here we can do any testing we like with the parameters. e.g. test out of bounds
-    expect(x < WIDTH);
-    expect(y < HEIGHT);
+    expect(x < WIDTH) catch unreachable;
```
DrDeano: Panic, over unreachable

test/runtime_test.zig
```diff
@@ -215,7 +215,7 @@ pub const RuntimeStep = struct {
         _ = try self.os_proc.kill();
 
         // Join the thread
-        thread.wait();
+        thread.join();
```
DrDeano, line 218: Finally, the normal way to wait on a thread with `join` lol

src/kernel/arch/x86/multiboot.zig
```diff
@@ -1,3 +1,52 @@
+pub const __builtin_bswap16 = @import("std").zig.c_builtins.__builtin_bswap16;
```
DrDeano, line 1: I assume these changes are from the translate c of the multiboot.c

src/kernel/arch/x86/paging.zig
```diff
@@ -682,7 +684,7 @@ fn rt_accessMappedMem(v_end: u32) void {
     faulted = false;
     // Accessing mapped memory doesn't cause a page fault
     var ptr = @intToPtr(*u8, v_end - PAGE_SIZE_4KB);
-    var value = ptr.*;
+    log.info("Read value in mapped memory: {}\n", .{ptr.*});
```
DrDeano, line 687: I assume this is needed because the compiler optimises the previous `var value = ptr.*;` out. If so, might be worth adding a comment as having prints in a test is odd.

src/kernel/filesystem/vfs.zig
```diff
@@ -754,7 +756,7 @@ test "mount" {
     var testfs = try testInitFs(allocator);
     defer allocator.destroy(testfs);
     defer testfs.deinit();
-    defer testing.expectEqual(testfs.open_count, 0);
+    defer testing.expectEqual(testfs.open_count, 0) catch unreachable;
```
DrDeano: Catch panic

src/kernel/heap.zig
```diff
@@ -100,8 +99,12 @@ pub const FreeListAllocator = struct {
     /// Arguments:
     ///     IN self: *FreeListAllocator - The allocator being freed within
     ///     IN mem: []u8 - The memory to free
+    ///     IN alignment: u29 - The alignment used to allocate the memory
+    ///     IN ret_addr: usize - The return address passed by the high-level Allocator API. This is ignored.
     ///
-    fn free(self: *FreeListAllocator, mem: []u8) void {
+    fn free(self: *FreeListAllocator, mem: []u8, alignment: u29, ret_addr: usize) void {
```
DrDeano: Why not use the `Self` defined?

src/kernel/filesystem/fat32.zig
```diff
@@ -3985,18 +3983,7 @@ test "Fat32FS.openImpl - match short name" {
...
-}
-
-test "Fat32FS.openImpl - match long name" {
-    const test_fat32_image = try std.fs.cwd().openFile("test/fat32/test_fat32.img", .{});
-    defer test_fat32_image.close();
-
-    var test_fs = try initialiseFAT32(std.testing.allocator, test_fat32_image);
-    defer test_fs.destroy() catch unreachable;
-
-    const file_node = try Fat32FS(@TypeOf(test_fat32_image)).openImpl(test_fs.fs, &test_fs.root_node.node.Dir, "insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long_insanely_long.txt");
```
DrDeano: Could this test be skipped instead of removed, as in the future, this could be fixed.

src/kernel/kmain.zig
```diff
@@ -62,27 +61,26 @@ var kernel_heap: heap.FreeListAllocator = undefined;
...
 
-    kmain_log.info("Init arch " ++ @tagName(builtin.arch) ++ "\n", .{});
+    kmain_log.info("Init arch " ++ @tagName(builtin.cpu.arch) ++ "\n", .{});
     arch.init(&mem_profile);
     kmain_log.info("Arch init done\n", .{});
 
+    panic_root.init(&mem_profile, fixed_allocator.allocator()) catch |e| {
+        panic_root.panic(@errorReturnTrace(), "Failed to initialise panic: {}\n", .{e});
+    };
+
```
DrDeano: Given this is now move lower down, there are still calls to panic above. If these `panic` calls are called, what will happen? Will it just print address and no symbols. Might be worth changing the `init` function to `init_symbols` or something similar to indicate that this doesn't initialise the panic handler, but initialises the symbol mappings.

DrDeano (commented):
I congratulate the effort for this :) I expect only the finest vim macros were used ;)

build.zig
SamTebbs33: The `enable_qemu` field was been moved from there to the `Builder` unfortunately ([this commit](https://github.com/ziglang/zig/commit/b24cbecdb2abb4399d65553ebade318263cd57d3) for reference).

src/kernel/arch/x86/gdt.zig
SamTebbs33: The program will exit if the `expectEqual` fails though, right? I don't think the `catch` part will ever run.

src/kernel/arch/x86/idt.zig
SamTebbs33: Yep, good spot.

src/kernel/arch/x86/keyboard.zig
SamTebbs33: Yeah that sounds better to me

src/kernel/arch/x86/pit.zig
SamTebbs33: Hmm I don't remember, I'll uncomment them and hope they pass...

SamTebbs33: They're commented out in develop too. I don't think I have the knowledge of the PIT to get them working again.

src/kernel/arch/x86/multiboot.zig
SamTebbs33: Indeedy

src/kernel/arch/x86/paging.zig
SamTebbs33: Yeah exactly. Good idea

src/kernel/heap.zig
SamTebbs33: That is actually within the `Header` struct, but I can add one for `FreeListAllocator` too.

src/kernel/filesystem/fat32.zig
SamTebbs33: Agreed. How do I do that? Just comment it out or is there a fancy way of skipping tests?

src/kernel/kmain.zig
SamTebbs33: Yeah if something above this panics then it will just print the addresses without symbols. I like that idea, it makes the purpose clear.

build.zig
DrDeano: Fair enough

src/kernel/filesystem/fat32.zig
DrDeano: Can return `error.SkipZigTest`. This can be found in `special\test_runner.zig`

src/kernel/arch/x86/gdt.zig
DrDeano: `expectEqual` will return `error.TestExpectedEqual` if the values are not equal, so for ReleaseFast, the `catch unreachable` will be ignored and `expectEqual` won't fail the test

src/kernel/arch/x86/pit.zig
DrDeano: Interesting? I shall investigate

src/kernel/arch/x86/gdt.zig
SamTebbs33: Yep you're right. I think it used to work in the way I was thinking.

src/kernel/filesystem/fat32.zig
SamTebbs33: NIce

src/kernel/arch/x86/pit.zig
DrDeano: I had a look and these were commented from the beginning. These commented lines can be removed as they don't relate to the test. I think they might be from a copy and paste error and forgot to remove them.

src/kernel/arch/x86/pit.zig
DrDeano: Can remove these lines

DrDeano (approved):

SamTebbs33:
Thanks for the approval :) Looks like the checks are messing up. [This](https://stackoverflow.com/a/69169858) or temporarily allowing maintainers to bypass the checks may help.

DrDeano:
> Thanks for the approval :) Looks like the checks are messing up. [This](https://stackoverflow.com/a/69169858) or temporarily allowing maintainers to bypass the checks may help.

Sure, can just temporary disable the check, then re-enable after the merge

SamTebbs33:
Hmm it's still not working... I'll try submitting a new PR to see if it's just a problem with this one.

PR #314, opened 2021-10-31
Run CI in docker container

SamTebbs33:
This PR runs the CI in a docker container that has llvm 12 and zig ad33e3483 installed. The intention behind this is to get CI passing on a working version of zig master so we don't have to wait for #310﻿


DrDeano (approved):

SamTebbs33:
Not necessary, as of #310

PR #315, opened 2022-02-26
Use prebuilt zig of commit ad33e3483

SamTebbs33:
This PR makes CI use a prebuilt tar of zig at commit ad33e3483, which is the last known version to work compile pluto.﻿


SamTebbs33:
Ah of course, I built this for arm64 macos but it needs to be for x86_64 linux.

DrDeano (approved):

SamTebbs33:
No longer necessary as of #310

PR #317, opened 2022-04-23
Improve syscall error conversion

SamTebbs33:
This PR makes the conversion from syscall errors to integers (and vice-versa) much cleaner.

### DIFF
--- a/src/kernel/arch/x86/syscalls.zig
+++ b/src/kernel/arch/x86/syscalls.zig
@@ -17,7 +17,7 @@ pub const INTERRUPT: u16 = 0x80;
 pub const NUM_HANDLERS: u16 = 256;
 
 /// A syscall handler
-pub const Handler = fn (arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) syscalls.Error!usize;
+pub const Handler = fn (arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) anyerror!usize;
 
 /// Errors that syscall utility functions can throw
 pub const Error = error{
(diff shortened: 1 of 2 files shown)


SamTebbs33:
ping for comments

SamTebbs33:
@DrDeano ping :)

src/kernel/arch/x86/syscalls.zig
```diff
@@ -108,21 +108,21 @@ pub fn registerSyscall(syscall: usize, handler: Handler) Error!void {
...
         : "ebx"
     );
-    const err = asm (""
+    const err = @intCast(u16, asm (""
         : [ret] "={ebx}" (-> usize),
-    );
+    ));
     if (err != 0) {
-        return syscalls.fromErrorCode(err);
+        return syscalls.fromErrorCode(@intCast(u16, err));
```
DrDeano: Is the int cast needed as you could try something:
```zig
const err = asm ("" : [ret] "={bx}" (-> u16));
```
Not sure if this would work, or it the zig's type system will recognise it being a u16

src/kernel/arch/x86/syscalls.zig
```diff
@@ -388,34 +390,34 @@ fn testHandler2(arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize)
...
-fn testHandler6(arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) syscalls.Error!usize {
+fn testHandler6(arg1: usize, arg2: usize, arg3: usize, arg4: usize, arg5: usize) anyerror!usize {
     // Suppress unused variable warnings
     _ = arg1;
     _ = arg2;
     _ = arg3;
     _ = arg4;
     _ = arg5;
-    return syscalls.Error.OutOfMemory;
+    return anyerror.OutOfMemory;
```
DrDeano: Should this be `error.OutOfMemory`. You have anyerror.

src/kernel/arch/x86/syscalls.zig
DrDeano: Also you are `@intCast`ing twice

DrDeano (commented):
Sorry for the late review :P
Some minor points

src/kernel/arch/x86/syscalls.zig
SamTebbs33: Well spotted!

src/kernel/arch/x86/syscalls.zig
SamTebbs33: That works too and looks better, cheers.

DrDeano (approved):
bam 🚀

SamTebbs33:
> bam rocket

Thanks :)

PR #318, opened 2022-04-23
Combine ComptimeBitmap and Bitmap

SamTebbs33:
This PR combines ComptimeBitmap and Bitmap, which is useful as they share a lot of functionality. It also makes ComptimeBitmap more flexible (e.g. it can now uses an array of bitmap types rather than a single one).

### DIFF
--- a/src/kernel/bitmap.zig
+++ b/src/kernel/bitmap.zig
@@ -1,206 +1,33 @@
 const std = @import("std");
 const builtin = std.builtin;
 const testing = std.testing;
+const expectEqual = testing.expectEqual;
 const Allocator = std.mem.Allocator;
 
-///
-/// A comptime bitmap that uses a specific type to store the entries. No allocators needed.
-///
-/// Arguments:
-///     IN BitmapType: type - The integer type to use to store entries.
-///
-/// Return: type.
-///     The bitmap type created.
-///
-pub fn ComptimeBitmap(comptime BitmapType: type) type {
-    return struct {
-        const Self = @This();
-
-        /// The number of entries that one bitmap type can hold. Evaluates to the number of bits the type has
-        pub const NUM_ENTRIES: usize = std.meta.bitCount(BitmapType);
-
-        /// The value that a full bitmap will have
-        pub const BITMAP_FULL = std.math.maxInt(BitmapType);
-
-        /// The type of an index into a bitmap entry. The smallest integer needed to represent all bit positions in the bitmap entry type
-        pub const IndexType = std.meta.Int(.unsigned, std.math.log2(std.math.ceilPowerOfTwo(u16, std.meta.bitCount(BitmapType)) catch unreachable));
-
-        bitmap: BitmapType,
-        num_free_entries: BitmapType,
-
-        ///
-        /// Create an instance of this bitmap type.
-        ///
-        /// Return: Self.
-        ///     The bitmap instance.
-        ///
-        pub fn init() Self {
-            return .{
-                .bitmap = 0,
-                .num_free_entries = NUM_ENTRIES,
-            };
-        }
-
-        ///
-        /// Set an entry within a bitmap as occupied.
-        ///
-        /// Arguments:
-        ///     IN/OUT self: *Self - The bitmap to modify.
-        ///     IN idx: IndexType - The index within the bitmap to set.
-        ///
-        pub fn setEntry(self: *Self, idx: IndexType) void {
-            if (!self.isSet(idx)) {
-                self.bitmap |= indexToBit(idx);
-                self.num_free_entries -= 1;
-            }
-        }
-
-        ///
-        /// Set an entry within a bitmap as unoccupied.
-        ///
-        /// Arguments:
-        ///     IN/OUT self: *Self - The bitmap to modify.
-        ///     IN idx: IndexType - The index within the bitmap to clear.
-        ///
-        pub fn clearEntry(self: *Self, idx: IndexType) void {
-            if (self.isSet(idx)) {
-                self.bitmap &= ~indexToBit(idx);
-                self.num_free_entries += 1;
-            }
-        }
-
-        ///
-        /// Convert a global bitmap index into the bit corresponding to an entry within a single BitmapType.
-        ///
-        /// Arguments:
-        ///     IN self: *const Self - The bitmap to use.
-        ///     IN idx: IndexType - The index into all of the bitmaps entries.
-        ///
-        /// Return: BitmapType.
-        ///     The bit corresponding to that index but within a single BitmapType.
-        ///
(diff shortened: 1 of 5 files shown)


src/kernel/scheduler.zig
```diff
@@ -14,6 +14,7 @@ const mem = @import("mem.zig");
 const fs = @import("filesystem/vfs.zig");
 const elf = @import("elf.zig");
 const pmm = @import("pmm.zig");
+const bitmap = @import("bitmap.zig");
```
DrDeano: This import isn't used

src/kernel/task.zig
```diff
@@ -163,21 +162,24 @@ pub const Task = struct {
...
-    if (!all_pids.isSet(pid)) {
+/// Error: BitmapError.
+///     OutOfBounds: The index given is out of bounds.
+///
+fn freePid(pid: usize) bitmap.BitmapError!void {
+    if (!(try all_pids.isSet(pid))) {
         panic(@errorReturnTrace(), "PID {} not allocated\n", .{pid});
     }
-    all_pids.clearEntry(pid);
+    all_pids.clearEntry(pid) catch unreachable;
```
DrDeano: Given this function now can return an error, can `try all_pids.clearEntry(pid)` instead of `catch unreachable`.
But perhaps this function shouldn't return an error and just panic as this would be used in `defer` statements, removing the need for catching the errors in defer statements.

src/kernel/task.zig
```diff
@@ -23,14 +22,14 @@ extern var KERNEL_STACK_START: *u32;
 pub const EntryPoint = usize;
 
 /// The bitmap type for the PIDs
-const PidBitmap = if (is_test) ComptimeBitmap(u128) else ComptimeBitmap(u1024);
+const PidBitmap = if (is_test) bitmap.Bitmap(true, usize, 128) else bitmap.Bitmap(true, usize, 1024);
```
DrDeano: Was there a purpose for having difference bitmaps for testing and for runtime as it does seem a bit odd.

src/kernel/vmm.zig
```diff
@@ -305,7 +305,10 @@ pub fn VirtualMemoryManager(comptime Payload: type) type {
         /// Error: pmm.PmmError
         ///     Bitmap(u32).Error.OutOfBounds - The address given is outside of the memory managed
         ///
-        pub fn isSet(self: *const Self, virt: usize) bitmap.Bitmap(u32).BitmapError!bool {
+        pub fn isSet(self: *const Self, virt: usize) bitmap.BitmapError!bool {
+            if (virt < self.start) {
+                return bitmap.BitmapError.OutOfBounds;
+            }
```
DrDeano, line 311: Do the exiting tests cover this new validation, or should there be an additional tests

src/kernel/pmm.zig
```diff
@@ -20,7 +21,7 @@ const PmmError = error{
 /// The size of memory associated with each bitmap entry
 pub const BLOCK_SIZE: usize = arch.MEMORY_BLOCK_SIZE;
 
-var bitmap: PmmBitmap = undefined;
+var the_bitmap: PmmBitmap = undefined;
```
DrDeano, line 24: Perhaps we could start to move to the newer way of globals where this will just be: `bitmap: PmmBitmap,`. As when importing `pmm`, it is already a struct.
`std.mem.Allocator` is an example.

src/kernel/bitmap.zig
```diff
@@ -666,10 +545,10 @@ test "setFirstFree multiple bitmaps" {
     // Allocate the entirety of the first bitmap
     var entry: u32 = 2;
     var expected: u8 = 7;
-    while (entry < Bitmap(u8).ENTRIES_PER_BITMAP) {
+    while (entry < @TypeOf(bmp).ENTRIES_PER_BITMAP) {
```
DrDeano: Am being picky, but would it be nicer to have:
```zig
const BmpTy = Bitmap(false, u8, null);
var bmp = try BmpTy.init(9, std.testing.allocator);
```
so you don't have to call `@TypeOf` all the time. But up to you.

src/kernel/bitmap.zig
```diff
@@ -1,205 +1,31 @@
...
 /// Arguments:
+///     IN static: comptime bool - Whether this bitmap is statically or dynamically allocated
 ///     IN BitmapType: type - The integer type to use to store entries.
+///     IN num_entries: usize or ?usize - The number of entries if static, else ignored (can be null)
 ///
 /// Return: type.
 ///     The bitmap type created.
 ///
-pub fn Bitmap(comptime BitmapType: type) type {
+pub fn Bitmap(comptime static: bool, comptime BitmapType: type, comptime num_entries: if (static) usize else ?usize) type {
```
DrDeano: Could you comptime check on whether `num_entries` is `null` and if not null, then is static, else is alloc. So can use the 'null capture if' statement: `if (num_entries) |blah| {static} else {alloc}`. Then this would remove the `static` parameter.

DrDeano (commented):
This is very nice

src/kernel/task.zig
SamTebbs33: I think I agree with making it `try` instead :+1:

src/kernel/task.zig
SamTebbs33: I don't remember actually. The tests pass after removing the special case :)

src/kernel/vmm.zig
SamTebbs33: It isn't covered, thanks for spotting that.

src/kernel/pmm.zig
SamTebbs33: I haven't heard of this new way of using globals, could you share alink?

src/kernel/bitmap.zig
SamTebbs33: Yeah that looks good to me.

src/kernel/pmm.zig
DrDeano: https://github.com/ziglang/zig/blob/master/lib/std/mem/Allocator.zig
or
https://github.com/ziglang/zig/blob/master/src/Compilation.zig
Looks like lots of the compiler is written like this

src/kernel/bitmap.zig
SamTebbs33: That's much nicer :)

src/kernel/pmm.zig
DrDeano: Perhaps this can be made into an issue as this doesn't relate to this PR and can be done as a separate PR

SamTebbs33: Oh that's pretty cool! I agree with it being an issue instead.

DrDeano (approved):

PR #322, opened 2022-06-05
Ignore zig-out

SamTebbs33:
Ignores the zig-out directory, which is the new standard directory for zig output.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 # Zig ignore
 **/zig-cache/
+**/zig-out/
 **/build/
 **/build-*/
 **/docgen_tmp/


.gitignore
```diff
@@ -14,3 +14,4 @@
 **/mock_framework.zig
 **/*.ramdisk
 **/*.img
+zig-out/
```
DrDeano: Shouldn't this go under the `# Zig ignore` section

DrDeano (approved):
Am approve once comment is addressed

.gitignore
SamTebbs33: It should :) I have a shell alias that just appends what I want to ignore to the end of the file

PR #323, opened 2022-06-05
Create task 0 like other tasks

SamTebbs33:
This PR makes the creation of task 0 the same as normal tasks, to make the code cleaner.

### DIFF
--- a/src/kernel/scheduler.zig
+++ b/src/kernel/scheduler.zig
@@ -137,15 +137,13 @@ pub fn init(allocator: Allocator, mem_profile: *const mem.MemProfile) Allocator.
     // Init the task list for round robin
     tasks = TailQueue(*Task){};
 
-    // Set up the init task to continue execution
-    current_task = try allocator.create(Task);
+    // Set up the init task to continue execution.
+    // The kernel stack will point to the stack section rather than the heap
+    current_task = try Task.create(0, true, &vmm.kernel_vmm, allocator, false);
     errdefer allocator.destroy(current_task);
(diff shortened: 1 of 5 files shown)


SamTebbs33:
@DrDeano ping a ding ding

src/kernel/scheduler.zig
```diff
@@ -138,14 +138,12 @@ pub fn init(allocator: Allocator, mem_profile: *const mem.MemProfile) Allocator.
     tasks = TailQueue(*Task){};
 
     // Set up the init task to continue execution
-    current_task = try allocator.create(Task);
+    current_task = try Task.create(0, true, &vmm.kernel_vmm, allocator);
     errdefer allocator.destroy(current_task);
-    // PID 0
-    current_task.pid = 0;
+
+    allocator.free(current_task.kernel_stack);
```
DrDeano: Might be worth adding a comment for why this free is here as a new kernel stack is allocated in `create`. Or have a way of not allocating a kernel stack in `Task.create`. Could have a `createWithStack` and you pass in address range for the kernel stack and a `createWithoutStack` which allocates the kernel stack. The naming of the functions can something else if you like. As allocating then immediately freeing memory does seem to be a bit dumb.

SamTebbs33: Yeah I definitely agree. Perhaps a boolean parameter would be a nice idea.

SamTebbs33:
I've finally finished this! I was stuck on a page fault and index out of bounds error for a little while.

PR #324, opened 2022-06-13
Putting arch file into a package.

### LINKED ISSUE #77: Import arch file as a package
Instead of the existing system of importing the arch.zig file, have the architecture arch.zig file be added as a package in the build script. This will remove the kernel level arch.zig file. This will also help with the testing.

Is blocked by:
https://github.com/ziglang/zig/issues/855

We can't do this as we have the kernel as a package and as the kernel will import the arch package and arch will import the kernel package, will need to wait for this.

Dawid33:
issue #77 

Unfortunately, arch.zig and the kernel are dependant on each other. I couldn't think of a way to have two pacakges directly depend on each other. I made a file pluto.zig that collects and re-exports all the files in the kernel. I made packages out of arch.zig and pluto.zig. kmain.zig then imports those packges.

This works for just running the kernel but getting the tests to work will be a bigger challenge.


build.zig
```diff
@@ -68,6 +72,14 @@ pub fn build(b: *Builder) !void {
     exec.setLinkerScriptPath(std.build.FileSource{ .path = linker_script_path });
     exec.setTarget(target);
 
+    const kernel_pkg = Pkg{ .name = "pluto", .path = .{ .path = pluto_src }, .dependencies = &[_]Pkg{exec_options.getPackage("build_options")} };
+    const arch_mock = Pkg{ .name = "arch_mock", .path = .{ .path = arch_mock_src }, .dependencies = &[_]Pkg{ kernel_pkg, exec_options
+        .getPackage("build_options") } };
+    const arch_pkg = Pkg{ .name = "arch", .path = .{ .path = arch_path }, .dependencies = &[_]Pkg{ kernel_pkg, arch_mock, exec_options
```
SamTebbs33: Theoretically we don't need both the arch_mock and normal arch when testing, so we could just have one arch package that is the mocked one when testing and the normal one when not testing.

I'm not sure if this will work so I can test this.

build.zig
```diff
@@ -88,6 +100,9 @@ pub fn build(b: *Builder) !void {
         inline for (&[_][]const u8{ "user_program_data", "user_program" }) |user_program| {
             // Add some test files for the user mode runtime tests
             const user_program_step = b.addExecutable(user_program ++ ".elf", null);
+            user_program_step.addPackage(kernel_pkg);
+            user_program_step.addPackage(arch_mock);
+            user_program_step.addPackage(arch_pkg);
```
SamTebbs33: I don't think the user programs will need the packages as they're just assembly files.

src/kernel/arch.zig
```diff
@@ -3,7 +3,7 @@ const builtin = @import("builtin");
 const is_test = builtin.is_test;
 const build_options = @import("build_options");
 
-pub const internals = if (is_test) @import("../../test/mock/kernel/arch_mock.zig") else switch (builtin.cpu.arch) {
+pub const internals = if (is_test) @import("arch_mock") else switch (builtin.cpu.arch) {
     .i386 => @import("arch/x86/arch.zig"),
```
SamTebbs33: We already have a package for "arch/x86/arch.zig" so could this import that package instead?

src/kernel/arch/x86/arch.zig
```diff
@@ -525,7 +525,7 @@ pub fn initMem(mb_info: BootPayload) Allocator.Error!MemProfile {
 /// Error: std.mem.Allocator.Error
 ///     OutOfMemory - There wasn't enough memory to allocate what was needed
 ///
-pub fn initKeyboard(allocator: Allocator) Allocator.Error!*Keyboard {
+pub fn initKeyboard(allocator: Allocator) Allocator.Error!*keyboard.Keyboard {
```
SamTebbs33: We should still be able to just use `Keyboard` without the `keyboard.` prefix.

src/kernel/arch/x86/idt.zig
```diff
@@ -7,8 +7,8 @@ const builtin = @import("builtin");
 const is_test = builtin.is_test;
 const panic = @import("../../panic.zig").panic;
 const build_options = @import("build_options");
-const gdt = if (is_test) @import("../../../../test/mock/kernel/gdt_mock.zig") else @import("gdt.zig");
-const arch = if (builtin.is_test) @import("../../../../test/mock/kernel/arch_mock.zig") else @import("arch.zig");
+const gdt = if (is_test) @import("arch_mock").gdt_mock else @import("gdt.zig");
+const arch = if (builtin.is_test) @import("arch_mock") else @import("arch.zig");
```
SamTebbs33: I love how much cleaner this is.

src/kernel/arch/x86/arch.zig
```diff
@@ -16,15 +16,15 @@ const serial = @import("serial.zig");
...
-const Serial = @import("../../serial.zig").Serial;
-const panic = @import("../../panic.zig").panic;
-const TTY = @import("../../tty.zig").TTY;
-const Keyboard = @import("../../keyboard.zig").Keyboard;
-const Task = @import("../../task.zig").Task;
+const Keyboard = @import("pluto").keyboard.Keyboard;
+const Serial = @import("pluto").serial.Serial;
+const panic = @import("pluto").panic_root.panic;
+const TTY = @import("pluto").tty.TTY;
+const Task = @import("pluto").task.Task;
```
SamTebbs33: It would be nice to import "pluto" once like in kmain.

test/mock/kernel/arch_mock.zig
```diff
@@ -1,18 +1,19 @@
...
-const TTY = @import("../../../src/kernel/tty.zig").TTY;
-const Keyboard = @import("../../../src/kernel/keyboard.zig").Keyboard;
-const task = @import("../../../src/kernel/task.zig");
-const x86_paging = @import("../../../src/kernel/arch/x86/paging.zig");
+const Serial = @import("pluto").serial.Serial;
+const TTY = @import("pluto").tty.TTY;
+const Keyboard = @import("pluto").keyboard.Keyboard;
+const task = @import("pluto").task;
+const x86_paging = @import("arch").paging;
+pub const cmos = @import("cmos_mock.zig");
```
SamTebbs33: It would be nice to import pluto once here too.

SamTebbs33 (changes requested):
I like how much cleaner this makes things.

build.zig
Dawid33: Yeah I wasn't sure what the code was doing so I put them in just in case.

src/kernel/arch/x86/arch.zig
Dawid33: Ah that was an accident, didn't notice that was in the commit.

build.zig
Dawid33: This is still very wip, I started working on this project today. I'm not entirely sure how these test work, so I put in arch_mock to see if I can get it to run under another package. If I understand what your suggesting, its to move all of test/mock/kernel into kernel/arch/x86?

src/kernel/arch.zig
Dawid33: If you're talking about line 7, no it can't because that would make pluto_pkg dependent on arch_pkg which I don't think can work. I'm trying to think of a way to remove this file entirely and just have the kernel do @import("arch") everywhere. I'm not even sure if it will be possbie to completely seperate the kernel from arch without some sort of glue in between like a common package that has no depencancies

Dawid33: I really don't know though. I've never done kernel dev nor have I done much with zig, I just started working on this today haha, thank you for taking the time to review this though

SamTebbs33: I see, that makes sense. It would be cool if we could just import "arch" everywhere but if that isn't possible then no problem.

Dawid33: So what I'm thinking is to make a package arch-common that both pluto and arch depend on. It would be filled with optional function pointers that return types. Then if you implement an arch, you just have to implement everything in arch-common to get that arch working under pluto. The main obvious downside to this is that pluto would need to null check anything it tries to use from arch-common. Its a lot more work that I thought it would be. I'll try that anyway, see how far I get.

Dawid33:
Sorry for the stream of force pushes, I got it working properly. I thought it wasn't possible to have circular references but it actually is :), I went off a on a tangent yesterday haha. *Most* of the tests are passing. Still needs some work.

fat32_cp.sh
```diff
@@ -8,7 +8,7 @@ whoami
 if [ "$(whoami)" = "root" ]; then
 	echo "Am root"
 	mount -o utf8=true $IMAGE_PATH_DIR test/fat32/mnt/
-	cp -r test/fat32/test_files/. test/fat32/mnt/
+	cp -r test/fat32/test_files/. test/fat32/mnt/ &&
```
SamTebbs33: What does this do?

SamTebbs33 (approved):
Thanks for doing this, awesome work. We like to keep merges to one commit for conciseness, so if you don't mind squashing them into one and force pushing, then I'm happy for this to be merged.

I'd like to wait a reasonable amount of time for @DrDeano to have a look as well.

Closes #77

fat32_cp.sh
Dawid33: ah yeah I'll remove that. I'm constantly having an issue that when running that script without root it tries to umount that folder but its busy so it fails. Then I have to manually umount and delete. I cant remember if this worked or not. I'll get rid of this and force push that single commit

SamTebbs33: Yeah I've had that issue before as well. We're hoping to move away from mounting and use mtools.

src/kernel/arch/x86/syscalls.zig
```diff
@@ -3,12 +3,13 @@ const log = std.log.scoped(.x86_syscalls);
 const builtin = @import("builtin");
 const is_test = builtin.is_test;
 const build_options = @import("build_options");
-const arch = if (is_test) @import("../../../../test/mock/kernel/arch_mock.zig") else @import("arch.zig");
+const arch = if (is_test) @import("arch_mock") else @import("arch.zig");
```
DrDeano, line 6: For the else case, would the import be `@import("arch")` instead of `@import("arch.zig")`
Or does this not work for arch file and is only used for the kernel files.
This is the same for all files.

src/kernel/arch/x86/arch.zig
```diff
@@ -2,12 +2,13 @@ const std = @import("std");
 const Allocator = std.mem.Allocator;
 const log = std.log.scoped(.x86_arch);
 const builtin = @import("builtin");
+const is_test = builtin.is_test;
```
DrDeano: Look like this is not used.

DrDeano (commented):
Soz for the late review. I am slow.
Two small things.
This is very nice, (Insert the nice * mouth click* meme)

src/kernel/arch/x86/arch.zig
Dawid33: Ah good catch my bad.

src/kernel/arch/x86/syscalls.zig
Dawid33: So the reasoning behind this is that the arch package is `kernel/arch/<specific arch>/arch.zig`. If it were importing "arch" it would be importing itself. If you change arch.zig to arch you'll see an error saying that the arch package is missing because it is unnecessary for a package to import itself, it already has access to itself since it is itself. I'm not certain its possible for a package to import itself regardless. The root of the "arch" package is arch.zig, so it makes sense to import "arch.zig" anyway since its in that folder.

Using arch.zig instead of arch is only in the `arch/<specific arch>` directory. The rest of the kernel uses just "arch", which is a package that is decided at compile time depending on what architecture is used. That said, I was looking around the other branches and there's a few other things on the way so I was thinking to merge all the other feature branches with develop to bring them up to date to use this system of packages. Might open an issue for that if that's okay.

DrDeano: Fab, thanks for the clarification.

DrDeano (approved):
🚀

